Issue #31, opened 2024-05-21
Create backend web server implementation with Rust framework


ebma:
## Context
As agreed on in https://github.com/pendulum-chain/tasks/issues/159 and mentioned in #15, we'd like to eventually implement the backend for the pendulum-pay service in Rust. Since the knowledge/experience with Rust webserver frameworks is currently rather limited in the tech team, we want to start exploring this intricacies of implementing and deploying a Rust-based webserver. 

## TODO

Re-implement the web server with the [axum](https://github.com/tokio-rs/axum) framework. The branch for these changes should base on the [offramp-prototype-staging](https://github.com/pendulum-chain/pendulum-pay/tree/offramp-prototype-staging) branch. 

- Remove the contents of the [signer-service](https://github.com/pendulum-chain/pendulum-pay/tree/b22d2dd8995d4e9a1b1c9aaded11074cae3ba269/signer-service) directory.
- In the `signer-service` directory, start a new cargo project for the Rust web server. 
- Keep the same functionality as described [here](https://github.com/pendulum-chain/pendulum-pay/issues/20#issuecomment-2072989063) intact. The relevant code files for reference are [stellar.service.js](https://github.com/pendulum-chain/pendulum-pay/blob/b22d2dd8995d4e9a1b1c9aaded11074cae3ba269/signer-service/src/api/services/stellar.service.js), [stellar.route.js](https://github.com/pendulum-chain/pendulum-pay/blob/b22d2dd8995d4e9a1b1c9aaded11074cae3ba269/signer-service/src/api/routes/v1/stellar.route.js) and [stellar.controller.js](https://github.com/pendulum-chain/pendulum-pay/blob/b22d2dd8995d4e9a1b1c9aaded11074cae3ba269/signer-service/src/api/controllers/stellar.controller.js).

## Notes

This may or may not replace the current Javascript implementation of the prototype backend. As mentioned in the 'Context' section, the main outcome of this ticket is exploring the respective framework and deployment.

PR #37, opened 2024-06-06
[DO NOT MERGE] Create backend web server implementation with rust framework

b-yap:
addresses #31 

command: `RUST_LOG=info cargo run`, and it will show:
```
INFO signer_service_rust: 🚀LISTENING - Ok(127.0.0.1:3001)
```
Looks like the provided secret key doesn't work on PUBLIC network, so currently I used TESTNET.


ebma:
I'm closing this for now, we can reopen it once needed.

Issue #102, opened 2024-08-14
Implement the SEP24 flow in an iframe


prayagd:
## Context
Currently the MYKOBO flow is implemented as redirect to other tab when user clicks on "Start offramping", Instead of that it should be iframe on the same screen.

## Acceptance Criteria
* On the Vortex dApp after the user agrees and clicks on "Start offramping"
* The MYKOBO page should open-up as a iframe and user should be complete the steps from that iframe
* Show the input summary card with all the details shown in the design attached above the iframe
* If its the first time dApp user, then it should start from the KYC page. 
    * With the following title "Verify your identity" 
    * With the following text above the iframe "Please follow the steps below to complete your identity verification."
* If the user is already KYC'ed, they should land on the banking details page. 
    * With the following title "Enter bank details" 
    * With the following text above the iframe "Please do not change the amount on offramp UI"
* The text and title should change when the first time KYC user transitions to banking details page
* After completing the steps on the MYKOBO UI, the iframe should close and the vortex dAPP should start the next steps i.e signing of the polygon txn automatically. This should happen immediately.

## Designs
https://www.figma.com/proto/cOx99YQtxf8LLdlhZjvkoC/Vortex?page-id=330%3A556&node-id=330-557&node-type=FRAME&viewport=-2578%2C-399%2C0.34&t=VO6VI0x4i1oMtmwu-1&scaling=min-zoom&content-scaling=fixed&starting-point-node-id=330%3A557

PR #127, opened 2024-09-02
 [DO NOT MERGE - Iceboxed] Add Iframe and Summary card for the KYC

gianfra-t:
Available to test UI as offramp is mocked.

Issue #102. Any amount will work and launch the iframe page, which will complete after some seconds mocking the user finishing the process. 

### Changes
- Removes the redirect action for "Start Offramping" button that used to open the KYC site on a different page.
- Remove "Confirm" button and replace with only one click on the "Start Offramping" button to start the process.
- The KYC is now embedded in an iframe inside the application, which closes and continues to the progress page once the `sep24process` is completed (response from anchor).


TorstenStueber:
I would also suggest to just close it instead of keeping it open with the warning in the title.

ebma:
@gianfra-t before closing it, can you leave a comment with a brief summary that explains why we cannot proceed with this change? Namely pointing out the restrictions with the browser settings that interferes with the cookies in the iframe etc.

TorstenStueber:
@gianfra-t let's close this PR? We can still reopen later if we want to go back to the iframe.

PR #166, opened 2024-09-19
Fix input field bugs

### LINKED ISSUE #164: Fix clunky amount input field
The input field for the input amount has some clunky UX.

For example, 
* if the input is "0.12" and the user marks the text and presses "1", then
  * expected: the content of the input field changes to "1"
  * actually: the content of the input field changes to "0.11"
* if the input is "0.12" and the user types ".", then
  * expected: ignore the input
  * actually: the content changes to "0.1"
* if the input is "0.23" and the cursor is between "0" and "." and the user enters "2"
  * expected: input changes to "02.23"
  * actually: input changes to "0.22"
* if the input is "0.23" and the cursor is between "." and "2" and the user enters "4"
  * expected: input changes to "0.42"
  * actually: input changes to "0.24"

A mostly correct implementation of this logic is [the `handleOnChangeNumericInput` function in the Portal](https://github.com/pendulum-chain/portal/blob/eb11d0666edd65703c104001dc1523f0311bf6a8/src/components/Form/From/NumericInput/helpers.ts#L14-L22).

I say _mostly_ correct because that implementation has unexpected results when a user enters "." when there is already a "." in the input:
* if the user input is "0.23" and the cursor is between "2" and "3" and the user enters "."
  * expected: ignore the input as there is already one "." in the input field further to the left
  * actual: input changes to "0.2"
* if the user input is "12.34" and the cursor is between "1" and "2" and the user enters "."
  * expected: either ignore the input as there is already a "." in the input field further to the right or replace input to "1.234"
  * actual" input changes to "1.2"
  
## TODO
Implement the input handling so that all cases above behave like as expected.

gianfra-t:
Fixes #164.

Additionally, it fixes:
- Crash when backspace is tapped multiple times, causing input empty string.
- Management of trailing zeros.


ebma:
We merged the equivalent changes to the portal in https://github.com/pendulum-chain/portal/pull/561. @gianfra-t please copy the `src/components/Form/From/NumericInput/helpers.ts` and `src/components/Form/From/NumericInput/index.tsx` files here so we can profit from the same fixes.  No need to copy the test as it won't work in this repository due to the lacking jest configuration.

src/pages/swap/index.tsx
```diff
@@ -123,12 +122,12 @@ export const SwapPage = () => {
       // Calculate the final amount after the offramp fees
       const totalReceive = calculateTotalReceive(toAmount.toString(), toToken);
       form.setValue('toAmount', totalReceive);
-
-      setIsQuoteSubmitted(false);
+    } else if (!tokenOutData.isLoading || tokenOutData.error) {
+      form.setValue('toAmount', '0');
     } else {
-      form.setValue('toAmount', '');
+      // Do nothing?
```
TorstenStueber, line 128: I think it makes sense not to do anything while it is loading.

TorstenStueber (commented):
Can you please also fix the following bug: https://github.com/pendulum-chain/vortex/pull/166/files#diff-777e0c2221f5feeef076c359496045ddd95d6d376a0fc1395b330f8a9b64a5bbR91

Here we also need to check whether `tokenOutData.stableAmountInUnits` is the empty string. Otherwise the `Big` constructor throws an error. If it is the empty string, then we can treat `inputAmountIsStable` as `false`.

gianfra-t:
Yes, I copied the portal fix but it is not ready yet.  As you say it requires a few extra checks to work properly here.

src/pages/swap/index.tsx
```diff
@@ -1,4 +1,5 @@
 import { useEffect, useMemo, useRef, useState } from 'preact/hooks';
+import { Fragment } from 'preact';
```
TorstenStueber, line 2: This is not used.

TorstenStueber:
@gianfra-t Thanks for the changes. Can you please still implement this? https://github.com/pendulum-chain/vortex/pull/166#pullrequestreview-2344158687

gianfra-t:
@TorstenStueber I (re)added the empty string condition [here](https://github.com/pendulum-chain/vortex/pull/166/commits/3d15ae6ef4f908fa9c765a75e5047b1b458978df#diff-f1681ac32d3ac3056cda698dd932d65084890702345d1f658d1fb8269d7f4fe1R48), is it the same issue? `tokenOutData.stableAmountInUnits` can't be empty if the input field is not, I think.

Nevertheless I added the condition you were referring [here](https://github.com/pendulum-chain/vortex/pull/166/commits/48c29cdaf14a347d3ae3f6792bd09cc7ef675fa0).

TorstenStueber (approved):
Great. Thanks!

gianfra-t:
Hey @Sharqiewicz, we copied (most) of the sanitizing implementation from the portal as you [suggested](https://github.com/pendulum-chain/vortex/pull/166#pullrequestreview-2321606478). Thanks!

Issue #156, opened 2024-09-13
Update CTA button text to make clearer


prayagd:
### Requirement
Implement the CTA button in the following way:

1. After the user has input the amount they want to off ramp, the CTA for them reads - `Confirm`
2. On clicking the button, it changes briefly (~0.5 secs) to `Confirming` (with a buffer sign), instead of `Offramp in progress`
3. The next CTA should read `Enter Details` instead of `Start Offramping`
4. On clicking the button, it should change to `Processing Details` (with a buffer sign), instead of `Offramp in progress`

PR #180, opened 2024-09-26
Fix the main button caption

TorstenStueber:
This is a follow up to [this PR](https://github.com/pendulum-chain/vortex/pull/168), which did only partially implement the requirements of the associated issue https://github.com/pendulum-chain/vortex/issues/156, see [this comment](https://github.com/pendulum-chain/vortex/issues/156#issuecomment-2377211856).

This PR here corrects that.


ebma (approved):
LGTM, can confirm it works

Issue #120, opened 2024-08-26
Implement initial checks of various services required for a successful off ramp flow


prayagd:
### Context
There are several known reason due to which the offramping transaction using teh Vortex prototype, could fail. Therefore, it makes sense to implement a set of checks when a user lands on the app, such that if any of them fail, the user is unable to initiate the off ramping transaction.

### Requirements

1. Implement the following checks before the UI loads

- [x] Signer-service status: Confirms that both funding accounts have enough funds (the Pendulum funding account requires PEN, USDCaxl and EURC (Circle)).
- [x] Moonbeam status: Confirms that the Moonbeam executor account has sufficient funds and that the service itself works
- [x] Squidrouter status: Confirms that a subset of possible tokens the user can select to see if the app is returning any routes for bridging.
- [x] AMM pool status: Confirms if the AMM / DEX has enough liquidity for the swap transaction to be successful.
- [x] Spacewalk vault status: Confirms that the number of redeemable tokens with a single vault is larger than or equal to the amount to be offramped

2. Implement the following message / screen for the user in case any of the checks fail.

- [ ] TBC

If any of these checks where to fail, we know that the offramp will also fail, so it might be better to disable it until everything is ready. This check can be done upon loading the application and disabling the offramp button.

prayagd:
> Implement the following message / screen for the user in case any of the checks fail.


@pendulum-chain/devs What is the message we show here is the checks fail or should i create a ticket to implement that

ebma:
It shows

> Application initialization failed. Please reload, or try again later if the problem persists.



<img width="699" alt="image" src="https://github.com/user-attachments/assets/54fbe5cc-24e1-4cb7-a6f9-48b2b76dec0a">

PR #181, opened 2024-09-27
Initial checks to for avoidable errors

gianfra-t:
Issue: #120 

### Added checks
- On initialization of the app, we check if all accounts in the `signing-service` have the corresponding amounts. We also check inherently that the service is responsive.
- When the user clicks `Confirm`, we now check:
    - There exists at least one stellar vault that has redeemable tokens for the expected output of Nabla (w/ a margin, if defined) `minOutputAmount * margin`.
    - A route (squidrouter) was found.

If one of the checks fails, the `Confirm` button will be disabled, a generic message will show on top of it and the user will need to reload to perform the checks again.


signer-service/src/api/services/pendulum.service.js
```diff
@@ -56,7 +62,32 @@ exports.sendStatusWithPk = async () => {
   const { fundingAccountKeypair, fundingAmountRaw } = getFundingData(apiData.ss58Format, apiData.decimals);
   const { data: balance } = await apiData.api.query.system.account(fundingAccountKeypair.address);
 
-  if (Big(balance.free.toString()).gte(fundingAmountRaw)) {
+  const tokensToCheck = Object.keys(TOKEN_CONFIG);
```
gianfra-t, line 65: We can check all tokens in `TOKEN_CONFIG` and avoid manual definition.

signer-service/src/api/services/pendulum.service.js
```diff
@@ -56,7 +62,32 @@ exports.sendStatusWithPk = async () => {
...
+      const tokenBalanceResponse = await apiData.api.query.tokens.accounts(
+        fundingAccountKeypair.address,
+        tokenConfig.pendulumCurrencyId,
+      );
+
+      const tokenBalance = Big(tokenBalanceResponse?.free?.toString() ?? '0');
+      const maximumSubsidyAmountRaw = Big(tokenConfig.maximumSubsidyAmountRaw);
+      const subsidyRatio = tokenBalance.div(maximumSubsidyAmountRaw);
+
+      if (subsidyRatio.lt(SUBSIDY_MINIMUM_RATIO_FUND_UNITS)) {
```
gianfra-t, line 82: I've chosen a ratio between the balance and the maximum subsidy amount (defined to be `10` for now). This was mostly to avoid adding decimals to `TOKEN_CONFIG`.

signer-service/src/api/services/pendulum.service.js
```diff
@@ -56,7 +62,32 @@ exports.sendStatusWithPk = async () => {
...
   const { data: balance } = await apiData.api.query.system.account(fundingAccountKeypair.address);
 
-  if (Big(balance.free.toString()).gte(fundingAmountRaw)) {
+  const tokensToCheck = Object.keys(TOKEN_CONFIG);
+  let isTokensSufficient = true;
+
+  // Wait for all required token balances check.
+  await Promise.all(
+    tokensToCheck.map(async (token) => {
+      const tokenConfig = TOKEN_CONFIG[token];
```
ebma, line 71: Nit: We could also do 
```suggestion
    Object.entries(TOKEN_CONFIG).map(async ([token, tokenConfig]) => {
```

signer-service/src/api/services/pendulum.service.js
```diff
@@ -56,7 +62,32 @@ exports.sendStatusWithPk = async () => {
...
+      console.log(`Checking token ${token} balance...`);
+      const tokenBalanceResponse = await apiData.api.query.tokens.accounts(
+        fundingAccountKeypair.address,
+        tokenConfig.pendulumCurrencyId,
+      );
+
+      const tokenBalance = Big(tokenBalanceResponse?.free?.toString() ?? '0');
+      const maximumSubsidyAmountRaw = Big(tokenConfig.maximumSubsidyAmountRaw);
+      const subsidyRatio = tokenBalance.div(maximumSubsidyAmountRaw);
```
ebma, line 80: Maybe rename this and all related variables to something like `remainingSubsidiesAvailableCount` or similar. I think the word ratio is a bit misleading.

src/pages/swap/index.tsx
```diff
@@ -109,14 +124,44 @@ export const SwapPage = () => {
...
+    setIsInitiating(true);
+
+    const outputToken = OUTPUT_TOKEN_CONFIG[to];
+    const inputToken = INPUT_TOKEN_CONFIG[from];
+
+    // both route and stellar vault checks must be valid to proceed
+    const outputAmountBigMargin = Big(minimumOutputAmount.preciseString)
+      .round(2, 0)
+      .mul(1 + SPACEWALK_REDEEM_SAFETY_MARGIN); // add an X percent margin to be sure
+    const expecetedRedeemAmountRaw = multiplyByPowerOfTen(outputAmountBigMargin, outputToken.decimals).toFixed();
```
ebma, line 138:
```suggestion
    const expectedRedeemAmountRaw = multiplyByPowerOfTen(outputAmountBigMargin, outputToken.decimals).toFixed();
```

src/pages/swap/index.tsx
```diff
@@ -93,7 +108,7 @@ export const SwapPage = () => {
     tokenOutData.stableAmountInUnits != '' &&
     Big(tokenOutData.stableAmountInUnits).gt(Big(0));
 
-  function onSubmit(e: Event) {
+  function onConfirm(e: Event) {
```
ebma, line 111: Since this function is quite large, let's consider defining it with `useCallback()`.

src/services/squidrouter/route.ts
```diff
@@ -152,3 +150,25 @@ export async function getRouteTransactionRequest(
     data: routeResult.data,
   };
 }
+
+export async function testRoute(testingToken: InputTokenDetails, attemptedAmountRaw: string) {
```
ebma, line 154: I think it would be better if this function would actually test to get the routes for the tokens the user might want to use. Namely USDC and USDC.e -> axlUSDC. If we don't check for these specific routes, it might happen that squidrouter is able to return something for the test route but not for the actual routes we want to take.

signer-service/src/api/controllers/moonbeam.controller.js
```diff
@@ -46,3 +50,26 @@ exports.executeXcmController = async (req, res) => {
     res.status(500).send({ error: 'Internal Server Error' });
   }
 };
+
+exports.sendStatusWithPk = async () => {
+  const moonbeamExecutorAccount = privateKeyToAccount(MOONBEAM_EXECUTOR_PRIVATE_KEY);
```
ebma, line 55: Let's move this statement into the try{} block below as this can throw an error if the moonbeam private key is undefined or invalid.

ebma: If you want to make it available in the catch block as well, we would need to make it 
```
let moonbeamExecutorAccount;

try {
  moonbeamExecutorAccount = ...
```

and change the catch to
```
    return { status: false, public: moonbeamExecutorAccount?.address };
```

ebma (changes requested):
Sorry for the delayed review. Looks good overall but I left some notes.

Let's also fix the lint issues (though most of them were not introduced by this PR).

signer-service/src/api/services/pendulum.service.js
gianfra-t: True, quite confusing after I look again at it.

gianfra-t: I'll leave `SUBSIDY_MINIMUM_RATIO_FUND_UNITS` if that makes sense, since it is the actual ratio of "minimum amount of subsidies considering worse case scenario". Quite hard to find a name for it !

src/constants/constants.ts
```diff
@@ -7,6 +7,7 @@ export const PENDULUM_WSS = 'wss://rpc-pendulum.prd.pendulumchain.tech';
 //export const PENDULUM_WSS = 'ws://localhost:8000';
 export const NABLA_ROUTER = '6dQQoUKQ9LNDCrGMjoZjeHBXsuihSgQiQEgD9Z7VtHR82wfG'; // EURC circle instance
 
+export const SPACEWALK_REDEEM_SAFETY_MARGIN = 0.2;
```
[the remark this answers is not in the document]
gianfra-t: Sure, I just chose a sample value. Let's go with 5% and adjust if we see there are scenarios where this assumption is not conservative enough.

After all, the only scenario would be multiple redeems (vortex or not) happening at the same time.

src/pages/swap/index.tsx
gianfra-t: Why would you prefer it inside a `useCallback` hook? Since it depends of most of the state UI elements it will re-render probably every time anyway. You think it could affect performance if not?

src/services/squidrouter/route.ts
gianfra-t: Well the idea at least is that exactly! If you look at where `testRoute` is [called](https://github.com/pendulum-chain/vortex/pull/181/files#diff-777e0c2221f5feeef076c359496045ddd95d6d376a0fc1395b330f8a9b64a5bbR150) which is on confirm, it will perform the exact route the user is trying to offramp. The hardcoded parameters are the `to` token which is `axlUSDC` and the address, which shouldn't matter for the route. Also the amount is the one the user requires.

This is why I decided to put this check here and not when loading the app. There are too many variables (`from` and `amount`) to check all the possibilities, especially as we add more tokens. Let me know what you think.

signer-service/src/api/controllers/moonbeam.controller.js
gianfra-t: True. I was thinking that if it is undefined the service will not even start (we have that check), but it very well could be invalid.
I'll go for the second suggestion such that is easier to debug in that case.

gianfra-t: Actually, it is more or less the same since if it fails it will not be assigned anyway. So a `public = undefined` will hint that the key is invalid.

signer-service/src/constants/tokenConfig.js
```diff
@@ -6,6 +6,14 @@ const TOKEN_CONFIG = {
     vaultAccountId: '6bsD97dS8ZyomMmp1DLCnCtx25oABtf19dypQKdZe6FBQXSm',
     minWithdrawalAmount: '10000000000000',
     maximumSubsidyAmountRaw: '1000000000000', // 1 unit
+    pendulumCurrencyId: {
```
TorstenStueber, line 9: We could also avoid error-prone duplication here and determine this from `assetCode` and `assetIssuer` through some helper function – particularly once we add more currencies:

```js
      Stellar: {
        AlphaNum4: {
          code: `0x${Buffer.from(assetCode.padEnd(4, "\0")).toString("hex")}`,
          issuer: `0x${StrKey.decodeEd25519PublicKey(assetIssuer).toString("hex")}`,
        },
      },
```

src/pages/swap/index.tsx
```diff
@@ -109,14 +124,44 @@ export const SwapPage = () => {
...
+    const outputToken = OUTPUT_TOKEN_CONFIG[to];
+    const inputToken = INPUT_TOKEN_CONFIG[from];
+
+    // both route and stellar vault checks must be valid to proceed
+    const outputAmountBigMargin = Big(minimumOutputAmount.preciseString)
+      .round(2, 0)
+      .mul(1 + SPACEWALK_REDEEM_SAFETY_MARGIN); // add an X percent margin to be sure
+    const expectedRedeemAmountRaw = multiplyByPowerOfTen(outputAmountBigMargin, outputToken.decimals).toFixed();
+
+    const inputAmountBig = Big(fromAmountString);
```
TorstenStueber, line 140: At this point `fromAmountString` should not be the empty string. However, if we change the logic and it is, then this will throw an error. Use `fromAmount` returned by `useSwapForm` instead.

src/services/polkadot/spacewalk.tsx
```diff
@@ -74,7 +75,13 @@ function prettyPrintAssetInfo(assetInfo: any) {
   return assetInfo.code;
 }
 
-export async function getVaultsForCurrency(api: ApiPromise, assetCodeHex: string, assetIssuerHex: string) {
+// Filter each vault by required wrapped id and redeemable amount
+export async function getVaultsForCurrency(
+  api: ApiPromise,
+  assetCodeHex: string,
+  assetIssuerHex: string,
+  redeemableAmount: string,
```
TorstenStueber, line 83: Can we name it `redeemableAmountRaw` instead to make the type more clear?

src/services/polkadot/index.tsx
```diff
@@ -13,11 +13,14 @@ import { Extrinsic } from '@pendulum-chain/api-solang';
...
-async function createVaultService(apiComponents: ApiComponents, assetCodeHex: string, assetIssuerHex: string) {
-  const vaultsForCurrency = await getVaultsForCurrency(apiComponents.api, assetCodeHex, assetIssuerHex);
-  if (vaultsForCurrency.length === 0) {
-    throw new Error(`No vaults found for currency ${assetCodeHex}`);
-  }
+async function createVaultService(
+  apiComponents: ApiComponents,
+  assetCodeHex: string,
+  assetIssuerHex: string,
+  redeemAmount: string,
```
TorstenStueber, line 20: Can we call it `expectedRedeemAmountRaw` to make the type more clear?

src/services/squidrouter/route.ts
```diff
@@ -152,3 +150,25 @@ export async function getRouteTransactionRequest(
...
+  const { fromChainId, toChainId, axlUSDC_MOONBEAM } = squidRouterConfig;
+  const mockFromAddress = '0x2527db2c3dc99db4ab7689e1972bf7168bfe0417'; // realy shouldn't matter
```
TorstenStueber, line 156: It's probably fine to work with the mock address, but shouldn't we have the user address available at this point (the wallet should be connected)? That would make it possibly more robust.

src/pages/swap/index.tsx
```diff
@@ -109,14 +124,44 @@ export const SwapPage = () => {
...
+    const inputAmountRaw = multiplyByPowerOfTen(inputAmountBig, inputToken.decimals).toFixed();
+
+    Promise.all([
+      getVaultsForCurrency(
+        api!,
+        outputToken.stellarAsset.code.hex,
+        outputToken.stellarAsset.issuer.hex,
+        expectedRedeemAmountRaw,
+      ),
+      testRoute(fromToken, inputAmountRaw),
```
TorstenStueber, line 150: We didn't specify it but should we also add a little margin to this test and on top of `inputAmountRaw`?

TorstenStueber (commented):
Nicely done. I just have a few minor comments.

Maybe it is good to also define [this](https://github.com/pendulum-chain/vortex/pull/181/files#diff-5cd3bf0592a72c8ed25d2f160fee35a06c77d003f1a09bb057ff318bc2626e0bR30) (`'2.5'`) in `constant.js`

signer-service/src/api/controllers/moonbeam.controller.js
ebma: Thanks 👍  we still need that 
```
    return { status: false, public: moonbeamExecutorAccount?.address };
```
otherwise it might throw an error because it cannot find `address` on `undefined`.

src/pages/swap/index.tsx
ebma: Hmm right, it would have quite a few dependencies. I just thought caching with `useCallback` might make sense because of the calculations done in that function. But you're probably right that it wouldn't make much of a difference in this case.

src/services/squidrouter/route.ts
ebma: Oh I missed that, sorry! I think I was confused because the requirements in https://github.com/pendulum-chain/vortex/issues/120 all declare that these checks should happen on start of the app. But nevermind my comment then, it makes sense to check it like this 👍

src/pages/swap/index.tsx
gianfra-t: No strong opinion, just that it would make it even longer but we can do it.

signer-service/src/api/controllers/moonbeam.controller.js
gianfra-t: Of course!

src/pages/swap/index.tsx
ebma: No need, let's just leave it as is.

signer-service/src/api/services/pendulum.service.js
```diff
@@ -31,7 +37,7 @@ async function createPolkadotApi() {
 function getFundingData(ss58Format, decimals) {
   const keyring = new Keyring({ type: 'sr25519', ss58Format });
   const fundingAccountKeypair = keyring.addFromUri(PENDULUM_FUNDING_SEED);
-  const fundingAmountUnits = Big(FUNDING_AMOUNT_UNITS);
+  const fundingAmountUnits = Big(PENDULUM_FUNDING_AMOUNT_UNITS);
```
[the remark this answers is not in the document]
gianfra-t: Totally, I got confused there. Adding another variable `PENDULUM_EPHEMERAL_STARTING_BALANCE` to hold the initial funding amount.

signer-service/src/constants/tokenConfig.js
gianfra-t: Could be but I'm not so sure since across the repo we are generally using directly the `hex` representation (or both), probably since it's easier to get from the `assetRegistry`. If we add this change I would prefer to be consistent and change everywhere.

src/services/squidrouter/route.ts
gianfra-t: Yes we have it since we use this function only after the user confirms. But I thought we could use this one later in a more standalone way, perhaps in a test or at app initiation. I'll add it though.

src/pages/swap/index.tsx
gianfra-t: I'll use the same `spacewalk` margin of 5% to avoid adding more constants if that's okay.

signer-service/src/constants/tokenConfig.js
TorstenStueber: No, it's fine. Let's just leave it as it is.

TorstenStueber (approved):
Great, all done from my view. Thanks for the changes.

ebma (approved):
Great changes 👍

PR #187, opened 2024-10-02
Closing Anchor KYC window on success.

gianfra-t:
Related: [#tasks/396](https://github.com/pendulum-chain/tasks/issues/396)

Changes required to receive an event from the anchor site once the user completes the KYC flow successfully.

### DIFF
--- a/src/helpers/notifications.ts
+++ b/src/helpers/notifications.ts
@@ -2,6 +2,7 @@ import { ToastOptions, toast } from 'react-toastify';
 
 export enum ToastMessage {
   AMOUNT_MISMATCH = 'AMOUNT_MISMATCH',
+  KYC_COMPLETED = 'KYC_COMPLETED',
 }
 
 type ToastSettings = {
@@ -15,6 +16,13 @@ const ToastProperties: Record<ToastMessage, ToastSettings> = {
     options: {
       toastId: ToastMessage.AMOUNT_MISMATCH,
       type: 'error',
+    },
+  },
+  [ToastMessage.KYC_COMPLETED]: {
+    message: 'Successfully completed KYC. Offramp is being processed.',
+    options: {
+      toastId: ToastMessage.AMOUNT_MISMATCH,
+      type: 'success',
     },
   },
 };
--- a/src/helpers/notifications.ts
+++ b/src/helpers/notifications.ts
@@ -21,7 +21,7 @@ const ToastProperties: Record<ToastMessage, ToastSettings> = {
   [ToastMessage.KYC_COMPLETED]: {
     message: 'Successfully completed KYC. Offramp is being processed.',
     options: {
-      toastId: ToastMessage.AMOUNT_MISMATCH,
+      toastId: ToastMessage.KYC_COMPLETED,
       type: 'success',
     },
   },
--- a/src/hooks/useMainProcess.ts
+++ b/src/hooks/useMainProcess.ts
@@ -135,7 +135,8 @@ export const useMainProcess = () => {
           setAnchorSessionParams(anchorSessionParams);
 
           const fetchAndUpdateSep24Url = async () => {
-            const firstSep24Response = await sep24First(anchorSessionParams);
+            let firstSep24Response = await sep24First(anchorSessionParams);
+            firstSep24Response.url = firstSep24Response.url + '&callback=postMessage';
             setFirstSep24Response(firstSep24Response);
             console.log('SEP24 url:', firstSep24Response.url);
           };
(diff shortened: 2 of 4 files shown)


gianfra-t:
I know it was a research ticket yet the solution itself seemed to be easier than documenting the _why_. Since I tested it, I intended to merge it once others confirmed also on other browsers.

I'm glad that it worked on Safari also! I would test perhaps one more (Brave?) and then I would merge.

TorstenStueber:
Nice, let's also have one or two more people test (maybe @prayagd as other people will be out tomorrow due to a public holiday).

src/pages/swap/index.tsx
```diff
@@ -280,7 +305,7 @@ export const SwapPage = () => {
           <a
             href={firstSep24ResponseState.url}
             target="_blank"
-            rel="noreferrer"
+            rel="opener" //noopener forbids the use of postMessages.
```
ebma, line 308:
```suggestion
            rel="opener" //noreferrer forbids the use of postMessages.
```

ebma (approved):
I can also confirm that the automatic closing worked for me, very nice 👍 
Tested on chrome browser on macOS, so no news here. 

What do you think about showing a success-type toast message when programmatically closing the new window? Something as simple as 'KYC successful!'. As a user I feel the abrupt closing can be a bit confusing since the prototype doesn't give me any immediate feedback so I am not sure if this was intended or not. The toast would confirm that it's all good and expected.

gianfra-t:
@ebma Regarding the sudden close, I agree. We can use the toast for simplicity. It didn't feel so unexpected to me, but I'm probably biased since of course I am expect the thing to close 😆.

gianfra-t:
Toast added. Looks like this and the animation is displayed on time IMO.

image

src/helpers/notifications.ts
```diff
@@ -17,6 +18,13 @@ const ToastProperties: Record<ToastMessage, ToastSettings> = {
       type: 'error',
     },
   },
+  [ToastMessage.KYC_COMPLETED]: {
+    message: 'Successfully completed KYC. Offramp is being processed.',
+    options: {
+      toastId: ToastMessage.AMOUNT_MISMATCH,
```
ebma, line 24:
```suggestion
      toastId: ToastMessage.KYC_COMPLETED,
```

src/pages/swap/index.tsx
ebma: I wasn't saying it does, I only suggested a change for your comment because I thought you meant to mention `noreferrer` here 😅  I checked the difference between `noopener` and `noreferrer` again and they both achieve the same in this case (both would forbid using postMessage()) so nevermind my suggestion.

ebma (approved):

src/hooks/useMainProcess.ts
```diff
@@ -135,7 +135,8 @@ export const useMainProcess = () => {
           setAnchorSessionParams(anchorSessionParams);
 
           const fetchAndUpdateSep24Url = async () => {
-            const firstSep24Response = await sep24First(anchorSessionParams);
+            let firstSep24Response = await sep24First(anchorSessionParams);
+            firstSep24Response.url = firstSep24Response.url + '&callback=postMessage';
```
TorstenStueber, line 139: @gianfra-t Probably all anchors will have query parameters in their sep24 url but there is no guarantee that it will be like this (could also be more modern path parameters) and as far as I understand it is no requirement of SEP24.

So it could be better to do this defensively:

```JS
const url = new URL(firstSep24Response.url);
url.searchParams.append("callback", "postMessage");
firstSep24Response.url = url.toString();

```

src/pages/swap/index.tsx
```diff
@@ -168,12 +169,13 @@ export const SwapPage = () => {
     // Do not show any error if the user is disconnected
     if (isDisconnected) return;
 
-    if (typeof userInputTokenBalance === 'string') {
-      if (Big(userInputTokenBalance).lt(fromAmount ?? 0)) {
-        trackEvent({ event: 'form_error', error_message: 'insufficient_balance' });
-        return `Insufficient balance. Your balance is ${userInputTokenBalance} ${fromToken?.assetSymbol}.`;
-      }
-    }
+    // TESTING - REMOVE
```
TorstenStueber, line 172: @gianfra-t What about this commented section?

TorstenStueber (commented):
This has already been merged but I have two comments.

src/pages/swap/index.tsx
gianfra-t: Thanks for noticing this one, I forgot to remove after testing. Added a quick fix [here](https://github.com/pendulum-chain/vortex/pull/197).

prayagd:
@pendulum-chain/devs if i understand correctly there is no implementation ticket needed for this now right?

TorstenStueber:
Not anymore.

prayagd:
right it works, also tested on staging

PR #200, opened 2024-10-09
Create new production release

ebma:
As per request [here](https://github.com/pendulum-chain/vortex/pull/196#issuecomment-2401758642), the latest changes should be pushed to production.

@pendulum-chain/product can someone confirm with the deploy preview of this branch that offramping works as expected?


prayagd (approved):
The feature of automatic anchor tab close and all the text changes work as expected, i did not do the entire offramp as funds are low on the product account and needed for demo but we can merge.

Issue #173, opened 2024-09-24
Enable ARS on vortex UI


prayagd:
## Context
This is a single ticket for all the UI changes to be done for the ARS offramp on the Vortex UI

## Requirements
* Add ARS to the currency dropdown on the withdraw page, icon linked below
* Update the offramp fee to 2% when the user selects ARS in the dropdown
* Set a Max limit of $150 when ARS is selected as offramping currency

## Notes
ARS icon - https://static.anclap.com/coin/ars.png

prayagd:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Pendulum-6408a0365cda59c57d39bee5/issues/pendulum-chain/vortex/173?planning-poker) with Zenhub @ebma @Sharqiewicz @TorstenStueber @gianfra-t

TorstenStueber:
This has been unblocked as ARS is now available on Nabla: https://github.com/pendulum-chain/tasks/issues/401#issuecomment-2412361777

TorstenStueber:
I am closing this ticket as I merged it with https://github.com/pendulum-chain/vortex/issues/172.

Issue #172, opened 2024-09-24
Integrate ARS/Anclap into the offramp flow


prayagd:
## Context 
This ticket enables ARS offramps.

## TODO
* Add ARS to the currency dropdown on the withdraw page, icon linked below
* Extend the SEP-24 fee logic to allow for a relative and fixed fee
  * define the proper fee parameters for ARS offramps: 2% relative fee and 10 ARS fixed fee
* When user selects ARS as offramping currency and agrees to the quote
* Add ARS as an output currency and define the output token configuration accordingly
* Implement the Anclap anchor UI when the user selects ARS to offramp
* Introduce a SEP-10 master account secret configuration for the signer service
  * Use the corresponding account as the source account for SEP-10 when Anclap is used as the anchor
  * In that case securely add the SEP-10 signature on the backend

## Notes
* TOML file: https://api.anclap.com/.well-known/stellar.toml 

## Notes
ARS icon - https://static.anclap.com/coin/ars.png

gianfra-t:
@pendulum-chain/product can this be worked already?

prayagd:
> @pendulum-chain/product can this be worked already?

I will get this ready and estimated till then

prayagd:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Pendulum-6408a0365cda59c57d39bee5/issues/pendulum-chain/vortex/172?planning-poker) with Zenhub @b-yap @bogdanS98 @ebma @gianfra-t @TorstenStueber

gianfra-t:
I'll assign myself to this since I can test the flow with my bank.

TorstenStueber:
Just want to point out that ARS is not available yet on the AMM. But this should not be a blocker for this ticket here.

gianfra-t:
Yes @TorstenStueber thanks. Since the changes are very small, I'll wait for `ARS` to be "ready" to work on this and test an off-ramp.

TorstenStueber:
@gianfra-t FYI, this has been unblocked now as ARS is now available on Nabla: https://github.com/pendulum-chain/tasks/issues/401#issuecomment-2412361777

prayagd:
@gianfra-t if and when you are unblocked for testing. please record the entire flow and share the recording

prayagd:
Should we close this ticket? the linked PR is outdated

TorstenStueber:
Yes, can be closed.

PR #201, opened 2024-10-09
[OUTDATED] - Add ARS offramping support.

gianfra-t:
Issues: #173 and #172.

Besides implementing the basic Anclap integration, this PR adds a temporary endpoint for the `signer-service` to allow a single account controlled by Vortex to sign the `sep-10`, only valid for Anclap. This account must hold some ARS token before the off-ramp.


src/assets/coins/ARS.svg
ebma, line 1: This doesn't look like ARS to me 😅

src/services/anchor/index.ts
```diff
@@ -194,10 +197,20 @@ export async function sep24First(sessionParams: IAnchorSessionParams): Promise<I
...
   // at this stage, assetCode should be defined, if the config is consistent.
-  const sep24Params = new URLSearchParams({
-    asset_code: sessionParams.tokenConfig.stellarAsset.code.string,
-    amount: sessionParams.offrampAmount,
-  });
+  let sep24Params;
+
+  if (sessionParams.tokenConfig.stellarAsset.code.string === 'ARS\0') {
+    sep24Params = new URLSearchParams({
+      asset_code: sessionParams.tokenConfig.stellarAsset.code.string.replace('\0', ''),
```
ebma, line 204: Since we explicitly check for the ARS token here, I guess you can also just do
```suggestion
      asset_code: 'ARS',
```

src/services/anchor/index.ts
```diff
@@ -194,10 +197,20 @@ export async function sep24First(sessionParams: IAnchorSessionParams): Promise<I
...
+      amount: sessionParams.offrampAmount,
+      account: stellarPublic, // For Ancalap, we need to send this parameter.
```
ebma, line 206:
```suggestion
      account: stellarPublic, // For Anclap, we need to send this parameter.
```

signer-service/src/constants/tokenConfig.js
```diff
@@ -20,6 +20,14 @@ const TOKEN_CONFIG = {
...
     maximumSubsidyAmountRaw: '1000000', // 1 unit
   },
+  ars: {
+    tomlFileUrl: 'https://api.anclap.com/.well-known/stellar.toml',
+    assetCode: 'ARS\0',
+    assetIssuer: 'GCYE7C77EB5AWAA25R5XMWNI2EDOKTTFTTPZKM2SR5DI4B4WFD52DARS',
+    vaultAccountId: '6bE2vjpLRkRNoVDqDtzokxE34QdSJC2fz7c87R9yCVFFDNWs',
+    minWithdrawalAmount: '10000000000000',
+    maximumSubsidyAmountRaw: '100000000000000', // 100 unit ~ 0.1 USD @ Oct/2024
+  },
```
TorstenStueber, line 30: What about the `pendulumCurrencyId`?

src/constants/tokenConfig.ts
```diff
@@ -89,6 +89,29 @@ export const OUTPUT_TOKEN_CONFIG: Record<OutputTokenType, OutputTokenDetails> =
...
+        string: 'ARS\0',
+      },
+      issuer: {
+        hex: '0xb04f8bff207a0b001aec7b7659a8d106e54e659cdf9533528f468e079628fba1',
+        stellarEncoding: 'GCYE7C77EB5AWAA25R5XMWNI2EDOKTTFTTPZKM2SR5DI4B4WFD52DARS',
+      },
+    },
+    vaultAccountId: '6bE2vjpLRkRNoVDqDtzokxE34QdSJC2fz7c87R9yCVFFDNWs',
+    erc20WrapperAddress: '6cNENXUqHUeEGSm4psQCeykZiLXJL9VzMQnvSoouyeEEoJpe',
+    minWithdrawalAmountRaw: '100000000000000', // 100 ARS?
```
TorstenStueber, line 111: The limits according to https://api.anclap.com/transfer24/info are 11 and 500.000

TorstenStueber (commented):
It seems that Anclap has a [fixed and a percentage fee together](https://api.anclap.com/transfer24/info) (as is permitted by SEP-24), which would mean that we need to general our fee calculation. Also the token configuration cannot just have a `offrampFeesBasisPoints` but that would need to have a fixed and a relative part now.

gianfra-t:
@pendulum-chain/devs I will address your requested changes, but as we discussed in this [thread](https://satoshipay.slack.com/archives/C01V1F56RMJ/p1730482546101089?thread_ts=1730311660.344149&cid=C01V1F56RMJ), we will probably never merge this PR, all these changes will be incorporated in the PR regarding the `memo` implementation for `sep-10` (issue [here](https://github.com/pendulum-chain/vortex/issues/237)).

src/assets/coins/ARS.svg
gianfra-t: We are part of the EU now  😂. It slipped, I'll remove it.

src/services/anchor/index.ts
gianfra-t: True, good one.

signer-service/src/constants/constants.js
```diff
@@ -15,6 +15,7 @@ require('dotenv').config();
 const PENDULUM_FUNDING_SEED = process.env.PENDULUM_FUNDING_SEED;
 const FUNDING_SECRET = process.env.FUNDING_SECRET;
 const MOONBEAM_EXECUTOR_PRIVATE_KEY = process.env.MOONBEAM_EXECUTOR_PRIVATE_KEY;
+const SEP10_MASTER_SECRET = process.env.SEP10_MASTER_SECRET;
```
ebma, line 18: Let's maybe set this to the same value so that it's easy to change back if we decide to have two separate accounts in the future?
```suggestion
const SEP10_MASTER_SECRET = FUNDING_SECRET;
```

signer-service/src/constants/tokenConfig.js
gianfra-t: Was added locally, not yet pushed at the time. Pushed now.

signer-service/src/api/routes/v1/stellar.route.js
```diff
@@ -8,4 +8,9 @@ router.route('/create').post(validateCreationInput, controller.createStellarTran
 
 router.route('/payment').post(validateChangeOpInput, controller.changeOpTransaction);
 
+// TODO maybe add validator.
```
gianfra-t, line 11: These are the temporary endpoints. `GET` will return the public key, required to start the `sep-10` on the first place. `POST` will provide the signature.

I will leave this comment about validators, but I don't think is of too much value to add one now, considering these are only temporary.

signer-service/src/api/services/sep10.service.js
```diff
@@ -0,0 +1,76 @@
+const { Keypair } = require('stellar-sdk');
+const { TransactionBuilder, Networks } = require('stellar-sdk');
+const { fetchTomlValues } = require('../helpers/anchors');
+
+const { TOKEN_CONFIG } = require('../../constants/tokenConfig');
+const { SEP10_MASTER_SECRET } = require('../../constants/constants');
+
+const NETWORK_PASSPHRASE = Networks.PUBLIC;
+
+exports.signSep10Challenge = async (challengeXDR) => {
```
gianfra-t, line 10: This endpoint will receive the challenge, check the operations to ensure that only Anclap is valid, and that the to-be-authorized account is our master account.

It is of course not the safest.

src/components/FeeCollapse/index.tsx
```diff
@@ -6,9 +6,15 @@ import { useEventsContext } from '../../contexts/events';
...
 export function calculateTotalReceive(toAmount: Big, outputToken: OutputTokenDetails): string {
   const feeBasisPoints = outputToken.offrampFeesBasisPoints;
-  const fees = toAmount.mul(feeBasisPoints).div(10000).round(2, 1);
-  const totalReceive = toAmount.minus(fees).toFixed(2, 0);
-  return totalReceive;
+  const fixedFees = new Big(outputToken.offrampFeesFixedComponent ? outputToken.offrampFeesFixedComponent : 0);
+  const fees = toAmount.mul(feeBasisPoints).div(10000).add(fixedFees).round(2, 1);
+  const totalReceiveRaw = toAmount.minus(fees);
+
+  if (totalReceiveRaw.gt(0)) {
```
gianfra-t, line 13: Correct for fixed fee from Anclap.

src/services/anchor/index.ts
```diff
@@ -100,8 +113,16 @@ export const sep10 = async (
...
-  transactionSigned.sign(ephemeralKeys);
+  // only sign if not ARS token, else call the temporary
+  // signing service's endpoint
+  let maybeMasterPublic;
+  if (outputToken !== 'ars') {
+    transactionSigned.sign(ephemeralKeys);
+  } else {
+    const { masterSignature, masterPublic } = await fetchMasterSignatureSep10(transactionSigned.toXDR());
+    transactionSigned.addSignature(masterPublic, masterSignature);
+    maybeMasterPublic = masterPublic;
```
TorstenStueber, line 124: Shouldn't this always be the same public key as the `masterSep10Public` fetched above anyway? Can we reuse it to make it more clear that this must be the same key?

src/services/anchor/index.ts
```diff
@@ -194,10 +218,23 @@ export async function sep24First(sessionParams: IAnchorSessionParams): Promise<I
   const { sep24Url } = tomlValues;
 
   // at this stage, assetCode should be defined, if the config is consistent.
-  const sep24Params = new URLSearchParams({
-    asset_code: sessionParams.tokenConfig.stellarAsset.code.string,
-    amount: sessionParams.offrampAmount,
-  });
+  let sep24Params;
+
+  if (sessionParams.tokenConfig.stellarAsset.code.string === 'ARS\0') {
```
TorstenStueber, line 223: Let's also rather use a flag in the token config instead of a hard coded condition here.

src/services/anchor/index.ts
```diff
@@ -194,10 +218,23 @@ export async function sep24First(sessionParams: IAnchorSessionParams): Promise<I
...
+    if (!masterPublic) {
+      throw new Error('Master must be defined at this point.');
+    }
+    sep24Params = new URLSearchParams({
+      asset_code: 'ARS',
+      amount: sessionParams.offrampAmount,
+      account: masterPublic, // Since we signed with the master from the service, we need to specify the corresponding public here
```
TorstenStueber, line 230: Shouldn't this be the ephemeral account instead? The SEP-24 standard defines `account` as:

> The Stellar or muxed account the client will use as the source of the withdrawal payment to the anchor. Defaults to the account authenticated via SEP-10 if not specified.

signer-service/src/api/services/sep10.service.js
```diff
@@ -0,0 +1,76 @@
...
+  }
+  if (!firstOp.value || firstOp.value.length !== 64) {
+    throw new Error('First manageData operation should have a 64-byte random nonce as value');
+  }
+
+  // Flags to check presence of required operations
+  let hasWebAuthDomain = false;
+
+  // Verify extra manage_data operations
+  for (let i = 1; i < operations.length; i++) {
```
Sharqiewicz, line 48: Why not use operations.map() instead of a for loop?

src/services/anchor/index.ts
```diff
@@ -77,9 +79,20 @@ export const sep10 = async (
   const NETWORK_PASSPHRASE = Networks.PUBLIC;
   const ephemeralKeys = Keypair.fromSecret(stellarEphemeralSecret);
   const accountId = ephemeralKeys.publicKey();
-  const urlParams = new URLSearchParams({
-    account: accountId,
-  });
+
+  let urlParams;
+  if (outputToken === 'ars') {
```
Sharqiewicz, line 84: I think refactoring this to a function and then `const urlParams = theFunction()` would improve readability and reusability

signer-service/src/constants/tokenConfig.js
```diff
@@ -20,6 +20,22 @@ const TOKEN_CONFIG = {
     decimals: 6,
     maximumSubsidyAmountRaw: '1000000', // 1 unit
   },
+  ars: {
+    tomlFileUrl: 'https://api.anclap.com/.well-known/stellar.toml',
+    assetCode: 'ARS\0',
+    assetIssuer: 'GCYE7C77EB5AWAA25R5XMWNI2EDOKTTFTTPZKM2SR5DI4B4WFD52DARS',
+    vaultAccountId: '6bE2vjpLRkRNoVDqDtzokxE34QdSJC2fz7c87R9yCVFFDNWs',
+    minWithdrawalAmount: '10000000000000',
```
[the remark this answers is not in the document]
gianfra-t: You're right. I will just leave it to `11` for consistency. We may need it soon.

signer-service/src/api/services/sep10.service.js
```diff
@@ -0,0 +1,76 @@
...
+    throw new Error('The first operation should be manageData');
+  }
+  // We only want to accept a challenge that would authorize the master key.
+  if (firstOp.source !== masterStellarKeypair.publicKey()) {
+    throw new Error('First manageData operation must have the master signing key as the source');
+  }
+
+  // Purposely hardcode Anclap's key, we don't want to sign this type of transaction
+  // for Mykobo
+  const expectedKey = `api.anclap.com auth`;
```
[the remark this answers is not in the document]
gianfra-t: To make it generic for future sep-10 with memos, you mean? Sure, I hard coded it to save time, but I can make it generic.

src/services/anchor/index.ts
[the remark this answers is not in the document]
gianfra-t: Yes, totally.

src/services/anchor/index.ts
gianfra-t: Actually, good catch. No need to return it along with the signature, we can use the same as the one above.

src/services/anchor/index.ts
gianfra-t: Yes, you are right. But it doesn't work with it. It only works with the one that did the `sep-10`. I missed that part of the docs and found this by trial an error.

signer-service/src/api/services/sep10.service.js
gianfra-t: I found the explicit syntax more clear for this case such that it's visible what are we filtering for, but we could add a generic `verifyOperation` with all the rules and map over that, WDYT?
And return the first error it encounters.

src/services/anchor/index.ts
gianfra-t: Alright! I will add it also including this [comment](https://github.com/pendulum-chain/vortex/pull/201#discussion_r1829797828).

signer-service/src/api/services/sep10.service.js
```diff
@@ -0,0 +1,76 @@
...
+  const transactionSigned = new TransactionBuilder.fromXDR(challengeXDR, NETWORK_PASSPHRASE);
+  if (transactionSigned.source !== anchorSigningKey) {
+    throw new Error(`Invalid source account: ${transactionSigned.source}`);
+  }
+  if (transactionSigned.sequence !== '0') {
+    throw new Error(`Invalid sequence number: ${transactionSigned.sequence}`);
+  }
+
+  // Verify manage_data operations
+  const operations = transactionSigned.operations;
```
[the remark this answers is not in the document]
Sharqiewicz: wdyt?

src/services/stellar/index.tsx
```diff
@@ -188,7 +191,7 @@ async function createOfframpAndMergeTransaction(
       throw new Error(`Unexpected offramp memo type: ${memoType}`);
   }
 
-  const stellarAsset = new Asset(code.string, issuer.stellarEncoding);
+  const stellarAsset = new Asset(code.string.replace('\0', ''), issuer.stellarEncoding);
```
Sharqiewicz, line 194: If we're using string.replace('\0', '') or assetCode.replace('\0', '') in multiple places, perhaps we should extract it into a separate function. This would enhance both readability and reusability. What do you think?

signer-service/src/api/services/sep10.service.js
Sharqiewicz: I think writing a function with a clear name and adding it to .map() can make the code more readable and self-documenting. Using a function `verifyOperation` with `.map` makes it easier to follow what is happening. wdyt?

gianfra-t: The issue is that it will require quite a lot of context and although it is easier to read, the control flow becomes harder as well as the errors returns. But let's see, I'll try to see how it looks.

src/services/anchor/index.ts
```diff
@@ -100,8 +113,16 @@ export const sep10 = async (
   }
 
   // More tests required, ignore for prototype
-
-  transactionSigned.sign(ephemeralKeys);
+  // only sign if not ARS token, else call the temporary
+  // signing service's endpoint
+  let maybeMasterPublic;
```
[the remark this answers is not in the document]
gianfra-t: Yup! Improved [here](https://github.com/pendulum-chain/vortex/pull/201/commits/29a9147614fb8e8e01f253d0d490292bc209f2f9#diff-e8e6c9ebebea2e994a257e9f9c83bb19da16efb95e46255edfd5837a1f2a2f99R113). The `sep10Account` variable now holds either the ephemeral or the "client master".

signer-service/src/api/helpers/anchors.js
```diff
@@ -0,0 +1,27 @@
+const fetchTomlValues = async (tomlFileUrl) => {
+  const response = await fetch(tomlFileUrl);
+  if (response.status !== 200) {
+    throw new Error(`Failed to fetch TOML file: ${response.statusText}`);
+  }
+
+  const tomlFileContent = (await response.text()).split('\n');
+  const findValueInToml = (key) => {
```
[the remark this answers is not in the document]
gianfra-t: I think it's simplified now!

gianfra-t:
@pendulum-chain/devs as we discussed in the sync and on slack, this solution (signing with a fixed/master account always) doesn't work anymore and will probably stay that way. Nevertheless, many of the changes (like adding ARS icon, config, toml, etc) will remain.  
This [PR](https://github.com/pendulum-chain/vortex/pull/236) will be merged and has included the integration (with ephemeral) and many of the requested changes.

gianfra-t:
I will close this PR to avoid confusion. TL;DR:
- The solution of a single account is incorrect/incomplete.
- The remaining part of the integration was done [here](https://github.com/pendulum-chain/vortex/pull/236), which also addresses the comments.

gianfra-t:
I will close this PR to avoid confusion. TL;DR:
- The solution of a single account is incorrect/incomplete.
- The remaining part of the integration was done [here](https://github.com/pendulum-chain/vortex/pull/236), which also addresses the comments.

PR #208, opened 2024-10-10
Fill the input field when clicked on balance

Sharqiewicz:
🦈 🟢 ready
---
- [x] fix hover on select token
- [x] implement click on balance to set max amount


src/pages/swap/sections/BankDetails.tsx
```diff
@@ -16,11 +16,16 @@ export const BankDetails: FC<BankDetailsProps> = ({ registerBankAccount, registe
     <p className="my-5 font-thin text-center text-gray-400">Which bank account should we send the funds to?</p>
     <LabeledInput
       label="PIX ID"
```
TorstenStueber, line 18: This whole component is not even used anymore. We should be able to remove the whole file.

TorstenStueber (approved):
👍 

Code is all good, just that one component can be removed.

src/pages/swap/sections/BankDetails.tsx
gianfra-t: What is this component about?

prayagd (approved):
LGTM works as expected

ebma (approved):

PR #209, opened 2024-10-11
Bug: UI signing box not properly in the foreground

Sharqiewicz:
- [x] Fix z-index of SigningBox
- [x] Fix empty space between footer and main sections of SigningBox


TorstenStueber (approved):
Can confirm that it works.

PR #213, opened 2024-10-14
Add percentage view to success screen

### LINKED ISSUE #192: Update the progress screen text
## Context
Currently on the progress screen, 
- the user is asked to not close the screen but it could be improved saying to leave the window open and check later
- the user is also shown the steps 1 to 11, which arent very helpful to the user. Instead we show a progress bar with %

## Requirement
* Update the text "Do not close this tab" to “Leave the window open and check back in 5 minutes”
* Remove the 1 to 11 steps and replace it with % bar
* As there are 11 steps, completion of each step should show a 9% increase on the loading bar (rounding off this number) reaching to 100%
* Show this bar below the text "Your transaction is in progress"
* (Torsten) I added more requirements [below](https://github.com/pendulum-chain/vortex/issues/192#issuecomment-2407145229)

## Note
For reference the screen to update text
Screenshot 2024-10-07 at 11 52 45 AM

Loading bar for reference
Screenshot 2024-10-10 at 11 08 05 AM

TorstenStueber:
Closes #192

### DIFF
--- a/src/pages/progress/index.tsx
+++ b/src/pages/progress/index.tsx
@@ -3,21 +3,35 @@ import { ExclamationCircleIcon } from '@heroicons/react/20/solid';
 import { OfframpingPhase, OfframpingState } from '../../services/offrampingFlow';
 import { Box } from '../../components/Box';
 import { BaseLayout } from '../../layouts';
+import { INPUT_TOKEN_CONFIG, OUTPUT_TOKEN_CONFIG } from '../../constants/tokenConfig';
 
-const OFFRAMPING_PHASE_MESSAGES: Record<OfframpingPhase, string> = {
-  prepareTransactions: 'Preparing transactions',
-  squidRouter: 'Bridging assets via Axelar',
-  pendulumFundEphemeral: 'Creating Pendulum ephemeral account',
-  executeXCM: 'Bridging assets via XCM',
-  subsidizePreSwap: 'Compensating swap risk',
-  nablaApprove: 'Approving Forex AMM',
-  nablaSwap: 'Swapping on Forex AMM',
-  subsidizePostSwap: 'Compensating swap risk',
-  executeSpacewalkRedeem: 'Bridging assets via Spacewalk',
-  pendulumCleanup: 'Cleaning up Pendulum ephemeral account',
-  stellarOfframp: 'Offramping on Stellar',
-  stellarCleanup: 'Cleaning up Stellar ephemeral account',
-};
+function createOfframpingPhaseMessage(offrampingState: OfframpingState): string {
+  const inputToken = INPUT_TOKEN_CONFIG[offrampingState.inputTokenType];
+  const outputToken = OUTPUT_TOKEN_CONFIG[offrampingState.outputTokenType];
+
+  switch (offrampingState.phase) {
+    case 'prepareTransactions':
+    case 'squidRouter':
+    case 'pendulumFundEphemeral':
+      return `Bridging ${inputToken.assetSymbol} from Polygon --> Moonbeam`;
+    case 'executeXCM':
+      return `Transferring ${inputToken.assetSymbol} from Moonbeam --> Pendulum`;
+    case 'subsidizePreSwap':
+    case 'nablaApprove':
+    case 'nablaSwap':
+    case 'subsidizePostSwap':
+      return `Swapping to ${outputToken.stellarAsset.code.string} on Vortex DEX`;
+    case 'executeSpacewalkRedeem':
+      return `Bridging ${outputToken.stellarAsset.code.string} to Stellar via Spacewalk`;
+    case 'pendulumCleanup':
+    case 'stellarOfframp':
+    case 'stellarCleanup':
+      return 'Transferring to local partner for bank transfer';
+
+    default:
+      return '';
+  }
+}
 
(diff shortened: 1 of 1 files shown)


src/pages/progress/index.tsx
```diff
@@ -48,26 +53,101 @@ const WarningSection: FC = () => (
...
-);
+const ProgressContent: FC<{ currentPhase: OfframpingPhase; currentPhaseIndex: number }> = ({
+  currentPhase,
+  currentPhaseIndex,
+}) => {
+  const [currentPercentage, setCurrentPercentage] = useState<number>(
+    Math.round((100 / numberOfPhases) * currentPhaseIndex),
+  );
+  const [targetPhaseIndex, setTargetPhaseIndex] = useState<number>(currentPhaseIndex);
+  const currentPhaseRef = useRef<OfframpingPhase>(currentPhase);
```
TorstenStueber, line 64: I had to define some variables as refs instead of state so that the interval that is created inside the `useEffect` would not need to be recreated whenever anyone of the variables changes. This way the `useEffect` hook has an empty dependency array.

TorstenStueber: Only the `currentPercentage` and `targetPhaseIndex` are state variables because changing them needs to trigger a rerender (either show another percentage or start a new transition animation).

src/pages/progress/index.tsx
```diff
@@ -1,8 +1,9 @@
-import { FC } from 'preact/compat';
+import { FC, useEffect, useState } from 'preact/compat';
 import { ExclamationCircleIcon } from '@heroicons/react/20/solid';
 import { OfframpingPhase, OfframpingState } from '../../services/offrampingFlow';
 import { Box } from '../../components/Box';
 import { BaseLayout } from '../../layouts';
+import { useRef } from 'react';
```
ebma, line 6: Nit: import from `preact/compat` instead

src/pages/progress/index.tsx
```diff
@@ -48,26 +53,101 @@ const WarningSection: FC = () => (
...
+    Math.round((100 / numberOfPhases) * currentPhaseIndex),
+  );
+  const [targetPhaseIndex, setTargetPhaseIndex] = useState<number>(currentPhaseIndex);
+  const currentPhaseRef = useRef<OfframpingPhase>(currentPhase);
+  const currentPhaseIndexRef = useRef<number>(currentPhaseIndex);
+  const currentPhaseStartPercentage = useRef<number>((100 / numberOfPhases) * currentPhaseIndex);
+  const expectedEndTimeOfPhase = useRef<number>(Date.now() + OFFRAMPING_PHASE_SECONDS[currentPhase] * 1000);
+
+  useEffect(() => {
+    const intervalId = setInterval(() => {
```
ebma, line 70: Should we define at least a small timeout for the interval?

ebma (approved):
Looking great 🙏 I tested it locally by making the offramping phases advance in fixed time intervals without actually executing the related functions and I can confirm that it works as expected, ie. the progress bar is catching up if phases complete earlier than expected or waits if it takes longer than expected.

src/pages/progress/index.tsx
TorstenStueber: This would be necessary if the flow completely stalls although there is no error (if there is an error, the page will automatically reload after 30 seconds or it will transition to the failure screen). In that case we could merely ensure that the user's browser will not require any more CPU, is that the purpose of this suggestion?

src/pages/progress/index.tsx
TorstenStueber: Thanks, absolutely.

src/pages/progress/index.tsx
ebma: Yes I was just concerned about the performance. It's probably okay to set the `timeout` to 100 or even 500 ms and still get a smooth transition. I'm referring to the `setInterval(f, timeout)` field if that's not clear.

prayagd:
Please dont merge until product reviews this

ebma (approved):
The code looks good to me but I didn't test the flow again.

TorstenStueber:
@pendulum-chain/product can you review this please?

prayagd:
@TorstenStueber All good, just one small change to add text “this usually takes 6-8 mins” under "Your transaction is in Progress"

prayagd:
I tested the entire flow, here is the [video](https://www.loom.com/share/82db6805ac374e54981fbd6518ea3001?sid=497f51a5-a6d1-4a00-8211-94559b6b6d7e)

TorstenStueber:
@prayagd since this is not part of the designs (?), any preference for text color, font size, font weight? And should I place it between "Your transaction is in Progress" and the text describing the current step, or below that one?

prayagd:
> And should I place it between "Your transaction is in Progress" and the text describing the current step

Yes

> any preference for text color, font size, font weight?

Same as the "Your transaction is in Progress" but just dont make it bold

prayagd:
@TorstenStueber can we make the quick fix and close this?

prayagd:
#229 so we can push it

prayagd (approved):

TorstenStueber:
@prayagd I pushed the change.

prayagd:
Thank you

PR #219, opened 2024-10-16
Updating the support text and support link

prayagd:
- Updating as all users are not expected to have TG so we are adding a form link for support


prayagd:
> Next time, please open the PR directly from this repo without forking. 
commited the same mistake

prayagd:
fixed [here](https://github.com/pendulum-chain/vortex/pull/220)

PR #220, opened 2024-10-16
updated tg link to support link

prayagd:
- Update the TG link to support link as all users are not expected to TG

### DIFF
--- a/src/config/index.ts
+++ b/src/config/index.ts
@@ -62,5 +62,5 @@ export const config = {
     mockSep24: false,
     overwriteMinimumTransferAmount: false,
   },
-  telegramUrl: 'https://t.me/vortex_fi',
+  telegramUrl: 'https://forms.gle/bgH4XTTbQ3YbwQ3t7',
 };
--- a/src/config/index.ts
+++ b/src/config/index.ts
@@ -62,5 +62,5 @@ export const config = {
     mockSep24: false,
     overwriteMinimumTransferAmount: false,
   },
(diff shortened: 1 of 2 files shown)


src/config/index.ts
```diff
@@ -62,5 +62,5 @@ export const config = {
     mockSep24: false,
     overwriteMinimumTransferAmount: false,
   },
-  telegramUrl: 'https://t.me/vortex_fi',
+  telegramUrl: 'https://forms.gle/bgH4XTTbQ3YbwQ3t7',
```
ebma, line 65: Nit: let's rename this to `supportUrl` then, if this is not related to telegram anymore.
```suggestion
  supportUrl: 'https://forms.gle/bgH4XTTbQ3YbwQ3t7',
```

src/pages/failure/index.tsx
```diff
@@ -52,9 +52,9 @@ export const FailurePage = ({ finishOfframping, continueFailedFlow, transactionI
           </p>
         )}
         <p className="mb-6 text-center text-gray-400">
-          Contact our support team on{' '}
+          Contact our{' '}
           <a href={config.telegramUrl} target="_blank" rel="noreferrer" className="underline">
```
ebma, line 56:
```suggestion
          <a href={config.supportUrl} target="_blank" rel="noreferrer" className="underline">
```

prayagd:
@ebma suggestions committed

src/config/index.ts
TorstenStueber: Good point, it is not even telegram now anymore but a Google form.

PR #221, opened 2024-10-18
Add call to switch chain to submission handler

### LINKED ISSUE #218: Allows txn signing from unsupported network
## Context
If i am existing user of vortex(already connected by wallet earlier) and if i am not on the polygon network, the dApp allows the user to create a request from other network and when it comes to signing it shows a wrong network signing details.

## Expected outcome
* If the user is on not supported network apart from polygon
* Before the user accepts the quote, check if they are on Polygon if not prompt to change the network

## Reference 
image

ebma:
Adds a call to switch to polygon to the submission handler. If the user is already on Polygon, nothing happens and the flow continues as usual. The documentation for `isSwitchchain()` [here](https://wagmi.sh/react/api/hooks/useSwitchChain).

Closes #218.

### DIFF
--- a/src/constants/constants.ts
+++ b/src/constants/constants.ts
@@ -12,6 +12,3 @@ export const TRANSFER_WAITING_TIME_SECONDS = 6000;
 export const SIGNING_SERVICE_URL =
   config.maybeSignerServiceUrl ||
   (config.isProd ? 'https://prototype-signer-service-polygon.pendulumchain.tech' : 'http://localhost:3000');
-
-// The chain id for polygon is 137, see [here](https://wagmi.sh/react/api/chains#available-chains)
-export const POLYGON_CHAIN_ID = 137;
--- a/src/constants/constants.ts
+++ b/src/constants/constants.ts
@@ -12,3 +12,6 @@ export const TRANSFER_WAITING_TIME_SECONDS = 6000;
 export const SIGNING_SERVICE_URL =
   config.maybeSignerServiceUrl ||
   (config.isProd ? 'https://prototype-signer-service-polygon.pendulumchain.tech' : 'http://localhost:3000');
+
+// The chain id for polygon is 137, see [here](https://wagmi.sh/react/api/chains#available-chains)
+export const POLYGON_CHAIN_ID = 137;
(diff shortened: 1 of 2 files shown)


TorstenStueber:
The code looks good. However, when using with Phantom Wallet on Firefox, I still get the "Wrong network" in formation in the button. Probably it switches to the correct network once I sign, but why don't we do switch already at startup of the app?

Screenshot 2024-10-18 at 08 06 20

ebma:
#218 defines the expected outcome

> - If the user is on not supported network apart from polygon
> - Before the user accepts the quote, check if they are on Polygon if not prompt to change the network

And this is how I implemented it. It will only prompt to change the network when the user clicks on 'Confirm'. I think it's okay to leave it like that, considering that if a first-time user clicks on 'Connect wallet' and has the wallet extension set for a different network, they will also get prompted to switch to the polygon network if they are not using that already. That's covered by the library we use for managing the wallet connection. So the 'Wrong network' can only happen for users that already connected once and then come back to the vortex app.

gianfra-t:
Tested and with Metamask works as expected. With Talisman also works but silently changes the network on the background.

src/constants/constants.ts
```diff
@@ -12,3 +12,6 @@ export const TRANSFER_WAITING_TIME_SECONDS = 6000;
 export const SIGNING_SERVICE_URL =
   config.maybeSignerServiceUrl ||
   (config.isProd ? 'https://prototype-signer-service-polygon.pendulumchain.tech' : 'http://localhost:3000');
+
+// The chain id for polygon is 137, see [here](https://wagmi.sh/react/api/chains#available-chains)
+export const POLYGON_CHAIN_ID = 137;
```
Sharqiewicz, line 17: @ebma 
What about:
import { polygon } from 'viem/chains';

polygon.id

Sharqiewicz: or even
import { polygon } from 'wagmi/chains';

polygon.id

TorstenStueber (approved):
@ebma Okay, understood, I will approve this PR and move my suggestion somewhere else. Thanks!

src/constants/constants.ts
ebma: Nice one. Changed.

TorstenStueber (approved):

PR #222, opened 2024-10-18
Add additional events to track on vortex app

### LINKED ISSUE #198: Add additional events to track on Vortex app
### Context
After the first version of the tracking went live, we want to gain additional insights of the user behaviour during their journey.

### Requirement

- [ ] Track the following actions / events on Google Analytics:

1. User clicks 'Enter details' which opens MYKOBO journey
2. The 2 signatures are provided to the user - would tell us if both the signatures were triggered
3. The 2 signatures are signed by the user - would tell us if both the transactions were signed by the user (if it was shown to them)
4. The user is on the Progress page
5. Each step in the background - Would tell us where in the flow the user dropped / closed the tab (in case they do)
6. Submission of email on transaction successful screen
7. Submit email on transaction unsuccessful screen

- [ ] Suggest the events / parameters structure for the above

### LINKED ISSUE #212: Bug: UseEffect called conditionally
The final `useEffect` hook on `src/pages/swap/index.tsx` is called conditionally, i.e., there are executions of `SwapPage` where it will not be executed. React Hooks must be called in the exact same order in every component render.

## TODO
Move the `useEffect` before any conditions that could stop the execution early.

ebma:
- [x] Add handling for new events to the code
- [x] Configure new events in GTM

### Events 
- GA4 Event | email_submission
- GA4 Event | signing_requested
- GA4 Event | transactions_signed
- GA4 Event | progress
- GA4 Event | kyc_started

### Triggers
- DL Event | email_submission
- DL Event | signing_requested
- DL Event | transactions_signed
- DL Event | progress
- DL Event | kyc_started

### Variables
- DLV | phase (used for tracking of the progress)
- DLV | transaction_status (used for tracking of the email submission)

--- 
Note: #198 defines the requirement

> The 2 signatures are provided to the user - would tell us if both the signatures were triggered

This is not easy to track. I implemented it so that the `signing_requested` event is emitted after the first transaction signature was requested from the user. 

Closes #198 and closes #212.


prayagd:
@vadaynujra can you please review this?

src/contexts/events.tsx
```diff
@@ -35,13 +39,30 @@ export interface WalletConnectEvent {
...
+    | 'transaction_failure';
   from_asset: string;
   to_asset: string;
   from_amount: string;
   to_amount: string;
 }
 
+export interface ProgressEvent {
+  event: 'progress';
+  phase: number;
```
gianfra-t, line 58: Could we also send the phase message or key? Perhaps easier to keep track than the integer alone.

src/services/squidrouter/process.ts
```diff
@@ -40,6 +41,8 @@ export async function squidRouter(
   } catch (e) {
     console.error('Error in squidRouter: ', e);
     return { ...state, failure: 'unrecoverable' };
+  } finally {
+    trackEvent(createTransactionEvent('signing_requested', state));
```
gianfra-t, line 45: That's the thing we could not differentiate between a sign rejection, or an error. We can assume that if we reach the next event it was successful and if not, it was either an error or was rejected.

gianfra-t (approved):
Looks good to me ✅.

src/contexts/events.tsx
ebma: We could but do you think it would improve the tracking? 🤔 I would imagine that the index is enough, and it's also easier to visualize than keys/strings. I can add it, no problem. But I'm not sure if it would be used in practice.

src/services/squidrouter/process.ts
ebma: Hmm well now that you mention it, the catch statement returns from this function in l. 43 so the `finally` part will only be reached if the `try` block doesn't fail. Maybe I should duplicate the code and put the call to `trackEvent` below the call to `await writeContract()` ie. between l. 40-41 as well as above the return in l. 42-43. What do you think @gianfra-t?

src/contexts/events.tsx
gianfra-t: I think it only makes sense when manually exploring the events and avoid having to look for what phase does this particular integer represent. No idea how this will be used for analytics, probably in that case it doesn't make sense.

src/services/squidrouter/process.ts
gianfra-t: Mmm.. I think the `finally` statement will execute regardless! Not sure if it will do so before or after returning, but it should execute it. in that sense I think it is fine.

Assuming it is tracked on both scenarios, what I meant is that the user may not have seen the request but still it will be equivalent as if it was requested but rejected, judging from our events only. But probably it is threading too thin 😅.

TorstenStueber: The `finally` will always execute. Either after the `return` of the `catch` block before it actually returns or after a successful execution of the `try` block.

TorstenStueber (approved):
All the code is fine the way it is. Just minor comments.

Something to clarify: the ticket reads "The 2 signatures are provided to the user" and "The 2 signatures are signed by the user".

This is ambiguous: you track the first transaction provided for signing and the second transaction signed. I interpret this more that we are supposed to track 4 events: `signing_requested` and `transaction_signed` for each transaction.

So maybe it is better to track `signing_requested` before `writeContract` (for the approve) and before `sendTransaction` and then track `transaction_signed` only when the `try` block is successful.

ebma:
>  I interpret this more that we are supposed to track 4 events: signing_requested and transaction_signed for each transaction.
So maybe it is better to track signing_requested before writeContract (for the approve) and before sendTransaction and then track transaction_signed only when the try block is successful.

How would you change the event structure though? I wanted to group these events with the others in the `TransactionEvent` because they are a necessary part of the flow. In contrast to the email submission which is optional. I can split the `signing_requested` and `transactions_signed` into separate events which also have an index attribute that indicates which of the transactions was signed (ie. 1 or 2).

TorstenStueber:
Yes, it's a good idea to just add the index attribute. @vadaynujra how to interpret the requirement and would adding the index attribute address that?

vadaynujra:
@ebma `signing_requested` and `transactions_signed` as separate events, with 1 and 2 as attributes, would be fine as it would allow us to track them as 4 separate yet mandatory events as correctly interpreted by @TorstenStueber

ebma:
I added the requested changes, and changed the GTM config as follows

## Changed Tags
- Rename `GA4 Event | transactions_signed` to `GA4 Event | transaction_signed`
- Add event parameter `name` to `GA4 Event | progress`
- Remove `{{Transaction Event Variables}}` from both `GA4 Event | signing_requested` and `GA4 Event | transaction_signed`
- Add event parameter `index` to both `GA4 Event | signing_requested` and `GA4 Event | transaction_signed`

### New variables
- Add variable `DLV | progress name`
- Add variable `DLV | index`

Note that the `index` for the signed transactions are `1` and `2` instead of `0` and `1`. I thought this way makes it easier to interpret in GA but we can also change it to the more technical indexing.

TorstenStueber (approved):
All good!

I think your change also resolved https://github.com/pendulum-chain/vortex/issues/212

gianfra-t (approved):
Looks good 👍.

Issue #225, opened 2024-10-21
Add client_domain to the information sent to the anchor


prayagd:
### Context
In order to realise the terms of the agreement with anchors i.e. reduced fees, increased limits for off-ramp (and possibly more), there must be a way for the anchor to be able to identify transactions / users coming from Vortex.

Our client domain is `satoshipay.io`.

### Requirement

- [ ] Send the client signature with client_domain ID for SatoshiPay / Vortex to the anchors.

gianfra-t:
Is this blocked until we get responses from Mykobo and Anclap? Or shall we explore this regardless.

vadaynujra:
@gianfra-t from my understanding this has to be done for MYKOBO regardless. For Anclap it's dependent on them agreeing to it. Right @TorstenStueber ?

ebma:
For the client_domain, we need to set up a stellar.toml with metadata about our organization, similar to what Mykobo did [here](https://circle.anchor.mykobo.co/.well-known/stellar.toml). I don't think we have that yet, at least I didn't find one on our validator [here](http://stellar-de-fra.satoshipay.io:11625/.well-known/stellar.toml).

The client domain is explained [here](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0010.md#verifying-the-client-domain). The structure of stellar.toml files is explained [here](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0001.md) and the important field we need is the `SIGNING_KEY`. 

@pendulum-chain/product you can help us fill some of the fields already. It would be nice if you can think of what to put for the 'organization' parameters [here](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0001.md#organization-documentation). You can draw inspiration from what Mykobo does of course.

The protocol doesn't say which fields are obligatory and which are not but I think for the 'General information' it would be enough to have `VERSION`, `NETWORK_PASSPHRASE` and `SIGNING_KEY`.

TorstenStueber:
We have a stellar.toml file, it is located at `https://satoshipay.io/.well-known/stellar.toml` and its content is:

```
VERSION="2.0.0"

NETWORK_PASSPHRASE="Public Global Stellar Network ; September 2015"

ACCOUNTS=[
  "GBJQUIXUO4XSNPAUT6ODLZUJRV2NPXYASKUBY4G5MYP3M47PCVI55MNT",
[...]

PR #236, opened 2024-10-28
Add client domain corresponding signature when needed.

gianfra-t:
Original issue: #225. 

This PR also adds partially ARS integration: #172 and #173. Partially because given a UI particularity of Anclap, the ephemeral cannot perform the offramp because it doesn't have ARS token when created.

### Main addition

Adds a new endpoint to the signer service, which upon request, will sign the corresponding challenge from the anchor. Also modifies the UI logic to add the extra `client_domain` parameter, and append the signature to the response when required.

### Security

Although the challenge transaction is sent by the UI, we perform corresponding challenge operations checks.


signer-service/src/api/services/stellar.service.js
```diff
@@ -151,4 +152,21 @@ async function sendStatusWithPk() {
...
 
-module.exports = { buildCreationStellarTx, buildPaymentAndMergeTx, sendStatusWithPk };
+async function signSep10Challenge(challengeXDR, outToken) {
+  const keypair = Keypair.fromSecret(CLIENT_SECRET);
+
+  const { signingKey } = await fetchTomlValues(TOKEN_CONFIG[outToken].tomlFileUrl);
+
+  const transactionSigned = new TransactionBuilder.fromXDR(challengeXDR, NETWORK_PASSPHRASE);
+  if (transactionSigned.source !== signingKey) {
+    throw new Error(`Invalid source account: ${transactionSigned.source}`);
```
gianfra-t, line 162: We are fetching the `signingKey` of the anchor from our own `.toml` url. I see no vector from which someone could send a malicious transaction for us to sign. 

This, in conjunction with the `0` sequence number check.

gianfra-t:
@pendulum-chain/devs added to the security concern, we of course have this endpoint open, so although we will not sign any transaction that would withdraw funds, any person could request this signature and act in our regard as Client Application.

We discussed at some point for the `subsidized` endpoints that we could implement checks that the user has indeed transferred to the `squidrouter` contract, for it to be enabled.

This case will be harder to protect, since this operation is performed at the beginning and there is no guarantee that the user will go through with the off-ramp with us. 

If we then reverse the order of operations and make the user first sign the `commitment` transactions to swap, before performing the `sep-10`, then we run the risk that the funds will get stuck because the user did not or could not complete the KYC.

Therefore the best option I see is to modify the code such that the UI first signs the `sep-10` and later the backend sends the challenge response, potentially relaying back the `jwt` to the UI if we think this is safe, I believe that for a `sep-24` should be, but we need to be sure that this `jwt` cannot be used for other endpoints (such as a `sep-12`) to identify users at will (by choosing the unique `memo`).

gianfra-t:
Following on the above, I think the `JWT` is safe judging by the [challenge](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0010.md#success), which has in the transaction operations the description of what account acts as client, client domain.

TorstenStueber (commented):
Let's think about the attack scenario and solutions here. But I don't think that we need to implement them yet (more below).

So what is the attack vector. I think it is just that someone could impersonate us by letting their own sep-10 challenge be signed by us. There is no way to prevent that with this endpoint as the endpoint can't know who created the challenge.

To make this secure, I think that the endpoint itself needs to request the sep-10 challenge. So the frontend does not request it directly but asks the service to request it. The backend can trust the anchor and the frontend can trust the backend.

However, this is still not safe as the backend can't trust the frontend. So I think that in order to make it secure, the backend should never return the resulting sep-10 token to the frontend. This means that the backend needs to initiate sep-24 as well. Would this then prevent the attack scenario? I think so.

Anyway, this could be too much effort for now. I think we should still keep it simple. Let's rather maintain a list of known weaknesses and solutions and prioritize them as required.

signer-service/src/api/routes/v1/stellar.route.js
```diff
@@ -8,4 +8,6 @@ router.route('/create').post(validateCreationInput, controller.createStellarTran
 
 router.route('/payment').post(validateChangeOpInput, controller.changeOpTransaction);
 
+router.route('/sep10').post(controller.signSep10Challenge);
```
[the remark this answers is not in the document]
gianfra-t: Yes, I will add checks now regarding the proper operations that the challenge [needs to have](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0010.md#success). This way we can then fully trust the fronted in that it is not sending any incorrect challenge.

gianfra-t:
> However, this is still not safe as the backend can't trust the frontend. So I think that in order to make it secure, the backend should never return the resulting sep-10 token to the frontend. This means that the backend needs to initiate sep-24 as well.

@TorstenStueber I thought the same, but then I proposed this [hypothesis](https://github.com/pendulum-chain/vortex/pull/236#issuecomment-2442510325). If we add these [extra](https://github.com/pendulum-chain/vortex/pull/236#discussion_r1820641024) checks I think we can trust the frontend, at least in terms of security. The first `manage_data` operation you [sign](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0010.md#success) specifies the account trying to do the `sep-10`. It is not stated (or I couldn't find) from the docs if the `JWT` obtained from such challenge can be used for something else, but it would not be logical.  Still to be seen if this will also work with the 1 account + memo solution.

On the other hand, we could initiate the `sep-24` but then for submitting it the user needs the token, so it would have to be submitted by us as a proxy, seems quite complicated (although it would solve the iframe issue probably).

But there is the impersonating issue. Even with all those checks in place, a "malicious" application could request the authentication with this endpoint, and later just send the funds directly to the anchor's deposit account while benefiting from our special fees, because there are no funds commitment. Is there even a way to prevent this fundamentally?

src/services/signingService.tsx
```diff
@@ -30,3 +36,19 @@ export const fetchSigningServiceAccountId = async (): Promise<SigningServiceStat
     throw new Error('Signing service is down');
   }
 };
+
+export const fetchClientDomainSep10 = async (
+  challengeXDR: string,
+  outToken: OutputTokenType,
+): Promise<ClientDomainSep10Response> => {
+  const response = await fetch(`http://localhost:3000/v1/stellar/sep10`, {
```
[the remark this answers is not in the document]
gianfra-t: Yes, it's for ease of testing.

gianfra-t:
In related with this [issue](https://github.com/pendulum-chain/vortex/issues/237) which I think is very connected (it may also need the `client_domain` and signature from backend), I tested locally a `sep-10` with a variant of this PR containing a `memo` field. 

In that case we would need to add a check that the `memo` field on the [challenge](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0010.md#success) exists and is not null, since the operations themselves won't be able to distinguish between the master account and the derived ones using a memo. This returns a `JWT` where the `sub` field is `account:memo` as [expected](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0010.md#token).

I also attempted a `sep-24` flow with the JWT containing the memo field, but had no luck yet on both anchors.

signer-service/src/api/routes/v1/stellar.route.js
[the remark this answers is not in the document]
TorstenStueber: Yes, just the simple validator like the other endpoints.

TorstenStueber:
> Is there even a way to prevent this fundamentally?

The problem in my opinion is not that the backend receives an incorrect challenge – of course it is still reasonable to make all required test to check the validity of the challenge ([I did not do when I wrote the sep-10 code initially](https://github.com/pendulum-chain/vortex/blob/54e34eb43bb603fc90e4adebcfaa184563bca6a4/src/services/anchor/index.ts#L102)).

The real problem is impersonating SatoshiPay/Vortex. Any malicious app could send their own SEP-10 challenge to use for cosigning. How can we detect that? I think only if we initiate SEP-10 from the backend itself. You are right, we don't need to execute SEP-24 on the backend, the resulting JWT would already contain only our own account (if we don't use the ephemeral account for SEP-10 anymore).

gianfra-t:
When you say that an "app could send their own SEP-10 challenge" what do you mean? Because we will detect before signing all the signatures and memos. This is very connected also with the polygon [verification](https://github.com/pendulum-chain/vortex/issues/237#issuecomment-2445097429) and we should solve both at the same time, I think, at least the ideal solution for the future.

It doesn't matter who starts the `sep-10` but as long as we verify the challenge's operations and that the requester is in control of the polygon address (and matches the memo), we would be secure. What we cannot verify is that the user or the bad actor app will make use of our runway after we cosigned, or will do it on it's own. 

The only way I see to solve this (besides being in control of the funds) is to implement the contract we discussed way at the beginning of the prototype. User would first transfer the tokens to that contract, they will be paused until the backend gives the "green light" to proceed. 

We can ensure that the polygon user is in control of the account given this first transfer happened in the first place(so no challenge), and use the account itself for memo. When the user finishes the KYC, we allow the contract to proceed with the swap, etc. 
The only issue is, if the user doesn't want to go through with the KYC, we can return the funds, but this will have a network cost.

TorstenStueber:
> When you say that an "app could send their own SEP-10 challenge" what do you mean?

Another app could request a SEP-10 challenge from the anchor, providing our client_domain. Then they could just call our signing-service endpoint to add the `SIGNING_KEY` signature to it. Our signing-service endpoint has no way to determine whether this comes through our own app or another app (all the other checks you mention wouldn't help). The 3rd party app would then have a SEP-10 token containing our `client_domain`.

I don't think that we need a contract to solve it if our backend initiates the SEP-10 itself. Right now of course the backend does not control the rest of the flow and it would only start to be secure when the backend executes all the remaining flow, but we will implement it eventually that way.

gianfra-t:
Yes totally, I think we are talking about the same potential issue, I also agree that is enough for now, I wanted to be sure that there exists a secure modification that we can do with some extra effort, even if we don't do it now.

So how should we proceed with this now? Once the anchor accepts our `.toml` and we test this properly, do we go with this implementation?

TorstenStueber:
I am fine with using the implementation you did here. Instead of optimizing for the ideal solution now, let's rather move forward and improve this later.

TorstenStueber:
@gianfra-t is this blocked any longer?

gianfra-t:
It is. We need to be whitelisted by the anchors. We cannot fully tests this and it adds no value to merge now, since we will have to comment or deactivate the functionality anyway (sending a `client_domain` signature while not whitelisted makes the sep 10 fail of course)

It will also be useful for this [ticket](https://github.com/pendulum-chain/vortex/issues/237), the memo ticket, as we discussed the `client_domain` must be sent too (to be confirmed with Anclap of course). The branch I'm working with uses code from this PR. That's why if develop the memo solution before they whitelist us (quite likely), this PR will become outdated and subsumed by the memo ticket.

TorstenStueber:
What about whitelisting? Is this defined in SEP-10? I can't find it.

Adding the client domain right now is particularly important for Mykobo.

gianfra-t:
As I was saying [here ](https://github.com/pendulum-chain/vortex/issues/225#issuecomment-2458020498) also, It doesn't work as-is. Sending our client domain address will fail. This was 5 days ago.

Besides being or not in the `sep-10` it makes sense they only allow client domains that they recognize as valid, right?

gianfra-t:
@pendulum-chain/devs I merged the changes from #201, the `client_domain` is working now with both Anchors.

gianfra-t:
Update from my last [comment](https://github.com/pendulum-chain/vortex/pull/236#issuecomment-2460389409). Solution in #201 is not viable anymore, so I removed the corresponding parts. 

I kept `ARS` "vanilla" integration with ephemeral for now. It may be the case that they fix their UI such that this actually works.

signer-service/src/api/services/sep10.service.js
```diff
@@ -0,0 +1,95 @@
...
+
+  const { operations } = transactionSigned;
+  // Verify the first manage_data operation
+  const firstOp = operations[0];
+  if (firstOp.type !== 'manageData') {
+    throw new Error('The first operation should be manageData');
+  }
+
+  // Only authorize a session that corresponds with the ephemeral client account
+  if (firstOp.source !== clientPublicKey) {
```
gianfra-t, line 38: `clientPublicKey` is supplied by the user, so this doesn't add protection against the UI. It is mostly an extra check to ensure the challenge operation is consistent.

signer-service/src/api/middlewares/validators.js
```diff
@@ -98,6 +98,22 @@ const validatePostSwapSubsidizationInput = (req, res, next) => {
   next();
 };
 
+const validateSep10Input = (req, res, next) => {
+  const { challengeXDR, outToken, clientPublicKey } = req.body;
+  if (!challengeXDR) {
+    return res.status(400).json({ error: 'Missing Anchor challenge' });
```
ebma, line 104: Nit: might be nice to give the exact name of the parameter that is missing in these `400` messages. In this case, mentioning that `challengeXDR` is missing.

signer-service/src/api/services/sep10.service.js
```diff
@@ -0,0 +1,95 @@
...
+
+  //  the web_auth_domain and client_domain operation must be present
+  if (!hasWebAuthDomain) {
+    throw new Error('Transaction must contain a web_auth_domain manageData operation');
+  }
+  if (!hasClientDomain && clientDomainEnabled) {
+    throw new Error('Transaction must contain a client_domain manageData operation');
+  }
+
+  let clientDomainSignature;
```
gianfra-t, line 86: This is our barrier against signing for Anchors that we don't to sign client domain. Since [above](https://github.com/pendulum-chain/vortex/pull/236/files#diff-51df3211523eb9208a1e9b827aa544a7f01fdb5d9fefc57a9e5eece517f33a3fR13) we also fetch the `anchorSigningKey` corresponding to the token type, the UI shouldn't be able to game this.

signer-service/src/api/services/sep10.service.js
```diff
@@ -0,0 +1,95 @@
...
+  if (transactionSigned.sequence !== '0') {
+    throw new Error(`Invalid sequence number: ${transactionSigned.sequence}`);
+  }
+
+  // See https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0010.md#success
+  // memo field should be empty as we assume (in this implementation) that we use the ephemeral
+  // to authenticate. But no memo sub account derivation.
+  if (transactionSigned.memo.value === '') {
+    throw new Error('Memo does not match');
+  }
```
ebma, line 28: I think the condition needs to be flipped and we can give a better error message.

```suggestion
  if (transactionSigned.memo.value !== '') {
    throw new Error('Memo should be empty');
  }
```

signer-service/src/api/services/sep10.service.js
```diff
@@ -0,0 +1,95 @@
...
+    throw new Error('The first operation should be manageData');
+  }
+
+  // Only authorize a session that corresponds with the ephemeral client account
+  if (firstOp.source !== clientPublicKey) {
+    throw new Error('First manageData operation must have the client account as the source');
+  }
+
+  const { anchorExpectedKey: expectedKey, clientDomainEnabled } = TOKEN_CONFIG[outToken];
+  if (firstOp.name !== expectedKey) {
```
ebma, line 43: Are you sure it's called `name` and not `key`? The protocol [says](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0010.md#success) it's supposed to be `key` but maybe it's different in practice.

```suggestion
  if (firstOp.key !== expectedKey) {
```

signer-service/src/api/services/sep10.service.js
```diff
@@ -0,0 +1,95 @@
...
+    const op = operations[i];
+
+    if (op.type !== 'manageData') {
+      throw new Error('All operations should be manage_data operations');
+    }
+
+    // Verify web_auth_domain operation
+    if (op.name === 'web_auth_domain') {
+      hasWebAuthDomain = true;
+      if (op.source !== anchorSigningKey) {
```
ebma, line 65: I think this can't work. In the previous (first) manage data operation, the `key` is `<home domain> auth`. Here, the value is `web_auth_domain` and in your check this includes the ` auth` suffix as that's how the value is defined in the token config. But the protocol doesn't mention the ' auth' suffix again here. So I guess we need to make sure it's not included (?)

signer-service/src/api/services/sep10.service.js
```diff
@@ -0,0 +1,95 @@
...
+    }
+  }
+
+  //  the web_auth_domain and client_domain operation must be present
+  if (!hasWebAuthDomain) {
+    throw new Error('Transaction must contain a web_auth_domain manageData operation');
+  }
+  if (!hasClientDomain && clientDomainEnabled) {
+    throw new Error('Transaction must contain a client_domain manageData operation');
+  }
```
ebma, line 84: Should we alleviate this restriction as it's marked as optional in the protocol?

src/constants/tokenConfig.ts
```diff
@@ -88,6 +90,32 @@ export const OUTPUT_TOKEN_CONFIG: Record<OutputTokenType, OutputTokenDetails> =
...
+    tomlFileUrl: 'https://api.anclap.com/.well-known/stellar.toml',
+    decimals: 12,
+    fiat: {
+      assetIcon: 'ars',
+      symbol: 'ARS',
+    },
+    stellarAsset: {
+      code: {
+        hex: '0x41525300',
+        string: 'ARS\0',
```
ebma, line 105: Should we add the `new Asset()` constructor compatible version to the token config maybe so we don't need to `replace()` the `\0` everywhere? Or we refactor the code to reduce the calls to `replace()`.
```suggestion
        string: 'ARS',
        raw: 'ARS\0',
```

signer-service/src/api/services/sep10.service.js
gianfra-t: Yup, name. Here is an example

```
{
    source: 'GA6KB5YJ6B2CBJK7TZT2UWT7TWS4PNJVWHDMLVQ42ESMDD55XWJXJUWR',
    type: 'manageData',
    name: 'api.anclap.com auth',
    value: <...... more bytes>
  },
  ```

File: signer-service/src/api/services/sep10.service.js (hunk 7 above)
Comment (gianfra-t) [reply signer-service/src/api/services/sep10.service.js:28]: I will change the message, but we actually expect it to be empty. It's very important because if the UI were to send the memo they want, we would essentially validate their identity, without any check they provide us.

On the memo PR we will validate it.

File: signer-service/src/api/services/sep10.service.js (hunk 9 above)
Comment (gianfra-t) [reply signer-service/src/api/services/sep10.service.js:65]: This is actually an example of the operation we check (yes, it's confusing that when decoded, `name` is actually `key`:

```
{
    source: 'GDVHOU4AF2QINLYETV2YFC7YWPRVXN4SKR6SOJZ7LAWODJIZJ7ZPJUER',
    type: 'manageData',
    name: 'web_auth_domain',
    value: <Buffer 61 70 69 2e 61 6e 63 6c 61 70 2e 63 6f 6d>
  },
```

We want to ensure the `source` for that op is the one on the `.toml`. Quite like on the first check above. The `source` for the `<home domain> auth` name/key is [expected](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0010.md#success) to be the client account, in this case the ephemeral.

Comment (gianfra-t) [reply signer-service/src/api/services/sep10.service.js:65]: Of course, we should also check the `value` decoding. This I left because it seemed like too many checks for the prototype. After all, we are not doing so in the UI.

File: signer-service/src/api/services/sep10.service.js (hunk 10 above)
Comment (gianfra-t) [reply signer-service/src/api/services/sep10.service.js:84]: It's mostly for us, and check what the UI client is doing. We can alleviate it, but the intention is that if the UI asks for a challenge corresponding to a specific token/anchor, let's say `eurc`, we check that we allow signing `client_domain` proof for it. Only if we allow it, and the operation correspond to what is expected, we sign.

It may be a bit of too defensive and redundant. We can analyze removing it if it doesn't add value. But it's not too much code.

Comment (gianfra-t) [reply signer-service/src/api/services/sep10.service.js:84]: I should mention that the `client_domain` operation is defined by the initial GET request to `sep-10`, if it is defined there, it will appear in the operations.
And right now we (the backend) don't control this.

File: signer-service/src/api/services/sep10.service.js [hunk 12]
```diff
@@ -0,0 +1,95 @@
...
+    if (op.name === 'client_domain') {
+      hasClientDomain = true;
+      if (op.source !== clientDomainStellarKeypair.publicKey()) {
+        throw new Error('client_domain manage_data operation must have the client domain account as the source');
+      }
+    }
+  }
+
+  //  the web_auth_domain and client_domain operation must be present
+  if (!hasWebAuthDomain) {
```
TorstenStueber, line 79: According to [this section](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0010.md#abstract), it does not seem to be mandatory that the challenge transaction has a `web_auth_domain` entry.

src/services/anchor/index.ts
```diff
@@ -185,19 +198,35 @@ export async function sep12First(sessionParams: IAnchorSessionParams): Promise<v
...
   const { token, tomlValues } = sessionParams;
   const { sep24Url } = tomlValues;
 
-  // at this stage, assetCode should be defined, if the config is consistent.
-  const sep24Params = new URLSearchParams({
-    asset_code: sessionParams.tokenConfig.stellarAsset.code.string,
-    amount: sessionParams.offrampAmount,
-  });
+  let sep24Params;
+  if (outputToken == 'ars') {
```
TorstenStueber, line 214: I would again avoid token specific hard coded logic here, but I understand that this is really specific logic that is only valid for Anclap and we might remove it again once they implement this compliantly (`account` is the SEP-24 instead of the SEP-10 account).

TorstenStueber (approved):

signer-service/src/api/services/sep10.service.js
ebma: Okay but if you want the memo to be empty then we need to flip the condition because otherwise you will throw an error here if it's empty, no?

signer-service/src/api/services/sep10.service.js
ebma: Ah I see, what confused me is the naming then. In the `tokenConfig.js` file you define `anchorExpectedKey: 'api.anclap.com auth'` and because you gave this 'auth' item a name with `key` I mixed it up with the signing key (that is an actual public key). 

So let's rename `anchorExpectedKey` to something else? Personally, I think it might even be best if we adhere closely to the terminology of the SEP-10 protocol and just define it as `homeDomain: 'api.anclap.com'` and then add the `' auth'` programmatically once needed. Also, we could get rid of the extra declaration of `tomlFileUrl: 'https://api.anclap.com/.well-known/stellar.toml'` as this can also be derived from the `homeDomain` with `https://${homeDomain}/.well-known/stellar.toml`

signer-service/src/api/services/sep10.service.js
ebma: Alright, well I guess if we always want to use the `client_domain` we can just keep the check.

signer-service/src/api/services/sep10.service.js
gianfra-t: Actually you're right 🤦. I had the condition the other way around. And the reason it was working is because it's not empty but `null`. So I corrected to _expect_ a null.

signer-service/src/api/services/sep10.service.js
gianfra-t: Mmm okay! I like the idea. Less things to keep track of.

gianfra-t: Wait actually we cannot get rid of the `tomlFile` since for mykobo the domain is `mykobo.co` yet the url for the toml is `https://circle.anchor.mykobo.co/.well-known/stellar.toml`. I think the anchors have some freedom there.

signer-service/src/api/services/sep10.service.js
gianfra-t: This is also useful, I think, when we want to also disable some anchor. Yes I also understand we always want to send the `client_domain`. It's defensive, but for example right now we are experiencing a failure if we specify the `client_domain` with Anclap. So we disable it here and we can be sure it's not a misconfiguration on the frontend.

signer-service/src/api/services/sep10.service.js
gianfra-t: I understand from [here](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0010.md#success) that is not optional. You get that idea from step 7?

src/services/anchor/index.ts
gianfra-t: Yes, me too. But Anclap, therefore `ars`, seems to be the only one requiring the `account` on the `sep-24`. It seems from our current [conversations](https://satoshipay.slack.com/archives/C07PECGDX4Y/p1731032156696839?thread_ts=1730990813.226219&cid=C07PECGDX4Y) that it will continue to be this way.

signer-service/src/api/services/sep10.service.js
TorstenStueber: It sounds contradictory, I saw that. I thought that this is some backward compatible definition, i.e., an older version did not have this. Whoever creates the challenge should add this field in the current version. Whoever checks the challenge should accept also the older challenge without that field.

But who knows. We can just keep it as it is as long as it works with both anchors.

signer-service/src/api/services/sep10.service.js
ebma: Oh 😅 I think Mykobo are breaking the convention then though as the domain should be `circle.anchor.mykobo.co` if I understand the protocol [here](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0001.md#specification) correctly. But if we can't be certain that our partners adhere to this, then let's keep it separate. Let's still rename the variables maybe?

src/constants/tokenConfig.ts
ebma: Yes, that's good 👍  I see that you changed it from `string` to `stringRaw` back to `string`. I think for consistency it would be better to also call it `stringRaw` and `stringStellar` on the typescript/frontend part even if it means touching some references.

TorstenStueber (approved):
👍

Issue #217, opened 2024-10-16
Email not captured and feedback provided


prayagd:
## Context
This ticket is to fix
- Email submit not working and no feedback provided

## Description of bug
* For the second bug when the user enters email and tries submitting, nothing happens and neither is the email captured
* Expected outcome, user should be able to submit the email with message. "Email submitted"

ebma:
What does 'no feedback provided' mean? Can you elaborate on that?

prayagd:
> What does 'no feedback provided' mean? Can you elaborate on that?


Expected outcome, user should be able to submit the email with message. "Email submitted", also mentioned in the #134

ebma:
I still don't get it. What does 'feedback' mean to you?

gianfra-t:
@prayagd may this be an old bug? I'm testing locally and I'm able to see the "Success" feedback. Or an error, in case the service is not working.
<img width="500" alt="image" src="https://github.com/user-attachments/assets/a3d3305b-118e-4768-b340-1ac4c1384460">

<img width="500" alt="image" src="https://github.com/user-attachments/assets/8d95be2b-178e-4cb4-b190-6d9833220401">

gianfra-t:
Never mind, found it.

PR #239, opened 2024-10-30
Cache id to show in success and failure screen

gianfra-t:
#217 

The issue is that after our latest changes, `firstSep24ResponseState` is cleared quickly. When no `id` is passed, clicking `submit` on the email form does nothing. Also, we can see from failure screenshots that the `id` is not there anymore.


ebma (approved):
Nice catch 👍

Issue #241, opened 2024-10-31
Add Sentry


TorstenStueber:
In order to get some better understanding of common bugs, let's add Sentry (potentially only for some limited time).

The credentials are shared. The instructions can be found when logging in, there is an easy React integration. I think it is enough to add Sentry only to the frontend.

Sentry has an API that allows to log extra messages (`Sentry.captureMessage`) and we can use it to add a little bit of extra information, at least when switching to the next offramping phase.

## TODO
* integrate Sentry to the React frontend
* log extra information about phase changes (and maybe some other useful events that come to your mind)

TorstenStueber:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Pendulum-6408a0365cda59c57d39bee5/issues/pendulum-chain/vortex/241?planning-poker) with Zenhub @gianfra-t @ebma @Sharqiewicz

prayagd:
@TorstenStueber this is for the vortex UI?

TorstenStueber:
@prayagd yes.

PR #244, opened 2024-10-31
Add sentry

gianfra-t:
Issue #241.

Sentry integrated for Vortex. It will log unhandled errors from `production` and `development` environment, appending the userId (address), if the wallet was connected. Tests from `localhost` are not reported. Alerts have also been [set up](https://satoshipay.sentry.io/alerts/rules/vortex/87865/).

For now, only info messages (with `captureMessage`) are sent on successful app initialization and change of offramping phase. I believe more messages would be redundant with our report and transactions spreadsheet.

### Firefox strict tracking protection.

It has been noted that the tracking and reporting won't work in Firefox if the strict tracking protection is enabled. Apparently this can be overcome by using [tunnels](https://docs.sentry.io/platforms/javascript/troubleshooting/#using-the-tunnel-option).

### Building locally

On each build, Sentry will upload the source code for later trace analysis on the dashboard. This requires a token that should be located in a file named `.env.sentry-build-plugin` containing: `SENTRY_AUTH_TOKEN=xxxxxx`. This token has been added to Netlify, but if you want to automatically publish to the project when building locally, it needs to be added.


TorstenStueber (approved):
👍

PR #246, opened 2024-10-31
Use proper minimum amounts for Forex AMM

### LINKED ISSUE #240: Bug: set Forex AMM minimum output amount to lower number
The Forex AMM swap operation takes an additional argument `_amountOutMin`, which determines that the swap operation should revert if the actually achieved output amount of the swap would drop below that value.

The usual idea is to get an upfront quote, subtract some slippage tolerance from that amount and use that as the `_amountOutMin` for the subsequent swap operation.

That's not what is happening in Vortex right now. The logic originates from the early prototype and the `_amountOutMin` used is just the quoted output amount rounded down to two decimals.

If the quote is fairly close to its rounded down value, then there is a high risk that due to price changes the actual swap that will take place a few minutes later will revert.

This means that the flow in this early version will not be recoverable as the transaction has been executed on chain and the account nonce already incremented. The swap transaction can therefore not be retried at a later point.

## TODO
* set the `_amountOutMin` to a much lower value, e.g., 5% below the original quote
* before submitting the Nabla swap transaction first request again the current quote and only submit the transaction if the quote only deviates at most 2% from the orginal quote, if not, then go to the failure screen but allow to retry
  * this way we can also try to recover manually at a later point

TorstenStueber:
Closes #240 

I also streamlined some typings, calculations and namings that I found along the code path I had to change.

This function adds the two configuration constants
* `AMM_MINIMUM_OUTPUT_SOFT_MARGIN = 0.02`
* `AMM_MINIMUM_OUTPUT_HARD_MARGIN = 0.05`

The determine the "soft" and "hard" margins:
* The hard margin is the one we use for the minimum amount in the transaction. The transaction will fail and revert when the resulting output amount would be less. In that case the transaction cannot be submitted again as the nonce of the ephemeral account will already increase. The funds would become eternally locked.
* The soft margin is the one we use to check the quote again directly before submitting the swap transaction. If the quote is less than this margin, we don't submit the transaction in the first place and transition to the failure state. This allows to retry later or for us to continue the process manually as the swap transaction is not wasted.


src/components/FeeCollapse/index.tsx
```diff
@@ -4,30 +4,30 @@ import { roundDownToTwoDecimals } from '../../helpers/parseNumbers';
 import { OutputTokenDetails } from '../../constants/tokenConfig';
 import { useEventsContext } from '../../contexts/events';
 
-export function calculateTotalReceive(toAmount: string, outputToken: OutputTokenDetails): string {
+export function calculateTotalReceive(toAmount: Big, outputToken: OutputTokenDetails): string {
```
TorstenStueber, line 7: Made more sense to use `Big` directly in the argument. Before we changed a `Big` to a `string` to call this function, only to change it back into a `Big` inside.

src/constants/constants.ts
```diff
@@ -8,6 +8,9 @@ export const PENDULUM_WSS = 'wss://rpc-pendulum.prd.pendulumchain.tech';
 export const NABLA_ROUTER = '6dQQoUKQ9LNDCrGMjoZjeHBXsuihSgQiQEgD9Z7VtHR82wfG'; // EURC circle instance
 
 export const SPACEWALK_REDEEM_SAFETY_MARGIN = 0.05;
+export const AMM_MINIMUM_OUTPUT_SOFT_MARGIN = 0.02;
```
ebma, line 11: Let's have a brief explanation here maybe.

ebma (approved):
Looks good to me 👍 

One more general question though. Why would you even want to distinguish between the soft and hard margin? So we can try and shift the liquidity in the pools around and give the user a better rate for the existing offramp again?

TorstenStueber:
> Why would you even want to distinguish between the soft and hard margin

It is quite a problem if the swap transaction is submitted but reverts. For the current early version that means that the funds are unrecoverably lost (apart from issuing a root transaction).

I wanted to have a way to make this safe and don't even submit the transaction if the current quote deviates too far. But even then there is still some risk that the result changes again between requesting the quote and executing the call.

Alternatively, we could just set the minimum amount in the swap transaction to 0 (this sounded risky to me) or set the minimum amount to the current (i.e., directly after the AMM approval transaction has been submitted and before submitting the swap transaction) quote minus some margin. I think in practice all 3 solutions probably work equally well.

ebma:
> But even then there is still some risk that the result changes again between requesting the quote and executing the call.

I see. So the main reason we don't submit if x < soft_margin is to not get to a situation where the swap reverts.

PR #249, opened 2024-10-31
Only create api promise once

### LINKED ISSUE #248: Bug: Memory Leak in signer service
The signer service has a memory leak caused by initiating new connections via polkadot-js.

This already happens in the status endpoint. The status endpoint opens the following connections and queries on-chain data:
* to Stellar
* to Pendulum
* to Moonbeam

Through experiments I can conclusively say that the connections to Stellar and Moonbeam via their respective libraries do not cause memory leaks and allow that all their allocated space can be cleanly garbage collected.

However, each call to `sendPendulumStatusWithPk` causes about 12 MByte of heap allocation that is not eligible for garbage collection.

## TODO
Find a way to avoid the memory leak.

TorstenStueber:
Closes #248 

This solution avoids the memory leak and even after running for many requests, the heap will regularly be completely garbage collected.

However, I don't know whether this solution works reliably over a long time. I just tested locally and disconnected WIFI repeatedly and the service could successfully reconnect again. Unfortunately, I can't find any documentation for these functions.

### DIFF
--- a/signer-service/src/api/services/pendulum.service.js
+++ b/signer-service/src/api/services/pendulum.service.js
@@ -22,18 +22,35 @@ function multiplyByPowerOfTen(bigDecimal, power) {
 }
 
 let api;
+let previousSpecVersion;
 
 async function createPolkadotApi() {
-  if (!api) {
+  const getSpecVersion = async () => {
+    const runtimeVersion = await api.call.core.version();
+    return runtimeVersion.toHuman().specVersion;
+  };
+
+  const initiateApi = async () => {
     const wsProvider = new WsProvider(PENDULUM_WSS);
     api = await ApiPromise.create({
       provider: wsProvider,
     });
     await api.isReady;
+
+    previousSpecVersion = await getSpecVersion();
+  };
+
+  if (!api) {
+    await initiateApi();
   }
 
   if (!api.isConnected) await api.connect();
   await api.isReady;
+
+  const currentSpecVersion = await getSpecVersion();
+  if (currentSpecVersion !== previousSpecVersion) {
+    await initiateApi();
+  }
 
   const chainProperties = api.registry.getChainProperties();
   const ss58Format = Number(chainProperties?.get('ss58Format')?.toString() ?? 42);
--- a/signer-service/src/api/services/pendulum.service.js
+++ b/signer-service/src/api/services/pendulum.service.js
@@ -21,11 +21,18 @@ function multiplyByPowerOfTen(bigDecimal, power) {
   return newBigDecimal;
 }
 
+let api;
+
 async function createPolkadotApi() {
-  const wsProvider = new WsProvider(PENDULUM_WSS);
-  const api = await ApiPromise.create({
-    provider: wsProvider,
-  });
+  if (!api) {
+    const wsProvider = new WsProvider(PENDULUM_WSS);
+    api = await ApiPromise.create({
+      provider: wsProvider,
+    });
+    await api.isReady;
+  }
+
+  if (!api.isConnected) await api.connect();
   await api.isReady;
 
   const chainProperties = api.registry.getChainProperties();
--- a/signer-service/yarn.lock
+++ b/signer-service/yarn.lock
@@ -378,27 +378,27 @@ __metadata:
   linkType: hard
 
 "@polkadot/keyring@npm:^13.1.1":
-  version: 13.2.1
-  resolution: "@polkadot/keyring@npm:13.2.1"
+  version: 13.2.2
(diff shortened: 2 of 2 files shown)


signer-service/src/api/services/pendulum.service.js
```diff
@@ -21,13 +21,37 @@ function multiplyByPowerOfTen(bigDecimal, power) {
...
+  };
+
+  if (!api) {
+    await initiateApi();
+  }
+
+  if (!api.isConnected) await api.connect();
   await api.isReady;
 
+  const currentSpecVersion = await getSpecVersion();
```
TorstenStueber, line 50: Ensure that we reload all metadata after a runtime upgrade.

ebma (approved):
Looks good to me 👍 

In other services we already have a caching mechanism in place. It's implemented with [this](https://github.com/pendulum-chain/spacewalk-testing-service/blob/09574f7124870490125a2c51bf398bc42981b167/src/vault_service/api.ts#L11) class. Since we are only dealing with one network (pendulum) in the signer-service, I and don't need to cache the API per network, I think it's fine the way you implemented it.

TorstenStueber:
Thanks @ebma, I didn't even check how we do it elsewhere. Just looking at the other code, they don't reconnect the api in case it becomes disconnected. I didn't expect this to work.

I was just testing this case locally. Actually when the connection fails and you wait long enough, polkadot-js just crashes the complete app, so that it is forced to restart completely anyway. This is exactly not what I expect a library to do. 😱

So ignore my concern, the other codebases will work then.

ebma:
> Just looking at the other code, they don't reconnect the api in case it becomes disconnected. I didn't expect this to work.

It does actually. In the testing-service, if you have a look [here](https://github.com/pendulum-chain/spacewalk-testing-service/blob/09574f7124870490125a2c51bf398bc42981b167/src/vault_service/api.ts#L78-L80), the API is reconnected when a call fails (which might fail because the API is disconnected, or because the metadata is outdated, etc). The same logic is in the token-api [here](https://github.com/pendulum-chain/token-api/blob/35f4db9204da15fee0aecee6240f35124c0e866a/src/api/services/rpc.service.js#L37-L40) and in the faucet [here](https://github.com/pendulum-chain/faucet/blob/ee022fd01973a60a14637f6a894de296ef29ccd7/src/chain_service/api.ts#L63-L65).

> Actually when the connection fails and you wait long enough, polkadot-js just crashes the complete app, so that it is forced to restart completely anyway. This is exactly not what I expect a library to do. 😱

That's crazy though, does it throw a specific error or how does it crash?

TorstenStueber:
> It does actually. 

True, I missed that.

> does it throw a specific error

The error is:

```
2024-11-01 07:55:17        RPC-CORE: call(method: Text, data: Bytes, at?: BlockHash): Bytes:: No response received from RPC endpoint in 60s
<redacted>/signer-service/node_modules/@polkadot/rpc-provider/cjs/ws/index.js:497
                    handler.callback(new Error(`No response received from RPC endpoint in ${this.__internal__timeout / 1000}s`), undefined);
                                     ^

Error: No response received from RPC endpoint in 60s
    at WsProvider.__internal__timeoutHandlers (<redacted>/signer-service/node_modules/@polkadot/rpc-provider/cjs/ws/index.js:497:38)
    at Timeout._onTimeout (<redacted>/signer-service/node_modules/@polkadot/rpc-provider/cjs/ws/index.js:178:65)
    at listOnTimeout (node:internal/timers:573:17)
    at process.processTimers (node:internal/timers:514:7)

Node.js v21.7.0
```

Having a closer look at this, it might not be polkadot-js to blame but we possibly just miss some proper catch blocks around any calls to the polkadotjs Api object, the way it is done in the other codebases.

ebma:
Hmm okay, weird. I guess it's because of [this](https://github.com/polkadot-js/api/blob/9becdced121977a22c81df4d037c9d0e60b370a5/packages/rpc-provider/src/ws/index.ts#L140) timeout. It's 60 seconds by default but we can pass another one. Looking at [this](https://github.com/polkadot-js/api/blob/9becdced121977a22c81df4d037c9d0e60b370a5/packages/rpc-provider/src/ws/index.ts#L620), every call we do is stored as a handler and these handlers are looped to check if any of them timed out. Seems like handlers are only created for each call to `send()` and the API wouldn't time out just like that without being used 🤔

TorstenStueber:
I realized that we configured the express server wrong and the error handler is not called properly so that the whole app crashes instead of returning a 500. I just fixed this [in another PR](https://github.com/pendulum-chain/vortex/pull/251).

PR #251, opened 2024-11-01
Upgrade express version

### LINKED ISSUE #250: Bug: Express error handler not called
Whenever one of the signer service handler functions throw an unhandled error, then the Express app should invoke the error handler and return a 500 response.

Instead the whole app crashes and no response is sent at all.

## TODO
Correctly configure the error handler.

TorstenStueber:
Closes #250 

Turned out that upgrading to Express 5.0 resolved the problem.

### DIFF
--- a/signer-service/package.json
+++ b/signer-service/package.json
@@ -29,7 +29,7 @@
     "cors": "^2.8.3",
     "cross-env": "^7.0.3",
     "dotenv": "^16.4.5",
-    "express": "^4.15.2",
+    "express": "^5.0.1",
     "express-rate-limit": "^6.7.0",
     "express-validation": "^1.0.2",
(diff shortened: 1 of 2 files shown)


ebma:
Updating to v5.x.x is enough? So the buggy error handler was a known problem that they fixed in the meantime?

TorstenStueber:
I don't think it was a bug in Express, I think just the way the error handler [is invoked automatically](https://expressjs.com/en/guide/error-handling.html) is a breaking change between Express 4.x and 5.x and it looks to me like the way we configured our Express app was taken from the docs for Express 5.x instead of Express 4.x. So the way to fix it was just to update to the right version.

ebma (approved):
Ah I see, makes sense.

PR #252, opened 2024-11-01
Add more information to GTM events

### LINKED ISSUE #242: Add more information to GTM events
It would be useful to understand when (it what phase) failures happen. For that reason add the current offramping phase to the `transaction_failure` screen.

## TODO
* add `phaseName` parameter to the `transaction_failure` event, which is the `OfframpingPhase` string
* add `phaseIndex` parameter to the `transaction_failure` event, which is has the same meaning as the `phase` parameter for the `progress` event (see [this PR](https://github.com/pendulum-chain/vortex/pull/222/files#diff-7235bfd8349bca9fadbfdb21af86ddcd184fdcea4bc184b97330737a4d91b35bR51))
* define these two additional parameters in GTM

ebma:
I modified the GTM setup as follows:
- Rename `GA Event | transaction_failure` to `GA4 Event | transaction_failure`
- Add two new event parameters to `GA4 Event | transaction_failure`: `phase_name` and `phase_index`
- Add two new Variables `DLV | phase_index` and `DLV | phase_name`
- Modified the event parameters of `GA4 Event | progress` to use the new variables as well
- Deleted the variable `DLV | phase` and `DLV | progress name` (the naming was confusing and it's better to streamline with the other two news variables)

Closes #242.

### DIFF
--- a/src/contexts/events.tsx
+++ b/src/contexts/events.tsx
@@ -38,13 +38,22 @@ export interface WalletConnectEvent {
   wallet_action: 'connect' | 'disconnect' | 'change';
 }
 
-export interface TransactionEvent {
-  event: 'transaction_confirmation' | 'kyc_started' | 'kyc_completed' | 'transaction_success' | 'transaction_failure';
+interface OfframpingParameters {
   from_asset: string;
   to_asset: string;
   from_amount: string;
   to_amount: string;
 }
+
+export type TransactionEvent = OfframpingParameters & {
+  event: 'transaction_confirmation' | 'kyc_started' | 'kyc_completed' | 'transaction_success' | 'transaction_failure';
+};
+
+export type TransactionFailedEvent = OfframpingParameters & {
+  event: 'transaction_failure';
+  phase_name: string;
+  phase_index: number;
+};
 
 export interface ProgressEvent {
   event: 'progress';
@@ -86,6 +95,7 @@ export type TrackableEvent =
   | ClickDetailsEvent
   | WalletConnectEvent
   | TransactionEvent
+  | TransactionFailedEvent
   | ClickSupportEvent
   | FormErrorEvent
   | EmailSubmissionEvent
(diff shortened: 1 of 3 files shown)


src/contexts/events.tsx
```diff
@@ -38,18 +38,27 @@ export interface WalletConnectEvent {
...
-  event: 'transaction_confirmation' | 'kyc_started' | 'kyc_completed' | 'transaction_success' | 'transaction_failure';
+interface OfframpingParameters {
   from_asset: string;
   to_asset: string;
   from_amount: string;
   to_amount: string;
 }
 
+export type TransactionEvent = OfframpingParameters & {
+  event: 'transaction_confirmation' | 'kyc_started' | 'kyc_completed' | 'transaction_success' | 'transaction_failure';
```
ebma, line 49: `transaction_failure` is still contained in this list on purpose so we can use the `createTransactionEvent()` function for the `TransactionFailedEvent` as well

TorstenStueber:
@ebma is this ready to review? (no reviewers added yet)

ebma:
Missed adding reviewers, sorry.

src/hooks/useMainProcess.ts
```diff
@@ -67,7 +68,15 @@ export const useMainProcess = () => {
       if (state?.phase === 'success') {
         trackEvent(createTransactionEvent('transaction_success', state));
       } else if (state?.failure !== undefined) {
-        trackEvent(createTransactionEvent('transaction_failure', state));
+        const currentPhase = state?.phase;
+        const currentPhaseIndex = Object.keys(OFFRAMPING_PHASE_SECONDS).indexOf(currentPhase);
+
+        trackEvent({
+          ...createTransactionEvent('transaction_failure', state),
+          event: 'transaction_failure',
```
TorstenStueber, line 76: Tiny nit: this `event` here should not be required.

TorstenStueber (approved):
👍

src/hooks/useMainProcess.ts
ebma: True but if I don't include it, TypeScript will complain
```
TS2353: Object literal may only specify known properties, and phase_name does not exist in type TransactionEvent
```
for `phase_name` and `phase_index`. I can either add a cast to `as TransactionFailureEvent` for the whole object or keep the `event: 'transaction_failure'` as is.

Issue #237, opened 2024-10-29
Change the SEP24 flow to authenticate recurring users


vadaynujra:
### Context
The ARS anchor needs to be able to identify and authenticate recurring users to provide a compliant off-ramp service. 

### To do

- [ ] Add changes to SEP24 to include use the memo field such that the user doesn't need to reenter login details.

**Additional info:**
Memo info would be saved with the user and user does not need to reenter bank data etc. again
Can the tx come from the same user (vortex) and individual users are only identified through the memo

TorstenStueber:
The idea would be to use the same account (owned by us) for all SEP-10 authentication and distinguish users only be a text memo. The question is how to derive this text memo? We could hash their Polygon account and use the maximal initial segment we can use for a memo (or use a technique similar to Spacewalk, where we also extract information from an address to fit into a text memo).

vadaynujra:
Identifying the users - we can choose the approach, as long as we're using the memo field
KYC-ing the users - stays with anchor

gianfra-t:
I can think of:
- A subset of the hash of the account like you say. 
- A subset of the account itself.
- A subset of the hash of the signature from the challenge we will provide. I do not know which one has higher entropy and randomness. Probably the latter?

On a related topic, I already tried [a few things](https://github.com/pendulum-chain/vortex/pull/236#issuecomment-2444467268) on the client domain [PR](https://github.com/pendulum-chain/vortex/pull/236). Could the full flow be something like this?: 
- Send a random challenge
- User signs and sends the signature back.
- Derive the text memo.
- If okay, we fetch and sign the challenge from the anchor with corresponding memo, and our main account as source account. 
- We return either the jwt or the signed challenge to the UI.

This is disregarding the "free use of our Client App" [issue](https://github.com/pendulum-chain/vortex/pull/236#issuecomment-2444016907) I'm still not sure about.

TorstenStueber:
The question is whether we want to let the user sign another request in their wallet (UX not so nice this way) to prove their ownership of the Polygon account. I would argue that could skip this at this phase of the project.

prayagd:
> The question is whether we want to let the user sign another request in their wallet (UX not so nice this way) to prove their ownership of the Polygon account

I would say we should include this can keep this in the start itself when the user connects the account they sign this request. Pretty standard now in all dApps. 

>  I would argue that could skip this at this phase of the project.

Would skipping this introduce any security issues?

prayagd:
[...]

PR #257, opened 2024-11-06
Authenticate recurrent users with memo.

gianfra-t:
Closes: #237.

### Context and Preliminaries.

This PR adds a way to authenticate users by means of a signature using their wallet (commonly called: Sign in with web3, SIWE). The signer-service will use this signature in a `JWT` fashion to authorize the signing of the sep-10 operations from the signer-service.

The motivation for these changes is to add a scheme where the application is able to authenticate the users by it's initial Polygon/EVM address and avoid re-do of KYC and storing extra credentials.

### Detailed overview 

To identify the users based on their authenticated address, the sep-10 operations now correspond to the specification of a `memo` that is derived _uniquely_ and in a _deterministic_ way only from the _EVM_ address. 
The user identifier is then the combination of a `master` account concatenated with the `memo` (`master:memo`) and must be seen by the Anchor as a completely different user than one authenticated with only the `master` account. (relevant section of the [protocol](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0010.md#memos))

Only with a valid signature corresponding to the _EVM_ address in question will the signer-service respond with the corresponding signatures, which verifies: `memo`, `client-domain` and `master` account.

#### Security guarantees
Without these signatures from the service, the `sep-10` will fail when requested to the Anchor. Although it is not stated on the protocol (or I missed it) it is expected that the Anchor would ONLY allow a `memo` specification when the `client-domain` is defined, and valid. Apparently, anchors could also whitelist the `master` account for this purposes. 

#### Flow of operations

- The user opens Vortex.
- (optional) Upon connect/change of address it is presented with a modal asking to sign a message.
- The user attempts an offramp. If needed by Anchor, we search for the user's signature in the storage, check it's expiration time, and potentially ask the user to sign the message again if it doesn't exist or if it's expired. (we don't ask for validation to the service here, to save requests). 
- The user now sends a request to the signer-service with this signature, and the corresponding sep-10 parameters. The service returns the corresponding stellar signatures.
- If the user's signature turns out to be invalid (eg.: the service has lost it's state), then a new sign-in operation will be possible without interrupting the flow.

#### Session

So far, the concept of session is simply kept by means of the signature stored in the UI (local storage...). Again, drawing parallels to a `JWT`. The message signed has a date parameter checked by the signer-service.

Alternatively, if the user signs from another browser/machine etc, then the old signature will be invalidated on the signer-service.

Requirements for log-out have not been defined yet.

### Security risks, and potential improvements.

This is as far as I would go with this simple/prototypical implementation of using the signature as means of authentication. 
Update: We switched to use secure cookies now, instead of storing the signature in the local storage. Should be much more secure.
~~There is a potential risk since this will be stored in the `localStorage`, unlike secure cookies. Therefore, if we are going to aim for a comprehensive session management, I believe we should implement something like `iron-session` to _keep_ the session, while the initiation would still be by means of this signature verification. This seems to be the common standard practice.~~

### Extra libraries concerns.

This pr adds the library [siwe](https://www.npmjs.com/package/siwe) which is useful for creating and parsing the message, and potentially verifying. This is useful yet not critical, and the issue is that it requires `ethers` dependency. So to be more compatible with our current libraries and have the freedom to remove this one, we use `viem` to verify on the backend side.


signer-service/src/api/controllers/stellar.controller.js
```diff
@@ -54,12 +54,37 @@ exports.changeOpTransaction = async (req, res, next) => {
...
       req.body.challengeXDR,
       req.body.outToken,
       req.body.clientPublicKey,
+      req.body.maybeChallengeSignature,
+      req.body.maybeNonce,
     );
-    return res.json({ clientSignature, clientPublic });
+    return res.json({ masterClientSignature, masterClientPublic, clientSignature, clientPublic });
+  } catch (error) {
+    if (error.message.includes('Could not verify signature')) {
```
gianfra-t, line 66: We may want to use internal error codes later.

signer-service/src/api/services/sep10.service.js
```diff
@@ -88,8 +123,15 @@ exports.signSep10Challenge = async (challengeXDR, outToken, clientPublicKey) =>
     clientDomainSignature = transactionSigned.getKeypairSignature(clientDomainStellarKeypair);
   }
 
+  let masterClientSignature;
+  if (memo !== null && memoEnabled) {
```
gianfra-t, line 127: We've verified all operations so reaching here should be safe to append the signature.

src/hooks/useSignChallenge.ts
```diff
@@ -0,0 +1,123 @@
...
+      if (!stored) return null;
+
+      const data: SiweSignatureData = JSON.parse(stored);
+      return new Date(data.expirationDate) > new Date() ? data : null;
+    } catch {
+      localStorage.removeItem(storageKey);
+      return null;
+    }
+  }, [address, storageKey]);
+
```
gianfra-t, line 40: We want to return a promise that will resolve when the signing finishes (or fails). This is so we can use `checkAndWaitForSignature` and it's force counterpart on the `sep-10` process function.

src/hooks/useSignChallenge.ts
```diff
@@ -0,0 +1,123 @@
...
+        body: JSON.stringify({ walletAddress: address }),
+      });
+
+      if (!response.ok) throw new Error('Failed to create message');
+      const { siweMessage, nonce } = await response.json();
+
+      const message = new SiweMessage(siweMessage);
+      const signature = await signMessageAsync({ message: siweMessage });
+
+      const validationResponse = await fetch(`${SIGNING_SERVICE_URL}/v1/siwe/validate`, {
```
gianfra-t, line 64: Mostly so that an old session is removed. Otherwise we may not need to validate, and we can save the roundtrip.

src/hooks/useSignChallenge.ts
```diff
@@ -0,0 +1,123 @@
...
+    if (stored) return stored;
+    return signMessage();
+  }, [checkStoredSignature, signMessage]);
+
+  const forceRefreshAndWaitForSignature = useCallback(async (): Promise<SiweSignatureData> => {
+    localStorage.removeItem(storageKey);
+    return signMessage();
+  }, [storageKey, signMessage]);
+
+  // ask for signature on address change and on init
```
gianfra-t, line 108: I found this a bit invasive and annoying. But could be enabled. Perhaps with a timer.

src/services/anchor/index.ts
```diff
@@ -61,19 +64,71 @@ export const fetchTomlValues = async (TOML_FILE_URL: string): Promise<TomlValues
   };
 };
 
-async function getUrlParams(ephemeralAccount: string, supportsClientDomain: boolean): Promise<URLSearchParams> {
-  if (supportsClientDomain) {
-    return new URLSearchParams({ account: ephemeralAccount, client_domain: config.applicationClientDomain });
+//A memo derivation. TODO: Secure? how to check for collisions?
```
gianfra-t, line 67: Hash the address, and get the numeric representation of the hash. Apparently the memo needs to be numeric.

src/services/anchor/index.ts
```diff
@@ -61,19 +64,71 @@ export const fetchTomlValues = async (TOML_FILE_URL: string): Promise<TomlValues
...
+    };
   }
-  return new URLSearchParams({ account: ephemeralAccount });
+  return {
+    urlParams: new URLSearchParams({ account: ephemeralAccount, client_domain: config.applicationClientDomain }),
+    sep10Account: ephemeralAccount,
+  };
 }
 
+const sep10SignaturesWithLoginRefresh = async (
```
gianfra-t, line 107: Helper to retry once if the service returns a signature error, with a force refresh in the middle.

This is for the case the signer service has invalidated the signature, for some reason (it restarted or the user logged in elsewhere and then came back to use the old session)

src/services/anchor/index.ts
```diff
@@ -105,17 +162,38 @@ export const sep10 = async (
     throw new Error(`Invalid sequence number: ${transactionSigned.sequence}`);
   }
 
-  const { clientSignature, clientPublic } = await fetchSep10Signatures(
-    transactionSigned.toXDR(),
-    outputToken,
-    accountId,
+  let signatureData = await checkAndWaitForSignature();
```
gianfra-t, line 165: We will fetch the signature, if it's invalid or doesn't exist, will ask the user to sign and only then continue.

src/services/anchor/index.ts
```diff
@@ -210,16 +227,18 @@ export async function sep24First(
...
-    if (!sep10Account) {
-      throw new Error('Master must be defined at this point.');
-    }
+  if (usesMemo) {
     sep24Params = new URLSearchParams({
       asset_code: sessionParams.tokenConfig.stellarAsset.code.stringStellar,
       amount: sessionParams.offrampAmount,
-      account: sep10Account, // THIS is a particularity of Anclap. Should be able to work without it, or with a different one
-      // to that of the sep-10
+      account: sep10Account, // THIS is a particularity of Anclap. Should be able to work just with the epmhemeral account
```
gianfra-t, line 237: We are still waiting for Anclap to solve on their side, or tell us if we are sending something wrong. 
The commented `memo` parameter is deprecated... but it could be the case that will be required.

gianfra-t:
@pendulum-chain/devs You can try this one locally, including Anclap until the offramp UI (or use the "testing" credentials to interact)
The signer-service must be ran locally.

src/hooks/useSignChallenge.ts
ebma: I think we should enable it so that users are immediately requested to log in after connecting the wallet. Makes sense to combine these two events so that the user doesn't need to sign thrice later.

src/constants/localStorage.ts
```diff
@@ -10,6 +10,7 @@ export const storageKeys = {
   ANCHOR_SESSION_PARAMS: 'ANCHOR_SESSION_PARAMS',
   STELLAR_OPERATIONS: 'STELLAR_OPERATIONS',
   TOKEN_BRIDGED_AMOUNT: 'TOKEN_BRIDGED_AMOUNT',
+  SIWE_SIGNATURE_KEY_PREFIX: 'siwe-signature-',
```
ebma, line 13: Is the capitalization format required like this? If it's not, I'd say we make it similar to the others and capitalize to `SIWE_SIGNATURE_`

src/hooks/useMainProcess.ts
```diff
@@ -33,9 +32,14 @@ export interface ExecutionInput {
...
 
+interface UseMainProcessProps {
+  checkAndWaitForSignature: () => Promise<SiweSignatureData>;
+  forceRefreshAndWaitForSignature: () => Promise<SiweSignatureData>;
+}
+
 type ExtendedExecutionInput = ExecutionInput & { stellarEphemeralSecret: string };
 
-export const useMainProcess = () => {
+export const useMainProcess = ({ checkAndWaitForSignature, forceRefreshAndWaitForSignature }: UseMainProcessProps) => {
```
ebma, line 42: Since `useMainProcess` is a hook itself, you can call other hooks in it. Thus, you don't need to pass these two functions as props but you can just call `useSignChallenge` inside the body of `useMainProcess` and access the respective functions.

src/services/anchor/index.ts
```diff
@@ -61,19 +64,71 @@ export const fetchTomlValues = async (TOML_FILE_URL: string): Promise<TomlValues
...
+const sep10SignaturesWithLoginRefresh = async (
+  refreshFunction: () => Promise<SiweSignatureData>,
+  args: SignerServiceSep10Request,
+) => {
+  try {
+    return await fetchSep10Signatures(args);
+  } catch (error: any) {
+    if (error.message === 'Invalid signature') {
+      let { nonce, signature } = await refreshFunction();
+      let regreshedArgs = { ...args, maybeChallengeSignature: signature, maybeNonce: nonce };
```
ebma, line 116:
```suggestion
      let refreshedArgs = { ...args, maybeChallengeSignature: signature, maybeNonce: nonce };
```

ebma (commented):
Interesting approach with SIWE. I almost feel like it's too sophisticated for our use case and we can live with something simpler but since you already made it work, I'm happy to keep it as it's better security-wise.

signer-service/src/api/services/sep10.service.js
```diff
@@ -1,17 +1,46 @@
...
-exports.signSep10Challenge = async (challengeXDR, outToken, clientPublicKey) => {
+async function deriveMemoFromAddress(address) {
+  const hash = keccak256(address);
+  return BigInt(hash).toString().slice(0, 15);
+}
+
+// we validate a challenge for a given nonce. From it we obtain the address and derive the memo
+// we can then ensure that the memo is the same as the one we expect from the anchor challenge
+const getAndValidateMemo = async (nonce, userChallengeSignature) => {
+  if (!userChallengeSignature || !nonce) {
```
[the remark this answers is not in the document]
gianfra-t: Well, with the same signature you will get the same memo, but it's true that the nonce may be confusing given the name.
What if we name it `validateSignatureAndGetMemo`? If you prefer separated, I guess we can validate first and then derive the memo directly on `signSep10Challenge`.

signer-service/src/api/services/siwe.service.js
```diff
@@ -0,0 +1,74 @@
...
+  const siweMessage = new siwe.SiweMessage({
+    scheme,
+    domain,
+    address,
+    statement,
+    uri: origin,
+    version: '1',
+    chainId: polygon.id,
+    nonce,
+    expirationTime: new Date(Date.now() + 360 * 60 * 1000).toISOString(),
```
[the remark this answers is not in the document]
gianfra-t: Nothing specific, as we still need to define the log out and expiration behavior. I'll add it to constants though.

signer-service/src/api/services/siwe.service.js
```diff
@@ -0,0 +1,74 @@
...
+  siweMessagesMap.forEach((data, nonce) => {
+    if (data.address === maybeSiweData.address && nonce !== maybeSiweData.siweMessage.nonce) {
+      siweMessagesMap.delete(nonce);
+    }
+  });
+
+  return maybeSiweData.siweMessage;
+};
+
+// TODO we need some sort of session log-out.
```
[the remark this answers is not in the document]
gianfra-t: It will grow indefinitely for new addresses. This simple way of keeping valid sessions is very rudimentary, that's why I would use a session library if we want more functionality.

Regarding if we need it or not, technically no, but  I believe it's just a standard. For instance [polkassembly](https://pendulum.polkassembly.io/) has a log out "feature". Again we need to define this.

src/hooks/useSignChallenge.ts
```diff
@@ -0,0 +1,123 @@
...
+  // Used to wait for the modal interaction and/or return of the
+  // signing promise.
+  const signPromiseRef = useRef<{
+    resolve: (data: SiweSignatureData) => void;
+    reject: (reason: Error) => void;
+  } | null>(null);
+
+  const storageKey = `${storageKeys.SIWE_SIGNATURE_KEY_PREFIX}${address}`;
+
+  const checkStoredSignature = useCallback((): SiweSignatureData | null => {
```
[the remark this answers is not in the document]
gianfra-t: The function will refresh to a change in `storageKey` and `address` as the hook `useSiweSignature` itself depends on `address`.
I just keep `address` also in the array to filter for `undefined` values.

src/hooks/useSignChallenge.ts
```diff
@@ -0,0 +1,123 @@
...
+  expirationDate: string;
+}
+
+export function useSiweSignature(address?: `0x${string}`) {
+  const { signMessageAsync } = useSignMessage();
+  const [requiresSign, setRequiresSign] = useState(false);
+
+  // Used to wait for the modal interaction and/or return of the
+  // signing promise.
+  const signPromiseRef = useRef<{
```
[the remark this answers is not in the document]
gianfra-t: This one was interesting. At first I tried that, but since `checkAndWaitForSignature` and the `force` counterpart were being used in the `sep-10` function which is not a hook, I was experiencing issues and the state was not reflecting the change.

I later changed the implementation (before I was polling for the signature with a `setTimeout`) so it might be that there is no issue storing the promise in a simple state. I would have to test again.

src/hooks/useMainProcess.ts
gianfra-t: The thing is that we need `requiresSign`, `handleCancel`, etc for the modal component, which is on the same [level](https://github.com/pendulum-chain/vortex/pull/257/files#diff-777e0c2221f5feeef076c359496045ddd95d6d376a0fc1395b330f8a9b64a5bbR352) as `useMainProcess`. 

And those 2 functions as a [parameter](https://github.com/pendulum-chain/vortex/pull/257/files#diff-fedb8a30074ce22b1fb6aa02c40daa53038ea2a73c885a7bd7db6f5dfe5e8efaR152) in the `sep-10` which is inside `useMainProcess`. Alternatively we can set a context but seemed like more trouble, or return the modal states from `useMainProcess`.

I already faced the issue of using `useSignChallenge` on both levels and doesn't work of course.

src/hooks/useSignChallenge.ts
gianfra-t: I'll enable it so we can test it. I worry a bit that is invasive in the sense that if the user only wants to see quotes or offramp `EURC`, it's not needed and may scare some users. 

On the other hand it's even more spaced from the other 2 signatures so "feels" better.

signer-service/src/api/services/sep10.service.js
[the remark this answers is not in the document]
ebma: I'm okay with naming it `validateSignatureAndGetMemo`.

signer-service/src/api/services/siwe.service.js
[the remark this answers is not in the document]
ebma: Yeah maybe we should use sessions. Or at least persist the nonce map so that when the server restarts (due to redeployment or crash) users don't need to always sign in again which would be the case if we only keep an in-memory list.

src/hooks/useMainProcess.ts
ebma: Then we might want to move parts of the `useSignChallenge` hook into a context so that it's shared and every call to `useSignChallenge` returns the shared values. A similar example would be the `useBridgeSettings` hook in the portal that [uses a shared context](https://github.com/pendulum-chain/portal/blob/44f8a2ff8743f31dd1665a394f7310d60152424a/src/hooks/spacewalk/useBridgeSettings.ts#L72) for storing the settings a user chose between issue and redeem requests.

TorstenStueber:
@gianfra-t as I don't have ARS, I tried to offramp EUR. When I click on the Confirm button, I am presented the sign in popup.

When I click on "Sign Message", nothing happens, the popup just disappears and the form is shown again. If I click on the Confirm button again, the same happens – indefinitely.

Is it even intended that the popup is shown in the first place? The identity verification via signature is not required for Mykobo.

gianfra-t:
> If I click on the Confirm button again, the same happens – indefinitely.

This behavior is not intended definitely. What the console shows? We need some more error handling from the UI side, perhaps a toast, I left that open on purpose since there was also no agreement on that.

Nevertheless, for me there is no issue when selecting EURC and it works the same as before. So I imagine there must be a variable missing when you run `signer-service`. 

Also we need to add a condition to NOT ask for signature with EURC probably (this, I forgot), although I will not add it now until I understand what happens in your case.

TorstenStueber:
Yes, sorry, should have provided that:

Screenshot 2024-11-14 at 08 47 02

I guess the signer service of the preview deployment does not have the new changes yet and I would need to run locally instead.

gianfra-t:
Yes you need to run it locally. Sadly we don't deploy a service preview so far.

src/pages/swap/index.tsx
```diff
@@ -177,6 +182,7 @@ export const SwapPage = () => {
           outputTokenType: to as OutputTokenType,
           amountInUnits: fromAmountString,
           offrampAmount: tokenOutAmountData.roundedDownQuotedAmountOut,
+          setInitializeFailed,
```
gianfra-t, line 185: I realized we had no action for when `handleOnSubmit` failed. Probably this doesn't happen to often since `sep-10` is quite reliable, but now we add the message sign in the middle which could fail from the s. service.

With this, we will just show the `Application initialization fail...` message in this case. Same as on the pre-flight check. Let me know what you think.

TorstenStueber:
I didn't review the entire code yet, but when I tested locally I found the following.

1. Why create the siwe message in the backend? I would not expect that we need this new endpoint and we can simply do it as [in this example](https://github.com/spruceid/siwe-quickstart/tree/main/01_frontend).

2. Metamask shows the following warning when I want to sign the signing request. Is this only happening because I run on localhost and it would disappear once deployed in the cloud (I tried to change the parameters for `siwe.SiweMessage` but could not make the warning disappear)? It would be good if this works for any domain, particularly also for the netlify preview domains.

Screenshot 2024-11-15 at 06 32 20

3. I need to sign in even when I just select EUR, can we avoid signing in if we will not use the memo for the respective anchor (i.e., if `usesMemo: false`). Let's avoid any conversion rate hurdles.

4. When I select EUR, then click on "Confirm" so that the button shows "Continue with Partner", then select ARS, then the button still shows "Continue with Partner" and I will start the Mykobo anchor flow. We should reset when switching the output currency.

signer-service/src/api/services/siwe.service.js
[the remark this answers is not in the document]
gianfra-t: I think that adding more complexity to this implementation, for instance the persistence after a crash, comes very close to a full session implementation which I would prefer to do with a library already.

I'm even dubious about the logout. Maybe we can skip it for now wdyt? And in the next iteration we use a proper session library.

gianfra-t:
Nice observations @TorstenStueber:

>  I would not expect that we need this new endpoint

It is true that the message can be created either in the backend, or in the frontend. But even when created on the fronted, we need an endpoint that sends a random nonce or a way to derive the nonce such that we are able to verify it on the service, which seems more complex. The nonce is to prevent replay attacks, more info [here](https://docs.login.xyz/additional-support/security-considerations).

Now it is true that the message could be generated on the frontend and then validate the values on the backend. See this [example](https://1.x.wagmi.sh/examples/sign-in-with-ethereum) from wagmi. Still, the nonce is fetched. So I thought it was better to abstract the message creation to the backend already. I don't see much benefits/cons on either way.

> Metamask shows the following warning

Yes, I thought of using vortex main domain, but I will try to find a way to remove the warning for the staging and previews.

> I need to sign in even when I just select EUR

We were discussing it [here](https://github.com/pendulum-chain/vortex/pull/257#discussion_r1838700337), if we add a signature when the app loads/address changes or just when clicking confirm and the asset requires it. So far it's enabled to test how it "feels".

> When I select EUR, then click on "Confirm"

I didn't think about this case! Will add a reset.

TorstenStueber:
As long as the user is not malicious, they will protected from replay attacks if we create the nonce in the frontend. For malicious users – as you point out – they can just create the message anyway and not use the backend.

If there are equal solutions, I would always pick the simpler one – in this case this means to handle this completely in the frontend. Also: this would make it easier to just use the current domain for the message so that Metamask does not show the warning.

> We were discussing it https://github.com/pendulum-chain/vortex/pull/257#discussion_r1838700337,

This is a user facing decision and we should escalate to @pendulum-chain/product. I will move the discussion elsewhere.

src/hooks/useMainProcess.ts
gianfra-t: Moved to a [context](https://github.com/pendulum-chain/vortex/pull/257/commits/22f23390e28bbcb3393dc5eacec4fa8ad5f68486), I followed the `events` one which I found quite clean. This is what you had in mind?

Is this more or less what you had in mind? I don't think we can move only partially the hook into a context.

gianfra-t:
@pendulum-chain/devs on the last [additions](https://github.com/pendulum-chain/vortex/pull/257/commits/e12b1fe126ae3e4fede2b7648c0f6b8421827cce) we switched to secure cookies from the local storage, to avoid leaking the signature.

signer-service/src/api/controllers/siwe.controller.js
```diff
@@ -0,0 +1,40 @@
...
+    await verifySiweMessage(nonce, signature);
+
+    const token = {
+      nonce,
+      signature,
+    };
+
+    res.cookie('authTokenSignature', token, {
+      httpOnly: true,
+      secure: false,
```
gianfra-t, line 30: Needs to be switched to `true`. Used for testing

signer-service/src/api/services/siwe.service.js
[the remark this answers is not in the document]
ebma: I agree and also would suggest we skip a logout for now. Doesn't add much value at this point.

src/hooks/useSignChallenge.ts
```diff
@@ -0,0 +1,116 @@
...
+
+      const signatureData: SiweSignatureData = {
+        signatureSet: true,
+        expirationDate: message.expirationTime!,
+      };
+
+      localStorage.setItem(storageKey, JSON.stringify(signatureData));
+      signPromise.resolve();
+    } catch (error) {
+      signPromise.reject(new Error('Signing failed'));
```
ebma, line 81: Nit: let's include the other error here. 
```suggestion
      signPromise.reject(new Error('Signing failed: ' + error.toString()));
```

src/hooks/useSignChallenge.ts
```diff
@@ -0,0 +1,116 @@
...
+    if (signPromise) {
+      signPromise.reject(new Error('User cancelled'));
+      setSignPromise(null);
+    }
+    setRequiresSign(false);
+  }, [signPromise, setRequiresSign, setSignPromise]);
+
+  const checkAndWaitForSignature = useCallback(async (): Promise<void> => {
+    const stored = checkStoredSignature();
+    console.log('stored', stored);
```
ebma, line 99: Nit: remove

src/hooks/useSignChallenge.ts
```diff
@@ -0,0 +1,116 @@
...
+import { SiweMessage } from 'siwe';
+
+export interface SiweSignatureData {
+  signatureSet: boolean;
+  expirationDate: string;
+}
+
+export function useSiweSignature(address?: `0x${string}`) {
+  const { signMessageAsync } = useSignMessage();
+  const [requiresSign, setRequiresSign] = useState(false);
```
ebma, line 14: I would rename this and all references to it to 
```suggestion
  const [showSigning, setShowSigning] = useState(false);
```
or 
```suggestion
  const [signingPending, setSigningPending] = useState(false);
```

so that this indicates how the variable is supposed to be used (ie. for the modal). Otherwise it's confusing.

src/hooks/useSignChallenge.ts
```diff
@@ -0,0 +1,116 @@
...
+      localStorage.removeItem(storageKey);
+      return null;
+    }
+  }, [address, storageKey]);
+
+  const signMessage = useCallback((): Promise<void> => {
+    return new Promise((resolve, reject) => {
+      setSignPromise({ resolve, reject });
+      setRequiresSign(true);
+    });
```
ebma, line 44: I think storing this promise is not necessary and I would avoid it. If you want to have the `handleSign` adjust based on a pending signature dialog or not, then it should be enough to check for `if (!requiresSign)` in l. 48, no?

ebma (commented):
As @TorstenStueber pointed out, creating the SiweMessage locally on the frontend might get rid of the warning shown by metamask (and probably other wallet extensions). If both solutions work, let's maybe still aim for the local message creation to avoid alerting the user.

signer-service/src/constants/constants.js
```diff
@@ -9,12 +9,14 @@ const SUBSIDY_MINIMUM_RATIO_FUND_UNITS = '10'; // 10 Subsidies considering maxim
 const MOONBEAM_RECEIVER_CONTRACT_ADDRESS = '0x0004446021fe650c15fb0b2e046b39130e3bfe36';
 const STELLAR_EPHEMERAL_STARTING_BALANCE_UNITS = '2.5'; // Amount to send to the new stellar ephemeral account created
 const PENDULUM_EPHEMERAL_STARTING_BALANCE_UNITS = '0.1'; // Amount to send to the new pendulum ephemeral account created
+const DEFAULT_EXPIRATION_TIME_HOURS = 24;
```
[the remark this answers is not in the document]
gianfra-t: That's a lot! But now that we use safe cookies I think there would be no issue.

signer-service/src/api/middlewares/validators.js
```diff
@@ -155,6 +155,27 @@ const validateSep10Input = (req, res, next) => {
...
+  if (!walletAddress) {
+    return res.status(400).json({ error: 'Missing address: walletAddress' });
+  }
+  next();
+};
+
+const validateSiweValidate = (req, res, next) => {
+  const { nonce, signature } = req.body;
+  if (!signature) {
+    return res.status(400).json({ error: 'Missing signature: signature' });
```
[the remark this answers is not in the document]
gianfra-t: Yes, looks bad 😅.

src/hooks/useSignChallenge.ts
ebma: I tried refactoring the code and while I could reduce parts of the `useSiweSignature` code overall, it didn't necessarily turn out any less complex or easier to read. So I guess we just have to live with some confusion/complexity around this part. I'm fine with keeping it the way you implemented it 👍

ebma:
I think the main comments are all addressed except for the part about having the Siwe creation on frontend vs backend. @gianfra-t are you still testing this or do you have hard feelings and would insist on keeping it the way it is right now?

gianfra-t:
@ebma I'm on it since this morning! Took a few more changes, but is almost there.

src/hooks/useSignChallenge.ts
gianfra-t: Okay! Out of curious, what did you attempted to remove the promise? `useEffect` ? Yeah the whole complexity fact is that the process is done inside the `sep10` function, sometimes is hard to mix hook and non hook processes like this.

signer-service/src/api/controllers/siwe.controller.js
```diff
@@ -29,12 +28,17 @@ exports.validateSiweSignature = async (req, res) => {
       httpOnly: true,
       secure: false, // TODO TODO TODO: Change to true in production
```
gianfra-t, line 29: Yes this will be removed right before merging.

signer-service/src/api/services/siwe.service.js
```diff
@@ -1,75 +1,139 @@
...
+
+  // Check if expiration is within a reasonable range from current time
+  const currentTime = new Date().getTime();
+  const expirationTimestamp = new Date(expirationTime).getTime();
+  console.log('expirationTimestamp', expirationTimestamp);
+
+  const expirationGracePeriod = 1000 * 60; // 1 minute
+  const expirationPeriodMs = DEFAULT_LOGIN_EXPIRATION_TIME_HOURS * 60 * 60 * 1000;
+  const expectedMinExpirationTimestamp = currentTime + expirationPeriodMs - expirationGracePeriod;
+  const expectedMaxExpirationTimestamp = currentTime + expirationPeriodMs;
```
ebma, line 108: So if I understand this correctly, you want to allow a range of 7days-1min < x < 7days for the expiration period? 

Wouldn't this check fail oftentimes for siwe messages that are not fresh? Let's say I am a user and I have an existing Siwe message that is valid for 7 days. If I authenticate with it one day after creating it, this logic here will throw an error because it's not within the min and max.

ebma: I think we can't really check for the min and max expiration time if we don't know the timestamp of the creation. 
However, generally this is more a problem of the user interface. If the user interface decides to create very short-lived siwe sessions, so be it. It's a bad UX but that's up to the developer of the frontend.

signer-service/src/api/services/siwe.service.js
```diff
@@ -1,75 +1,139 @@
...
+  if (expirationTimestamp < expectedMinExpirationTimestamp) {
+    throw new ValidationError('Expiration time is too low');
+  }
+
+  if (expirationTimestamp > expectedMaxExpirationTimestamp) {
+    throw new ValidationError('Expiration time is too high');
+  }
+};
+
+const initialVerifySiweMessage = async (nonce, signature, siweMessage) => {
```
ebma, line 119: Maybe rename to 
```suggestion
const verifyAndStoreSiweMessage = async (nonce, signature, siweMessage) => {
```
or similar? It's a bit difficult to see the logical difference between `verifySiweMessage` and `initialVerifySiweMessage` otherwise.

ebma (commented):
Didn't test it again locally but logic looks good overall 👍

signer-service/src/api/services/siwe.service.js
gianfra-t: Agreed. That will help the comments.

signer-service/src/api/services/siwe.service.js
gianfra-t: >  7days-1min < x < 7days

Yes exactly. Keep in mind that this check is done only on validation (initial verify and store) of the cookie. After that, the `verifyMessageFields` is not used, and we just check the expiration itself of the cookie and also of the signature/token itself.

The 1 min would then be the max delay that the UI has for sending the crafted message. This is all such that the message would be similar to one created on the backend.

But for the last point, you're right. The most important check is x < 7 days that will check a too-long expiration time. It's not a security concern that a UI sends short messages, but I think is good for being consistent.

signer-service/src/api/services/siwe.service.js
```diff
@@ -1,75 +1,139 @@
...
-
-  return { siweMessage: preparedMessage, nonce };
+  return { nonce };
 };
 
-exports.verifySiweMessage = async (nonce, signature) => {
+// Used to verify the integrity and validity of the signature
+// For the initial verification, the siweMessage must be provided as parameter
+const verifySiweMessage = async (nonce, signature, initialSiweMessage) => {
   const maybeSiweData = siweMessagesMap.get(nonce);
```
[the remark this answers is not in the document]
gianfra-t: I was missing Rust's Option type that day 😅. Sorry, changed.

signer-service/src/api/services/siwe.service.js
ebma: Oh right, I missed that it's only called by `verifyAndStoreSiweMessage()`. I'm okay if we keep it as is then but let's maybe rename `verifyMessageFields()` to `verifyInitialMessageFields()` to make the difference to `verifySiweMessage()` clearer 😬

ebma (approved):
I think it looks good 👍 @gianfra-t could you maybe do another test offramp?

TorstenStueber (approved):
I think that this solution is already a bit too professional and complex and the logic is harder to follow and verify this way.

But I approve this PR anyway. Great job.

signer-service/src/constants/constants.js
```diff
@@ -9,12 +9,25 @@ const SUBSIDY_MINIMUM_RATIO_FUND_UNITS = '10'; // 10 Subsidies considering maxim
...
+  'https://polygon-prototype-staging--pendulum-pay.netlify.app',
+  'https://app.vortexfinance.co',
+];
 
 require('dotenv').config();
 
 const PENDULUM_FUNDING_SEED = process.env.PENDULUM_FUNDING_SEED;
 const FUNDING_SECRET = process.env.FUNDING_SECRET;
 const MOONBEAM_EXECUTOR_PRIVATE_KEY = process.env.MOONBEAM_EXECUTOR_PRIVATE_KEY;
+const SEP10_MASTER_SECRET = FUNDING_SECRET;
```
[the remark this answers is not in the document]
gianfra-t: I think so, although I tried to offramp only with it and it didn't login immediately. Since Anclap is still doing testing (I believe with this account), and the code should not allow to sign a transaction without a memo for them, we can leave it like this for now.

signer-service/src/api/services/siwe.service.js
```diff
@@ -0,0 +1,140 @@
...
+    throw new ValidationError('Must define a valid expiration time');
+  }
+
+  // Check if expiration is within a reasonable range from current time
+  const currentTime = new Date().getTime();
+  const expirationTimestamp = new Date(expirationTime).getTime();
+
+  const expirationGracePeriod = 1000 * 60; // 1 minute
+  const expirationPeriodMs = DEFAULT_LOGIN_EXPIRATION_TIME_HOURS * 60 * 60 * 1000;
+  const expectedMinExpirationTimestamp = currentTime + expirationPeriodMs - expirationGracePeriod;
```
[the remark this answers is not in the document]
gianfra-t: Maybe, I can put 10 minutes, won't matter too much.

signer-service/src/constants/constants.js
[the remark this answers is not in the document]
TorstenStueber: Okay!

signer-service/src/api/services/siwe.service.js
```diff
@@ -0,0 +1,140 @@
...
+  verifyInitialMessageFields(validatedMessage);
+
+  // Verification complete. Update the map and append the message.
+  const siweData = siweMessagesMap.get(nonce);
+  siweData.siweMessage = validatedMessage;
+  siweMessagesMap.set(nonce, siweData);
+
+  // Remove messages with same address from map except for the one we just verified.
+  // This will keep one valid session per address.
+  siweMessagesMap.forEach((data, nonce) => {
```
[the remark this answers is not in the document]
gianfra-t: The original reasoning was to keep only one "session" per user, per address. This and the logout (that we decided not to do), where the original rationale for adding the map in the first place. Otherwise we would not really need it, it wouldn't add extra security.

Since it is relatively small amounts of logic, I decided to leave this functionality. Totally optional of course.

src/services/anchor/index.ts
```diff
@@ -61,19 +63,70 @@ export const fetchTomlValues = async (TOML_FILE_URL: string): Promise<TomlValues
...
+    if (!masterSep10Public) {
+      throw new Error('masterSep10Public not found in response.');
+    }
+
+    const sep10Account = masterSep10Public;
+
+    return {
+      urlParams: new URLSearchParams({
+        account: sep10Account,
+        client_domain: config.applicationClientDomain,
```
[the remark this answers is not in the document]
gianfra-t: Actually, yes. Thanks for flagging this inconsistency.

src/hooks/useSignChallenge.ts
```diff
@@ -0,0 +1,136 @@
...
+      const data: SiweSignatureData = JSON.parse(stored);
+      return new Date(data.expirationDate) > new Date() ? data : null;
+    } catch {
+      localStorage.removeItem(storageKey);
+      return null;
+    }
+  }, [address, storageKey]);
+
+  const signMessage = useCallback((): Promise<void> => {
+    return new Promise((resolve, reject) => {
```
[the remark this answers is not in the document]
gianfra-t: Should not be possible for it to be called at the same time with the current set-up, but you're right I will add it just in case to avoid potential future confusions.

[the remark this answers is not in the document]
TorstenStueber: Yes, even it is not possible right now (which I agree), it is better to protect again in this call for the code to be more robust in case it is changed somewhere else.

signer-service/src/api/services/siwe.service.js
```diff
@@ -0,0 +1,140 @@
...
+
+class ValidationError extends Error {
+  constructor(message) {
+    super(message);
+    this.name = 'SiweValidationError';
+  }
+}
+
+// Map that will hold the siwe messages sent + nonce we defined
+const siweMessagesMap = new Map();
```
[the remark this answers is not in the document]
gianfra-t: Yes the use of the map was what I was saying [here](https://github.com/pendulum-chain/vortex/pull/257#discussion_r1852905290). In retrospect, not super useful. It's the part of the solution least resilient and of course in production, should we want to keep the state, we should use redis or similar.

Regarding replay, I think it's safe even if the backend crashes and the state of the pair (nonce, address) is lost, then the worst case scenario will be an "invalid nonce" error. We never really check for repeated nonces, that apparently is not common practice in the implementations I've seen. The chances that attacker can use even a bunch of old signatures from a user, and gets the same nonce it got in the past must be quite small, and even then we would be protected by the expiration.

gianfra-t:
@ebma I tested another offramp and it works. I will merge this and then someone could test an EURC offramp, I can only tell that the anchor's window is opened.

vadaynujra:
@ebma I can do the EUR off ramp today using staging.

PR #268, opened 2024-11-08
Update wagmi and viem dependencies

### LINKED ISSUE #259: Bug: broken handling of aborted singing requests
This is an error that occurs on iOS Safari with the Metamask app.

When declining to sign a transaction (press "Cancel" on the popup that asks to open the wallet app), then the app will not recognize this – i.e., no event is reported to the app and the app stays in the state where it waits for the user to sign the transaction.

WhatsApp Image 2024-11-06 at 17 56 57

Only when reloading the page, an error will be shown in the UI and on the console:

WhatsApp Image 2024-11-06 at 18 41 19
Screenshot 2024-11-06 at 18 01 40

TorstenStueber:
Closes #259 

I resolved this by:
* updating the dependencies (hint from [this comment](https://github.com/rainbow-me/rainbowkit/discussions/2063#discussioncomment-9907115))
* only start the next offramping step when the wallet is already connected

I could not resolve the actual problem to detect when the user cancels opening the wallet app. I traced the execution and could ultimately trace the execution to [this function](https://github.com/WalletConnect/walletconnect-monorepo/blob/5776c2a234869687711595b83c828baf4f563a1f/packages/sign-client/src/controllers/engine.ts#L533) in wallet connect. This will instruct the browser to ask to open the wallet app. The promise will only resolve at [this point](https://github.com/WalletConnect/walletconnect-monorepo/blob/5776c2a234869687711595b83c828baf4f563a1f/packages/sign-client/src/controllers/engine.ts#L553-L559), where it waits for a `session_request` event.

If the user declines to open the app, then there is no other event detected and the function will neither resolve nor reject.

Anyway, with the changes in this PR the user will be safe to just reload the page and the popup will be shown again.


src/hooks/useMainProcess.ts
```diff
@@ -257,6 +257,8 @@ export const useMainProcess = () => {
   }, [updateHookStateFromState, offrampingState]);
 
   useEffect(() => {
+    if (wagmiConfig.state.status !== 'connected') return;
+
     (async () => {
       const nextState = await advanceOfframpingState(offrampingState, {
```
gianfra-t, line 263: So for me to understand, the app will "hang" at this stage if whatever step the advance function executes will request a signature, and the user rejects.
Could we add a timeout + refresh if "nothing happens" to avoid this? Potentially [here](https://github.com/pendulum-chain/vortex/blob/2e0dc8d83b6c068ddc34f2df8de449ebd400a06d/src/services/squidrouter/process.ts#L71). Would it make sense? 

Refresh or a modal

TorstenStueber: Sure, it's an option. Can do that.

TorstenStueber: There is just one thing to consider here. There are two situations where our website asks the browser to change the app: 1) when using wallet connect to connect the wallet and 2) when our code asks for signing requests. In both cases the browser shows a popup asking the user to switch the app.

I was testing for a long time yesterday and after some time I always observed something curious: the popup did not appear anymore in either one of the cases. For case 1) I could only make it show up by explicitly pressing the Retry button (see screenshots).





This looks to me like a security mechanism of Safari in order to protect the user from malicious website behavior. While testing I was repeatedly declining the request to open the app, as this is what I wanted to research. So Safari decided not to show the popup anymore, only in the context of a click handler.

The only way for me to turn on the popup was to change the domain or port, which suggests that Safari blocks by host.

The signing promise will only resolve once the user switched to the app and confirmed there. This can take a while, e.g., looking for a password to log into the app. If during this time we ask repeatedly in the background to switch the app again, it could be that Safari also considers this malicious. Since we start signing requests programmatically and not in click handlers, we risk that the user will never be able to sign anymore.

For that reason I would play it rather safe.

gianfra-t: Thanks for the thorough testing. So given all this the timer is either a bit pointless or even more confusing for the user. Then let's go with the solution you have right now.
I was thinking more on an absolute timeout in which the app will refresh completely, not just the open request. I don't know if your findings would also apply to this case.

But seems like a rather edge case the scenario in which the person rejects opening a new app but still waits for something to happen.

gianfra-t (approved):
Let's go! ✅

ebma (approved):

PR #270, opened 2024-11-08
Add url parameter to preselect tokens

### LINKED ISSUE #266: Add url parameter to preselect output token
In order to have efficient ad targeting, we the output token (and possibly input token) should be pre-selectable through url parameters.

## TODO
* upon startup, check whether there are url parameters `from` or `to`
  * if the url parameter `from` exists, then it will be one of the `InputTokenType` strings
    * if it is, then initially select that input token
  * if the url parameter `to` exists, then it will be one of the `OutputTokenType` strings
    * if it is, then initially select that output token
* if there are no parameters, then the default selection for the input token is `usdc`

TorstenStueber:
Closes #266

### DIFF
--- a/src/components/Nabla/schema.tsx
+++ b/src/components/Nabla/schema.tsx
@@ -8,8 +8,6 @@ export type SwapFormValues = {
   toAmount: string;
   slippage: number | undefined;
   deadline: number;
-  bankAccount: string;
-  taxNumber: string;
 };
 
 /* eslint-disable @typescript-eslint/no-explicit-any */
@@ -26,8 +24,6 @@ const schema = Yup.object<SwapFormValues>().shape({
   toAmount: Yup.string().required(),
   slippage: Yup.number().nullable().transform(transformNumber),
   deadline: Yup.number().nullable().transform(transformNumber),
-  bankAccount: Yup.string().required(),
-  taxNumber: Yup.string().required(),
 });
 
 export default schema;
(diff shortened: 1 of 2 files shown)


src/components/Nabla/schema.tsx
```diff
@@ -8,8 +8,6 @@ export type SwapFormValues = {
   toAmount: string;
   slippage: number | undefined;
   deadline: number;
-  bankAccount: string;
```
TorstenStueber, line 11: Just a cleanup. Not required anymore.

src/components/Nabla/useSwapForm.tsx
```diff
@@ -22,17 +22,25 @@ export const useSwapForm = () => {
...
 
-    return {
-      from: (storedFromTokenIsValid ? storageValues?.from : 'usdc') as InputTokenType,
-      to: (storedToTokenIsValid ? storageValues?.to : 'eurc') as OutputTokenType,
-      taxNumber: '',
-      bankAccount: '',
+    const initialValues = {
+      ...defaultValues,
+      ...searchParamValues,
+      ...storageValues,
```
gianfra-t, line 34: Shouldn't `searchParamValues` have higher priority than those stored? In case of a redirect.

TorstenStueber: I agree, changed.

gianfra-t (approved):
✅

PR #274, opened 2024-11-12
Fix wallet connection issues on mobile with appkit

### LINKED ISSUE #258: Bug: wallet connect error on mobile
This is an error that occurs when using Vortex on Safari iOS with the Metamask app.

The error occurred directly when signing the approval transaction in the Metamask app. The error occurs in the wallet connect source code and I could [this issue](https://github.com/orgs/WalletConnect/discussions/4233).

Screenshot 2024-11-06 at 17 35 38

When this error occurs, the app never continues to proceed to the second signature.

Once this error happens, it happens again every single time at exactly the same moment.

Then error disappears when I reconnected the wallet. However, after starting/aborting/executing one flow, it will happen again on every subsequent flow. Until I reconnect again.

Once this error shows up for signing requests, the same error is also reported when pressing the disconnect button in order to reconnect the wallet.

ebma:
### Changes

Replaces rainbowkit with appkit.

Closes #258.

### Findings

I did some tests and found that using the latest dependencies of wagmi/viem done in #268 in combination with Appkit works quite well. To elaborate, I found that the 'No matching key. History' error did not occur with Appkit ever. Neither for connecting the wallet nor for disconnecting nor when signing a message. Tested with and without an app refresh in between. 

It's still not perfect though. I noticed that signing requests don't always trigger the respective app. For example, when I connect to the prototype with WalletConnect and Zerion on my phone, the connection to the account works well. Zerion opens, let's me choose my account, etc. However, when a transaction signature is requested, Zerion **never** opens automatically. It just seems as if nothing happens to the user. When I open Zerion after the signing is requested (and it seems like nothing has happened) I can see the signing request in Zerion though. 
For Metamask it's similar but a little different. The wallet connection with Metamask works well bug when it comes to signing a transaction, _sometimes_ I see an intent where I can open Metamask and _sometimes_ I don't. When I don't see the intent to open the app, again, I can manually open Metamask and will see the request for signing. 

I found a related bug ticket https://github.com/reown-com/appkit/issues/1504 and it seems like this is not only an AppKit problem but also something related to wagmi. 
As a side note, I also found another bug ticket https://github.com/reown-com/appkit/issues/2076 mentioning that automatic redirect back to the calling application is not possible on iOS 17. This worked on my Android phone when using Metamask. 

### Outlook

All in all, I think this is an improvement over the connection with rainbowkit. I tested the current version on polygon-prototype-staging again (which has rainbowkit + the latest dependencies of wagmi/viem) and I still encountered the 'No matching key. History' and other similar errors. This didn't happen to me with Appkit.
Thus, I would suggest we try to deal with the shortcomings on Appkit, possibly showing a notification to the user when a signing is requested that if they don't see any action appearing on screen, they should check their connected wallet application manually. It's not a great UX but at least it works technically.


gianfra-t:
Tested this: 

Setup: 
Motorola edge 30 ultra. Chrome + Metamask

First attempt:
I connected a wallet and it consistently worked, meaning click on confirm prompt me to move to Metamask and sign.

Forcing a failure:
Nevertheless, I then disconnected from my wallet on Vortex, connected again by creating a new account on Metamask.
After this, clicking on confirm was consistently "failing", yet in Metamask I could see the request.

After a few attempts at changing back again to my first wallet (which was failing many times but finally succeeded), I changed back and forth from my first address and my second. Seems to be that always with my second address, the prompt to Metamask would not show while with the first it would.

ebma:
> Seems to be that always with my second address, the prompt to Metamask would not show while with the first it would.

That's an interesting observation 🤔 Could still be that it's a coincidence. As for me, I always tested with the same account and sometimes the prompt worked, sometimes it did not. 

But at least you can confirm that with manually opening the app, signing was always possible.

gianfra-t:
> But at least you can confirm that with manually opening the app, signing was always possible.

yes so far that's the case in this example.

TorstenStueber:
I tested on iOS Safari.

Generally I can't properly switch between wallets. The main menu to select wallets looks as follows:
WhatsApp Image 2024-11-13 at 18 06 52(1)

The "Browser Wallet" entry just shows the following page (the explanation is confusing – I don't have a browser wallet installed, so it is to be expected that none is found here):
WhatsApp Image 2024-11-13 at 18 06 53

The "All Wallets" entry leads to a long list of browser, however, 1) it does not contain Phantom and 2) it is not able to determine that I have Rabby wallet installed. This is unexpected as the "Browser Wallet" logo in the main menu indicates that both wallets should be well supported.

**Metamask**
* popup to change app always appear
* switch to app and display of signing request always happens
* after signing it works most of the time (I see something like `approvalHash <hash>` on the console), but sometimes I instead see this error message:
  Screenshot 2024-11-13 at 16 57 44
* when I reject the transaction in the wallet, the following messages shows in the webapp log:
  Screenshot 2024-11-13 at 17 00 22
    
**Trust Wallet**
* popup to change app always appear
* switch to app and display of signing request always happens
* after signing it works most of the time (I see something like `approvalHash <hash>` on the console), but sometimes I instead see this error message (however, some time after this message appears it seems that the `approvalHash <hash>` log message is posted as well):
  Screenshot 2024-11-13 at 17 27 15
* when I reject the transaction in the wallet, the following messages shows in the webapp log:
  Screenshot 2024-11-13 at 17 44 25


**Zerion Wallet**
* popup to change app frequently doesn't appear, although the transaction occurs in the wallet to sign when switching to it
* when singing, the behavior is similar to the other wallets above

I wanted to test Phantom and Rabby wallet but couldn't for the reasons at the beginning of my comment.

The selected address at the top of the screen overlaps with the logo on mobile:

WhatsApp Image 2024-11-13 at 18 16 46

TorstenStueber:
In general: I think this works better than rainbowkit, but I would like to test Rabby and Phantom as well, in order to see whether the issues with Zerion are an exception.

We should test the proper flow next in order to see whether switching to the anchor UI causes some extra issues.

ebma:
@TorstenStueber it's possible to define wallets that _need_ to show up in the list. I already did this for Metamask here as it was also missing in the list for some reason. I'll try to do the same for Rabby and Phantom so you can test with them as well. 
I'll also remove the custom logic for early signing so it can be tested with the 'normal' flow.

ebma:
@TorstenStueber you can give this another try. I added the config for rabby and phantom and restored the handler for the 'Confirm' button.

ebma:
I think that's because neither Rabby nor Phantom appear as 'WalletConnect certified' on the list [here](https://walletguide.walletconnect.network/). 

image

So the walletconnect support with Appkit seems to be limited to [these](https://walletguide.walletconnect.network/?badge_type=certified) wallets. Which can be good or bad. Assuming that the 'WalletConnect certified' means that the support is tested, it would ensure that we only show options that are confirmed to work.

ebma:
The code is ready for review.

gianfra-t (commented):
Looks good. Also, tested this config previously and worked well.

gianfra-t (approved):
👍

ebma:
> did anyone also test on Desktop? Anything not working there?

I missed that message before. I tested this on desktop as well and it works well for me. Tested both the walletconnect option connecting to the desktop browser with the phone (using Zerion) and the usual connection with Metamask and transaction signing was requested as expected.

@pendulum-chain/product do you also want to check out the deploy preview? As the underlying library changed, the connector dialog now looks different. Please have a look and let us know if it's good to go from your side.

src/wagmiConfig.ts
```diff
@@ -26,9 +13,46 @@ const transports = config.alchemyApiKey
...
 
-export const wagmiConfig = createConfig({
-  chains: [polygon],
-  connectors,
+// 2. Create a metadata object - optional
+const metadata = {
+  name: 'Vortex',
+  description: 'Vortex',
+  url: 'https://app.vortexfinance.co', // origin must match your domain & subdomain
+  icons: ['https://avatars.githubusercontent.com/u/179229932'],
```
TorstenStueber, line 21: What is the significance of this?

TorstenStueber (approved):
👍 Great change, let's keep fingers crossed that Rabby (and Phantom) will be supported properly soon.

src/wagmiConfig.ts
ebma: Nice catch, I forgot about this. If you specify the metadata, the wallet app can show additional info when the WalletConnect connection is requested. I removed the icon but while `icons` cannot be `undefined` it seems to be okay when it's an empty array.

TorstenStueber (approved):

Sharqiewicz (approved):

src/components/buttons/ConnectWallet/index.tsx
```diff
@@ -1,93 +1,80 @@
...
+          className="btn-vortex-secondary btn rounded-3xl group"
+        >
+          <p className="flex">
+            Connect <span className="hidden lg:block lg:ml-1">Wallet</span>
+          </p>
+          <PlayCircleIcon className="w-5 group-hover:text-pink-600" />
+        </button>
+      );
+    } else if (!isOnSupportedNetwork) {
+      return (
```
Sharqiewicz, line 42: We could extract these buttons and then just simply do:
`else if (!isOnSupportedNetwork) return <ButtonWrongNetwork/>`

vadaynujra:
Tested on desktop Windows + Chrome. 
Got second signing request.

src/components/buttons/ConnectWallet/index.tsx
ebma: We could but do you think it's worth it? All three buttons differ in more than just text as they also have different stylings.

Sharqiewicz: It improves the readability but it is optional, that's why I approved 😄

TorstenStueber (commented):
👍

ebma:
For some reason, the deploy preview failed when building the first time but clearing the cache and retrying a second time worked, see [here](https://app.netlify.com/sites/pendulum-pay/deploys/673759cbdde7c499ed81aca7). So I'll merge this regardless.

TorstenStueber (approved):

PR #280, opened 2024-11-13
Fix for incorrect assetCode property

gianfra-t:


signer-service/src/api/controllers/subsidize.controller.js
```diff
@@ -49,7 +49,7 @@ exports.subsidizePostSwap = async (req, res) => {
     const assetIssuerHex = `0x${Keypair.fromPublicKey(assetIssuer).rawPublicKey().toString('hex')}`;
     const pendulumCurrencyId = {
       Stellar: {
-        AlphaNum4: { code: assetCode.padEnd(4, '\0'), issuer: assetIssuerHex },
+        AlphaNum4: { code: assetCodeRaw.padEnd(4, '\0'), issuer: assetIssuerHex },
```
TorstenStueber, line 52: Isn't the idea that we don't need `.padEnd(4, '\0')` for `assetCodeRaw`?

TorstenStueber (commented):
Isn't it almost more confusing/error prone now that we have both `assetCode` and `assetCodeRaw`?

I propose that we just define `assetCode` (without padding) in the config and then have a convenience function that derives the Pendulum asset from that. This should be the only place where we need to pad the `assetCode` with the zero byte (?)

signer-service/src/api/controllers/subsidize.controller.js
gianfra-t: Yes you're right here, should remove the pad (which should have never been needed, we were already padding on the config).

signer-service/src/constants/tokenConfig.js
```diff
@@ -46,7 +46,7 @@ const TOKEN_CONFIG = {
 
 function getTokenConfigByAssetCode(cofig, assetCode) {
   for (const key in cofig) {
```
ebma, line 48:
```suggestion
  for (const key in config) {
```

src/services/polkadot/index.tsx
```diff
@@ -202,7 +202,7 @@ function checkBalancePeriodically(
       try {
         const someBalanceUnits = await getStellarBalanceUnits(
           stellarTargetAccountId,
-          outputToken.stellarAsset.code.stringRaw,
```
gianfra-t, line 205: This one was wrong to have been `raw` in the first place, as it compares with the return of the stellar balance query.

TorstenStueber (approved):
Thanks for implementing the change requests!

PR #284, opened 2024-11-18
Add AssetHub flow for offramping USDC

ebma:
## TODO

- [ ] Replace signing with external wallet signer. 
- [ ] Adjust network selection (currently hardcoded)

These are blocked by https://github.com/pendulum-chain/vortex/issues/271.

Closes #283.

PR #289, opened 2024-11-20
wip support polkadot wallet connect on vortex

Sharqiewicz:


Sharqiewicz:
@ebma This merge turned out to be huge, I'll clean it and describe Monday morning

src/pages/swap/index.tsx
```diff
@@ -414,7 +425,7 @@ export const SwapPage = () => {
           )}
         </div>
       </form>
-      {showCompareFees && fromToken && fromAmount && toToken && (
+      {showCompareFees && fromToken && fromAmount && toToken && fromToken.network === 'Polygon' && (
```
Sharqiewicz, line 428: unify this with a separate function

Sharqiewicz:
@pendulum-chain/devs Ready for review 🟢

src/components/NetworkSelector/index.tsx
```diff
@@ -0,0 +1,44 @@
+import { motion } from 'framer-motion';
+import { useState } from 'preact/hooks';
+
+import { NetworkIcon } from '../NetworkIcon';
+import { NetworkIconType } from '../../hooks/useGetNetworkIcon';
+import { Networks, useNetwork } from '../../contexts/network';
+
+const capitalize = (str: string) => str.charAt(0).toUpperCase() + str.slice(1);
+
+export const NetworkSelector = () => {
```
ebma, line 10: There is no click-away logic that would make the selector close when the user clicks elsewhere. Let's try to add that.

src/constants/constants.ts
```diff
@@ -1,10 +1,10 @@
 import { config } from '../config';
 
 export const HORIZON_URL = 'https://horizon.stellar.org';
-//export const HORIZON_URL = "https://horizon-testnet.stellar.org";
 export const BASE_FEE = '1000000';
 export const PENDULUM_WSS = 'wss://rpc-pendulum.prd.pendulumchain.tech';
-//export const PENDULUM_WSS = 'ws://localhost:8000';
+export const ASSETHUB_WSS = 'wss://polkadot-asset-hub-rpc.polkadot.io';
+export const ASSETHUB_ID = 'polkadot:68d56f15f85d3136970ec16946040bc1';
```
ebma, line 7: Let's add a comment to make clear that this is used for walletconnect.

src/contexts/network.tsx
```diff
@@ -0,0 +1,79 @@
...
+import { NetworkIconType } from '../hooks/useGetNetworkIcon';
+import { ASSETHUB_ID } from '../constants/constants';
+
+export enum Networks {
+  AssetHub = 'AssetHub',
+  Polygon = 'Polygon',
+}
+
+interface NetworkContextType {
+  polkadotSelectedNetworkId: string;
```
ebma, line 15: Again let's somehow make clear that this is only relevant for WalletConnect. Maybe rename the variable instead of adding a comment.

ebma (commented):
Really great changes 🙏 well done @Sharqiewicz some parts remind me of my changes in https://github.com/pendulum-chain/vortex/pull/284, maybe worth pointing that out in the description 😅 (assuming it's no coincidence lol). Also worth mentioning that this already points to the new Nabla instance as that's not obvious. 

I'd appreciate if we move the API connection to AssetHub somewhere so that it's shared instead of always reconnecting in the useEffect hook. Besides that it looks good to me 👍

src/assets/coins/ASSETHUB_POLYGON.svg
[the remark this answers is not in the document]
Sharqiewicz: haha, definitely

Sharqiewicz:
@ebma When USDT is selected and I try to change the network from Polygon to AssetHub there is an error, I'm implementing a fix

Sharqiewicz:
@ebma It is ready ✅

src/hooks/useGetAssetIcon.tsx
```diff
@@ -4,7 +4,7 @@ import USDC from '../assets/coins/USDC.png';
 import USDT from '../assets/coins/USDT.svg';
 import USDC_POLYGON from '../assets/coins/USDC_POLYGON.svg';
 import USDT_POLYGON from '../assets/coins/USDT_POLYGON.svg';
-import ASSETHUB_POLYGON from '../assets/coins/ASSETHUB_POLYGON.svg';
+import ASSETHUB_POLYGON from '../assets/coins/USDC_ASSETHUB.svg';
```
ebma, line 7: Let's also rename the import 😅 
```suggestion
import USDC_ASSETHUB from '../assets/coins/USDC_ASSETHUB.svg';
```

ebma (commented):
Thanks for adding the improvements 👌 

Two things: 
1) I would try to unify the way we handle the Assethub API and the Pendulum API. We need to keep two polkadot.js API objects around, one for getting AssetHub and one for Pendulum. You created a context for the Assethub API (which is nice) but the Pendulum API is managed by a `ApiManager` singleton in `polkadotApi.tsx`. We should either put both into the context or both in the `ApiManager`. I would prefer to put them both into the new context you created as that's cleaner than having the singleton. This requires some refactoring of the existing calling functions though. What would you prefer @Sharqiewicz?

2) It would also be nice to have a click-away for the network selector. It works for the account card though 👍

Sharqiewicz:
@ebma I agree that unifying the handling of substrate-related APIs is crucial for consistency and maintainability. I’ve removed the singleton and refactored the related files to be hooks and integrated usePendulumNode context

TorstenStueber:
Some comments about the UI/UX (also for @prayagd):
* shall we not call it "Polkadot AssetHub" instead of "AssetHub"?
* wouldn't it be more clear to also write the name of the currently selected network in the network select button instead of only the logo?
* wouldn't it make sense to also have a down chevron in the network select button to indicate that it is a dropdown list (similar to the token select buttons)?
* When selecting AssetHub, then only the "Connect Wallet" button in the upper right-hand corner lets me select Polkadot compatible wallets. The "Connect Wallet" button in the input form still connects to EVM wallets. I can then connect with Metamask without problems. Then I the "Connect Wallet" button in the upper right-hand corner still says "Connect Wallet". 
* When selecting AssetHub and then pressing the "Connect Wallet" button in the upper right-hand corner to connect to a Polkadot compatible wallet, then even if this works, the main button in the form still reads "Connect Wallet" and, when clicking it, wants me to connect to an EVM wallet.
* If connected to Metamask (when selecting Polygon) and I enter a valid input amount, the "Confirm" button stays disabled and I just see the error message "Something went wrong" under the output field.

Sharqiewicz:
@ebma The Network Selector still needs to be improved. When the user selects the network, the dropdown is not closing right now.

ebma (approved):
@Sharqiewicz thanks for merging the assethub and pendulum API components 👍  
I changed it so that the name of the network is shown in the selector. And I also refactored some things because I noticed some unnecessary rerenders.

I generally don't mind the hookyfied functions that we have now for the offramping flow. 

For the network selector, the click-away is working now but it doesn't automatically close when you click on it. I tried to fix it but it's not easy. The `Dropdown` component has a prop `open` but somehow it doesn't really behave as expected. I found [this](https://github.com/daisyui/react-daisyui/issues/412) issue ticket but switching between the `Dropdown.Details` or `Dropdown` didn't really make a difference. So I'd say we just live with the way it is now.

ebma:
The last 5 commits add the logic for the AssetHub transfer that were originally implemented in #284. As this PR already used some of the code of #284, I decided to just add the rest so that this PR now covers the whole AssetHub flow. Merging this PR in isolation wouldn't make much sense anyway as the AssetHub USDC would be selectable but not usable.

gianfra-t (commented):
Thanks for all these additions @Sharqiewicz and @ebma ! 🎉  To me is a go but I would like just to know you opinions on two comments I left: funding operation [before signing](https://github.com/pendulum-chain/vortex/pull/289#discussion_r1862572808), and the [pre-flight check](https://github.com/pendulum-chain/vortex/pull/289#discussion_r1862547580) of the polkadot API connection.

I tried to test the preview, but when I click confirm I get immediately a `Application initialization failed. Please reload, or try again later if the problem persists`. I didn't try to explore why yet, as maybe this is expected on this stage. The EVM side works, and also changing networks seems to work well 👍.

TorstenStueber (changes requested):
Due to the complexity of this PR I could not do a deep review.

I tried to test the flow but I always run into the following problem when selecting AssetHub and pressing the confirm button: the error message "Application initialization failed. Please reload, or try again later if the problem persists." will display. The reason is

```
Error: Token USDC is not supported on EVM chains
    at testRoute (route.ts:159:27)
    at Object.onConfirm (index.tsx:191:7)
    at HTMLFormElement.j (chunk-EUTVY7WI.js?v=10f33d6a:137:33)
```

prayagd:
Tested the deploy
- The limits dont work, i can input lower amounts than 10 EUR and click on confirm
- Compare quote feature is not visible
- If i join using polygon network and click on "Confirm" and then switch the network to Assethub, I still see the continue with partner button. It does not refresh's or starts from beginning
image

ebma:
> Compare quote feature is not visible

@prayagd this is because none of the three quote providers are able to offramp assets from AssetHub. So we have two options: 1) not show the compare quotes feature on AssetHub at all or 2) show the compare quotes table and show 'N/A' or a different wording in each row. What would you prefer?

vadaynujra:
@ebma, let's add comparative quotes from Polygon and state that somewhere in the UI.

Sharqiewicz:
@vadaynujra So should we show Polygon's comparison prices for AssetHub?

vadaynujra:
Yes @Sharqiewicz, we will show the Polygon quote, and mention Polygon in the text as `Sending 11 USDC (Polygon) with`. When the user hovers on Polygon, they see a tooltip message saying: _Quotes are for Polygon, as the providers don't support Asset hub yet._

Sharqiewicz:
@ebma @prayagd @gianfra-t 
When user submits the swap by clicking on "Confirm" button, the "Continue with Partner" button appears. When the user switches between the networks the "Continue with Partner" remains - this is because the request to the partner is updated automatically.

prayagd:
@pendulum-chain/devs any estimate how long this will take to go live?

ebma:
The last blocker we have is some bugs in the code we haven't managedto work around yet. Hard to tell.

src/pages/swap/index.tsx
```diff
@@ -351,8 +369,13 @@ export const SwapPage = () => {
   }
 
   if (offrampingState !== undefined || offrampingStarted) {
+    const isAssetHubFlow =
```
gianfra-t, line 372: Can we maybe name this `showMainScreenOnAssetHub` or something similar.

gianfra-t (approved):
I reviewed the new changes and it looks very good to me! Apart from the comment I left [here](https://github.com/pendulum-chain/vortex/pull/289#discussion_r1878640122), I see no issue, and that is maybe not so critical as the offramp won't start anyway. 

I didn't try yet to offramp from the last commit, but I was able to do so before and it worked.

ebma (approved):
I'll approve and merge this so we can test it out on staging.

Issue #46, opened 2024-06-20
Integrate google analytics query into the prototype


prayagd:
#### Context
We need to implement event tracking on Vortex PMF prototype to enable data collection via Google Tag Manager (GTM) and Google Analytics. Code template to create and push events to the data layer for the specified list of events was provided: 

```
document.getElementByID("ID").addEventListener('click', function(){
    dataLayer.push({
      'event': 'eventName', //such as select_vault_manually
      'label': 'vaultID', // use actual vaultID
  });
})
```

#### Requirements
- [ ] add the code to the events described in the [document](https://docs.google.com/document/d/1qL6DELne14fpPSZuS5kKhwx8FEGbJb4n/edit#heading=h.gjdgxs)

### Note

Header code

```
<!-- Google Tag Manager -->
<script>(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
'https://www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);
})(window,document,'script','dataLayer','GTM-T8JZSLD8');</script>
<!-- End Google Tag Manager -->
```

Footer Code:
```
<!-- Google Tag Manager (noscript) -->
<noscript><iframe src="https://www.googletagmanager.com/ns.html?id=GTM-T8JZSLD8"
height="0" width="0" style="display:none;visibility:hidden"></iframe></noscript>
<!-- End Google Tag Manager (noscript) -->
```

PR #290, opened 2024-11-20
Track amount type event 

### LINKED ISSUE #288: Bug: `amount_type` event not tracked anymore
The `amount_type` event should be tracked whenever the user interacts with the input field (see [this original ticket](https://github.com/pendulum-chain/vortex/issues/46)).

This is not the case anymore (if it ever worked): there is no code that ever tracks that event.

ebma:
As described [here](https://github.com/pendulum-chain/vortex/issues/46#issuecomment-2273198079):

> event amount_type: sent when user edits the "from amount" field for the first time

This PR adds changes so that the event `amount_type` is emitted when the user changes the value of the `fromAmount` field. Since `amount_type` is included in the `UNIQUE_EVENT_TYPES`, the event will only get tracked once, even if `trackEvent()` fires multiple times. 

--- 
Closes #288.

### DIFF
--- a/src/components/AssetNumericInput/index.tsx
+++ b/src/components/AssetNumericInput/index.tsx
@@ -9,6 +9,7 @@ interface AssetNumericInputProps {
   assetIcon: AssetIconType;
   tokenSymbol: string;
   onClick: () => void;
+  onChange?: (e: KeyboardEvent) => void;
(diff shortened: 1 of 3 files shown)


TorstenStueber (approved):
👍

PR #295, opened 2024-11-21
Remove trailing "/" on allowed origins.

gianfra-t:


gianfra-t:
@pendulum-chain/devs I believe the backend is blocking the staging URL just because of the extra "/".

TorstenStueber (approved):
👍

PR #296, opened 2024-11-22
Set to SameSite to None for auth cookie

gianfra-t:
Another unforeseen fix required after #257 

Cookies were set to `SameSite=Strict`, the issue is that our backend and frontend don't share the same home domain, and the cookie cannot be set on the deployment.

Regarding security:
This would open the possibility to CSRF attacks BUT we are also using CORS and limiting requests to the backend only from our deployments, so there should be no risk of that.

### DIFF
--- a/signer-service/src/api/controllers/siwe.controller.js
+++ b/signer-service/src/api/controllers/siwe.controller.js
@@ -26,8 +26,8 @@ exports.validateSiweSignature = async (req, res) => {
 
     res.cookie('authToken', token, {
       httpOnly: true,
-      secure: false, // TODO TODO TODO: Change to true in production
-      sameSite: 'Strict',
+      secure: true,
+      sameSite: 'None',
       maxAge: DEFAULT_LOGIN_EXPIRATION_TIME_HOURS * 60 * 60 * 1000,
     });
 


TorstenStueber (approved):
I will approve this.

If we want a safer option, we could alternatively add [rewrite rules to Netlify](https://docs.netlify.com/routing/redirects/#syntax-for-the-redirects-file) so that the backend always has the same domain as the frontend. E.g.:

1. create a file `_redirects` in the root
  ```
  /api/*  https://prototype-signer-service-polygon.pendulumchain.tech/:splat  200
  /*    /index.html   200
  ```
  
2. Ensure that this file is copied during the Netlify build process into `/dist`: e.g. `cp -R _redirects dist/_redirects`.

3. Just define `SIGNING_SERVICE_URL` to be `/api` in the frontend to refer to the backend.

Extra work required 
* we would need to ensure to point to the correct backend instance, but we could add different rewrite rules but we could use different rewrite rules and `SIGNING_SERVICE_URL`, e.g., `/api/staging`, `/api/production`, etc.

gianfra-t:
Oh that would be very nice. I didn't know it was possible. Certainly a good idea for production.

PR #299, opened 2024-11-22
Add new events to google analytics

### LINKED ISSUE #279: Add new events to google analytics
## Context 
There are some new events that are missing for tracking, this ticket to implement those

## Requirement
* For the "to_amount" from the "transaction confirmation" event, currently captures the Forex AMM quote replace it with Final quote after fees. see reference
* Add one parameter to the "transaction failure" event, called "error_msg" which captures the error why the transaction failed
* Add one more event "compare_quote", when the user clicks on the "Compare fees" button
    * `from_asset`
    * `to_asset`
    * `from_amount`
    * `to_amount`
    * `transak_quote`
    * `moonpay_quote`
    * `alchempy_quote`

## Reference
image

ebma:
#### New event
- `GA4 Event | compare_quote`
   - parameters:
      -  `{{ Transaction Event Variables }}`
      - `transak_quote`: `{{DLV | transak_quote}}`
      - `moonpay_quote`: `{{DLV | moonpay_quote}}`
      - `alchemypay_quote`: `{{DLV | alchemypay_quote}}`
   - trigger
      - `DL Event | compare_quote` 

#### Changed event
- `GA4 Event | transaction_failure`
   - changed parameters 
      - `error_message`: `{{DLV | error_message}}` 

## Logic changes
- Changes the emitted value of `to_amount` to the amount after fees
- Adds a new field `message` to `FailureType` so that the error message is available on the failure page (and is available  to emit in the respective event)
- The return values of `useSwapForm` were memoized to prevent infinite rerenders of the new `useEffect` hook in the `FeeProviderRow` component
- A new function `scheduleQuote()` was added to the events context. This function is used to accumulate the received quotes from within the `FeeProviderRow`s and emit the event only once all quotes are ready. As the quotes are only fetched and available in the `FeeProviderRow` components, we would need to lift this state out of these components if we wanted to implement it differently.

Closes #279.

### DIFF
--- a/src/contexts/events.tsx
+++ b/src/contexts/events.tsx
@@ -43,7 +43,7 @@ export interface WalletConnectEvent {
   account_address?: string;
 }
 
-interface OfframpingParameters {
+export interface OfframpingParameters {
   from_asset: string;
   to_asset: string;
   from_amount: string;
@@ -140,32 +140,17 @@ const useEvents = () => {
   const previousChainId = useRef<number | undefined>(undefined);
   const userClickedState = useRef<boolean>(false);
 
-  const [scheduledQuotes, setScheduledQuotes] = useState<Partial<Record<QuoteService, string>>>({});
+  const [_scheduledQuotes, setScheduledQuotes] = useState<
+    | {
+        from_asset: string;
+        to_asset: string;
+        quotes: Partial<Record<QuoteService, string>>;
+      }
+    | undefined
+  >(undefined);
 
   const trackedEventTypes = useRef<Set<EventType>>(new Set());
   const firedFormErrors = useRef<Set<FormErrorEvent['error_message']>>(new Set());
-
-  const scheduleQuote = useCallback((service: QuoteService, quote: string, state: OfframpingState) => {
-    setScheduledQuotes((prev) => {
-      const newQuotes = { ...prev, [service]: quote };
-      const sizeChanged = Object.keys(newQuotes).length !== Object.keys(prev).length;
-
-      // If all quotes are ready, emit the event
-      if (sizeChanged && Object.keys(scheduledQuotes).length === 3) {
-        trackEvent({
-          ...createTransactionEvent('compare_quote', state),
-          event: 'compare_quote',
-          transak_quote: newQuotes.transak,
-          moonpay_quote: newQuotes.moonpay,
-          alchemypay_quote: newQuotes.alchemypay,
-        });
-        // Reset the quotes
-        return {};
-      }
-
-      return newQuotes;
-    });
-  }, []);
 
   const trackEvent = useCallback((event: TrackableEvent) => {
     if (UNIQUE_EVENT_TYPES.includes(event.event)) {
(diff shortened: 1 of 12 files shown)


src/contexts/events.tsx
```diff
@@ -40,21 +43,35 @@ export interface WalletConnectEvent {
...
 
 export type TransactionEvent = OfframpingParameters & {
-  event: 'transaction_confirmation' | 'kyc_started' | 'kyc_completed' | 'transaction_success' | 'transaction_failure';
+  event:
+    | 'transaction_confirmation'
+    | 'kyc_started'
+    | 'kyc_completed'
+    | 'transaction_success'
+    | 'transaction_failure'
+    | 'compare_quote';
```
TorstenStueber, line 60: I think this is not intended. The event where `event` is `compare_quote` is the `CompareQuoteEvent` (as defined further below) and requires additionally the `..._quote` entries.

The only other reason to list it here is to make it callable for `createTransactionEvent`, but we don't do this anyway.

src/contexts/events.tsx
```diff
@@ -122,6 +140,14 @@ const useEvents = () => {
   const previousChainId = useRef<number | undefined>(undefined);
   const userClickedState = useRef<boolean>(false);
 
+  const [_scheduledQuotes, setScheduledQuotes] = useState<
```
TorstenStueber, line 143: Since we never read `_scheduledQuotes` directly but only inside the `setScheduledQuotes` update function, we could alternatively just make this a React ref. That way no rerender would be triggered when `scheduleQuote` is executed as no state is changed.

src/pages/swap/index.tsx
```diff
@@ -264,7 +264,7 @@ export const SwapPage = () => {
         <UserBalance token={fromToken} onClick={(amount: string) => form.setValue('fromAmount', amount)} />
       </>
     ),
-    [form, fromToken, setModalType],
+    [form, fromToken, setModalType, trackEvent],
```
TorstenStueber, line 267: Why do we need to add `trackEvent` here?

TorstenStueber (approved):
👍 I like the solution with the asynchronous firing of the `compare_quote` event.

Shouldn't we also use the actual output amount (after fees) for `kyc_started`? I think this is the only event that is still showing the old value.

ebma:
I noticed that there is an issue where quotes are scheduled prematurely, ie. the `from_amount` did not match the `to_amount` (the `to_amount` was using a stale/previous vortex price). I fixed it by keeping a reference to the previous vortex price.

The `kyc_started` event also uses the amount after fees, see [here](https://github.com/pendulum-chain/vortex/blob/5234c160316fc19426fda15c72fd22dc084394c9/src/hooks/useMainProcess.ts#L225).

src/contexts/events.tsx
ebma: Refactored to use a ref instead.

src/pages/swap/index.tsx
ebma: Because `trackEvent` is called above. After merging the latest changes into this branch it's not part of the PR anymore though.

TorstenStueber (approved):
👍

PR #300, opened 2024-11-25
Add more robust fix for cookies

### LINKED ISSUE #297: Add a more robust fix for authCookie attributes.
After the integration with ARS (#257) and the log in function, we are experiencing a bug on browser with strict privacy and security policies. 

On Firefox with strict settings, we see on the console after signing:

```
Cookie “authToken” has been rejected because it is foreign and does not have the “Partitioned“ attribute. [validate](https://prototype-signer-service-polygon-staging.pendulumchain.tech/v1/siwe/validate)

Cookie “authToken” has been rejected because it is foreign and does not have the “Partitioned“ attribute.
```

### Possible fixes:

- Try to set the Partitioned attribute (maybe not supported by express, has to be added manually in that case)
- Implement the solution suggested [here](https://github.com/pendulum-chain/vortex/pull/296#pullrequestreview-2454340447) and set again `SameSite=Strict,Lax`.

TorstenStueber:
Closes #297

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,10 +37,15 @@ matter what URL the browser requests.
 
 ## Env Variables
 
-- `VITE_SIGNING_SERVICE_URL`: Optional variable to point to a specific signing backend service URL. If undefined, it
+- `VITE_SIGNING_SERVICE_PATH`: Optional variable to point to a specific signing backend service URL. If undefined, it
   will default to either:
-  - http://localhost:3000 (if in development mode)
-  - https://prototype-signer-service-polygon.pendulumchain.tech (if in production mode)
+  - `http://localhost:3000` (if in development mode)
+  - `/api/production` (if in production mode)
+    - this will use the `_redirects` file to direct Netlify to proxy all requests to `/api/production` to
+      `https://prototype-signer-service-polygon.pendulumchain.tech`
+  - `/api/staging` (if in staging mode)
+    - this will use the `_redirects` file to direct Netlify to proxy all requests to `/api/staging` to
+      `https://prototype-signer-service-polygon-staging.pendulumchain.tech`
 - `VITE_ALCHEMY_API_KEY`: Optional variable to set the Alchemy API key for the custom RPC provider. If undefined, it
   will use dhe default endpoint.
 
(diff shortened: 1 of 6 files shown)


TorstenStueber:
With the change from https://github.com/pendulum-chain/vortex/pull/302 this PR here can now be reviewed and tested.

src/constants/constants.ts
```diff
@@ -14,5 +14,4 @@ export const AMM_MINIMUM_OUTPUT_HARD_MARGIN = 0.05;
 export const TRANSFER_WAITING_TIME_SECONDS = 6000;
 export const DEFAULT_LOGIN_EXPIRATION_TIME_HOURS = 7 * 24;
 export const SIGNING_SERVICE_URL =
-  config.maybeSignerServiceUrl ||
-  (config.isProd ? 'https://prototype-signer-service-polygon.pendulumchain.tech' : 'http://localhost:3000');
+  config.maybeSignerServiceUrl || (config.isProd ? '/api/production' : 'http://localhost:3000');
```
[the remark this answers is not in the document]
TorstenStueber: Actually I just kept the structure that was here already.

`maybeSignerServiceUrl` can only not be defined when running the code locally. In this case `isProd` will be false and then the assumption is that the signer service runs on `http://localhost:3000` (which might be wrong, but as I said, I just kept the code that was here already).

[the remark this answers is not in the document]
gianfra-t: I think it's fine, we never run locally and `isProd` is true.

README.md
```diff
@@ -37,10 +37,15 @@ matter what URL the browser requests.
 
 ## Env Variables
 
-- `VITE_SIGNING_SERVICE_URL`: Optional variable to point to a specific signing backend service URL. If undefined, it
+- `VITE_SIGNING_SERVICE_PATH`: Optional variable to point to a specific signing backend service URL. If undefined, it
```
ebma, line 40: I assume you decided to rename this variable as the values for `VITE_SIGNING_SERVICE_PATH` will not always be URLs but can also be just paths?

ebma (commented):
Looks good to me. Will you adjust the variables in netlify before merging this @TorstenStueber?

README.md
TorstenStueber: Yes, I forgot to describe this in the PR description. Yes, I defined new variables to have a smooth transition to the new system. I already defined them in Netlify.

TorstenStueber: The definition for production is `/api/production` and for staging and branch deploys is `/api/staging`.

ebma (approved):

PR #304, opened 2024-11-27
Hotfix: Update the Mykobo home domain

TorstenStueber:
This is a hotfix to use the new home domain of Mykobo.

See Slack for more info.


TorstenStueber:
I will merge this without a review as it is a trivial one line change and it can't break anything that is not already broken.

PR #308, opened 2024-11-27
Keep track of modal type and open states separately

### LINKED ISSUE #305: Bug: Output tokens shown in input tokens selection
When opening the popup to select an input token, then after clicking on one to select, the popup closes. During the moment of closing the listed tokens shown are not the input tokens but the output tokens (i.e., EURC and ARS).

This only happens during a split second but could indicate a more severe underlying problem so that we should fix it.

TorstenStueber:
Closes #305


gianfra-t (approved):
Looks good! I suppose it happened because `const definitions` was updated before the modal was hidden.

PR #322, opened 2024-12-12
Don't show Terms And Conditions modal

Sharqiewicz:


gianfra-t (approved):

src/hooks/useTermsAndConditions.ts
```diff
@@ -0,0 +1,38 @@
...
+  return {
+    termsChecked,
+    toggleTermsChecked,
+    termsAccepted,
+    acceptTerms,
+    rejectTerms,
+    setTermsAccepted: (accepted: boolean) => {
+      set(accepted ? 'accepted' : undefined);
+      setTermsAccepted(accepted);
+    },
```
ebma, line 36: I think these three are somewhat redundant and we don't need to have all of them. Either we only expose `acceptTerms` and `rejectTerms` and remove `setTermsAccepted` or we only keep `setTermsAccepted.

I would suggest we do
```suggestion
    setTermsAccepted: (accepted: boolean) => {
      if (accepted) {
       acceptTerms()
      } else {
       rejectTerms()
      }
    },
```

src/pages/swap/helpers/swapConfirm/index.ts
```diff
@@ -0,0 +1,92 @@
...
+    api,
+    to,
+    from,
+    selectedNetwork,
+    fromAmountString,
+    setIsInitiating,
+    setInitializeFailed,
+    handleOnSubmit,
+    setTermsAccepted,
+  }: SwapConfirmParams,
```
ebma, line 52: I think in this case it would make the code easier to read if we don't already extract the parameters here but rather do
```suggestion
  params: SwapConfirmParams,
```

and then below reference them as `params.address` etc. Only personal preference but I don't like the declaration of the function parameters spanning 16 lines.

src/pages/swap/helpers/swapConfirm/index.ts
```diff
@@ -0,0 +1,92 @@
...
+
+      handleOnSubmit({
+        inputTokenType: from,
+        outputTokenType: to,
+        amountInUnits: fromAmountString,
+        offrampAmount: validInputs.tokenOutAmountData.roundedDownQuotedAmountOut,
+        setInitializeFailed,
+      });
+    })
+    .catch((_error) => {
```
ebma, line 88: We should print the resulting error to console probably.

src/pages/swap/helpers/swapConfirm/index.ts
```diff
@@ -0,0 +1,92 @@
...
+
+import { ExecutionInput } from '../../../../hooks/offramp/useMainProcess';
+import { TokenOutData } from '../../../../hooks/nabla/useTokenAmountOut';
+import { Networks } from '../../../../contexts/network';
+
+import { calculateSwapAmountsWithMargin } from './calculateSwapAmountsWithMargin';
+import { performSwapInitialChecks } from './performSwapInitialChecks';
+import { validateSwapInputs } from './validateSwapInputs';
+
+interface SwapConfirmParams {
```
ebma, line 20: Let's at least order the params alphabetically or logically as they are a wild mix at the moment.

ebma (commented):
At the moment, the 'Confirm' button stays disabled as long as the T&C checkbox is not checked. However, I think it would be better if it was enabled (assuming the other checks for eg the input amount pass) and clicking on it will highlight the T&C checkbox in red color or similar. I as a user might miss the checkbox as it's not super prominent, which is okay, but I might wonder why I cannot click on the Confirm button in that case.

ebma (approved):
Thanks for addressing the changes, and I love the small scaling animation for the text. Let's merge.

Issue #292, opened 2024-11-21
Update flow to start off ramping from EVM networks


prayagd:
### User Story
As a user, I want to off my USDT / USDC from the following networks:
- [ ]  Ethereum
- [ ]  BSC
- [ ]  Arbitrum
- [ ]  Base
- [ ]  Avalanche 

## Assets
| Network  | USDC  |   USDT |
| -------- | ------- | ------- |
| Ethereum  | 0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48 | 0xdAC17F958D2ee523a2206206994597C13D831ec7|
| BSC  | 0x8AC76a51cc950d9822D68b83fE1Ad97B32Cd580d | 0x55d398326f99059fF775485246999027B3197955|
| Arbitrum  | 0xaf88d065e77c8cC2239327C5EDb3A432268e5831 | 0xFd086bC7CD5C481DCC9C85ebE478A1C0b69FCbb9|
| Base  |  0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913| 0xfde4C96c8593536E31F229EA8f37b2ADa2699bb2 |
| Avalance  | 0xB97EF9Ef8734C71904D8002F8b6Bc66Dd9c48a6E | 0x9702230A8Ea53601f5cD2dc00fDBc13d4dF4A8c7 |

### Acceptance criteria
- Update the Vortex flow such that it starts the off-ramp from the above listed networks
- Subsidy feature takes into account liquidity on the respective network
- Network dropdown allows user to choose network to off-ramp from
- Network selection creates a prompt for the user to change network in the wallet
- Asset list is specific to the network chosen
- Progress screen shows accurate asset and network
- Deep links for 'To' and 'From' assets should also include the network selection
- Compare fee should consider accurate asset / network
- Changing network between EVM chains should not disconnect wallet considering wallet is connected

## Note
* Assumption: This ticket is for both frontend and backend changes, assuming configuration changes to the current setup

prayagd:
@pendulum-chain/devs What all things to be considered when adding support to the above listed networks and assets?

ebma:
I would again just check for routes of all the listed tokens in the [squidrouter app/UI](https://app.squidrouter.com/). 

[...]

PR #323, opened 2024-12-12
Start off-ramping from evm networks

gianfra-t:
Note: Review and merge after [this refactor](https://github.com/pendulum-chain/vortex/pull/324) is merged into staging.

Issue: #292.

### Main changes
- Adds logic to support off-ramp from other `evm` chains.
- Adds relevant config for each new network.
- Adds relevant icons.
- General bug fixes and improvements.

### Extra (optional) changes
- Includes a [refactor](https://github.com/pendulum-chain/vortex/pull/323/commits/26ed1b51b605270e0bacfeb133054fbda9f63182) into `zustand` of the hooks for `sep24`. 

### Tests
Off-ramps tested to `ARS`:

- [x] Arbitrum
- [x] Base
- [ ] Avalanche
- [ ] Ethereum
- [x] BSC


src/hooks/useInputTokenBalance.ts
```diff
@@ -8,21 +8,22 @@ import { multiplyByPowerOfTen } from '../helpers/contracts';
 import { nativeToDecimal, USDC_DECIMALS } from '../helpers/parseNumbers';
 import { usePolkadotWalletState } from '../contexts/polkadotWallet';
 import { useAssetHubNode } from '../contexts/polkadotNode';
+import { useVortexAccount } from './useVortexAccount';
 
+// TODO maybe improve: if the user switches the network in the selector and REJECTS the switch in wallet, then balance will be 0.
```
gianfra-t, line 13: Metamask needs this explicit acceptance, while Talisman for instance does not.

Since `useReadContract` uses the chain that is connected in the wallet, it will read form an incorrect chain.
If this is a UI requirement, we could fetch with the correct chain using the `selectedNetwork` parameter.

src/services/offrampingFlow.ts
```diff
@@ -180,11 +187,13 @@ export async function constructInitialState({
     pendulumNode,
   );
 
-  const inputTokenDecimals = getInputTokenDetailsOrDefault(network, inputTokenType).decimals;
+  const { decimals: inputTokenDecimals, pendulumDecimals } = getInputTokenDetailsOrDefault(network, inputTokenType);
   const outputTokenDecimals = OUTPUT_TOKEN_CONFIG[outputTokenType].decimals;
 
   const inputAmountBig = Big(amountIn);
   const inputAmountRaw = multiplyByPowerOfTen(inputAmountBig, inputTokenDecimals || 0).toFixed();
+  const pendulumAmountRaw = multiplyByPowerOfTen(inputAmountBig, pendulumDecimals || 0).toFixed();
```
gianfra-t, line 195: We were previously using `inputAmountRaw` for all operations on Pendulum after writing the off-ramping state. Since the decimals on the starting chain and `axlUSDC` on Pendulum where the same (`6`), there was no issue. 

BSC uses 18 decimals so we need to account for this, or potentially for a different token combination.

src/stores/sep24Store.ts
```diff
@@ -0,0 +1,71 @@
+import { create } from 'zustand';
```
Sharqiewicz, line 1: As we separated the whole sep24 state into this store I think we can refactor the names to reduce cognitive load.
What about:

```
import { create } from 'zustand';
import { IAnchorSessionParams, ISep24Intermediate } from '../services/anchor';
import { ExecutionInput } from '../hooks/offramp/useMainProcess';

export type ExtendedExecutionInput = ExecutionInput & { stellarEphemeralSecret: string };

export interface Sep24State {
  sessionParams: IAnchorSessionParams | undefined;
  initialResponse: ISep24Intermediate | undefined;
  executionInput: ExtendedExecutionInput | undefined;
  urlInterval: number | undefined;
}

export interface Sep24Actions {
  setSessionParams: (params: IAnchorSessionParams | undefined) => void;
  setInitialResponse: (response: ISep24Intermediate | undefined) => void;
  setExecutionInput: (input: ExtendedExecutionInput | undefined) => void;
  setUrlInterval: (interval: number | undefined) => void;
  reset: () => void;
  cleanup: () => void;
}

interface Sep24Store extends Sep24State {
  actions: Sep24Actions;
}

export const useSep24Store = create<Sep24Store>()((set, get) => ({
  sessionParams: undefined,
  initialResponse: undefined,
  executionInput: undefined,
  urlInterval: undefined,

  actions: {
    setSessionParams: (params) => set({ sessionParams: params }),
    setInitialResponse: (response) => set({ initialResponse: response }),
    setExecutionInput: (input) => set({ executionInput: input }),
    setUrlInterval: (interval) => set({ pollInterval: interval }),

    reset: () =>
      set({
        sessionParams: undefined,
        initialResponse: undefined,
        executionInput: undefined,
        urlInterval: undefined,
      }),

    cleanup: () => {
      const { urlInterval } = get();
      const actions = get().actions;

      if (urlInterval !== undefined) {
        clearInterval(pollInterval);
        actions.setUrlInterval(undefined);
        actions.setInitialResponse(undefined);
        actions.setExecutionInput(undefined);
        actions.setSessionParams(undefined);
      }
    },
  },
}));

export const useActions = () => useSep24Store((state) => state.actions);
export const useInitialResponse = () => useSep24Store((state) => state.initialResponse);
export const useUrlInterval = () => useSep24Store((state) => state.pollInterval);
export const useSessionParams = () => useSep24Store((state) => state.sessionParams);
export const useExecutionInput = () => useSep24Store((state) => state.executionInput);

```

src/hooks/offramp/useOfframpAdvancement.ts
```diff
@@ -9,49 +9,61 @@ import { usePendulumNode } from '../../contexts/polkadotNode';
...
-  }, [offrampState, trackEvent, updateOfframpHookStateFromState, wagmiConfig]);
+  }, [
+    offrampState,
+    trackEvent,
+    updateOfframpHookStateFromState,
+    wagmiConfig.state.status,
+    selectedNetwork,
+    pendulumNode,
+    assetHubNode,
+    walletAccount?.address,
```
Sharqiewicz, line 67: missing dependencies: 'setOfframpSigningPhase', 'wagmiConfig', and 'walletAccount'

Sharqiewicz (changes requested):

src/contexts/network.tsx
```diff
@@ -1,20 +1,61 @@
...
   AssetHub = 'AssetHub',
   Polygon = 'Polygon',
+  Ethereum = 'Ethereum',
+  BSC = 'BSC',
+  Arbitrum = 'Arbitrum',
+  Base = 'Base',
+  Avalanche = 'Avalanche',
+}
+
+export function isNetworkEVM(network: Networks): boolean {
```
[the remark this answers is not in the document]
gianfra-t: I have no strong opinion of leaving this logic here, but I guess it could go to a helper. I did not find a better way to determine if the network is `evm`. The alternative would be to check if it's `substrate`, but then we are at the same issue if those grow too much.

src/contexts/network.tsx
Sharqiewicz, line 1: In a separate PR we can refactor it into a zustand store

Sharqiewicz:
@pendulum-chain/devs I have refactored the code, please review. 

Changes:
- [x] Divide SEP logic into SEP10 `anchor/sep10` and SEP24 `anchor/sep24`
- [x] Improve Navbar responsiveness
- [x] Fix fee comparison quotes 
- [x] Add `selectedChain` to `useEvmBalance`
- [x] Adjust Fee Comparison tooltip
- [x] Simplified `Networks` logic

src/wagmiConfig.ts
```diff
@@ -1,10 +1,11 @@
-import { polygon } from '@reown/appkit/networks';
+import { polygon, bsc, arbitrum, base, avalanche, mainnet } from '@reown/appkit/networks';
```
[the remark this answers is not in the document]
gianfra-t: Could be. But this are just constants right? Or you mean, to have a centralized place to handle network information in case we deviate from appkit's definitions.

src/services/quotes/index.ts
```diff
@@ -28,11 +28,20 @@ async function getQuoteFromService(
   amount: Big,
   network: Networks,
 ): Promise<Big> {
-  if (network !== polygon.name) {
-    throw new Error(`Network ${network} is not supported`);
+  let compatibleNetwork = network;
+  if (!isNetworkEVM(network)) {
```
gianfra-t, line 32: I'm not so sure we want to show the fee comparison of Polygon when on AssetHub. If we throw here, then it will display `N/A` which in my opinion is better. Especially since AssetHub has some higher costs.

gianfra-t:
Thanks for the improvements @Sharqiewicz . Seems cleaner! I just tested an off-ramp from BSC and on friday, I tested from AssetHub. I think it is good to go.

src/services/quotes/index.ts
Sharqiewicz: For AssetHub we want to show Polygon prices.
```
<div
    className="tooltip tooltip-primary before:whitespace-pre-wrap before:content-[attr(data-tip)]"
    data-tip="Quotes are for Polygon, as the providers don't support Asset hub."
            >
              (Polygon)
</div>
```

Sharqiewicz (approved):

PR #324, opened 2024-12-12
Refactor useMainProcess

Sharqiewicz:
## PR 🦈 🟢
---

### Overview

Introduce Zustand for feature-state management.

State management in the app:

Global State = `React Context + Zustand`
Feature State = `Zustand`
Server State = `Tanstack/react-query`

### Purpose

Currently, the app suffers from disorganized, logic and state management are entangled in a single large file, along with event listeners for multiple features. This makes introducing changes or even testing the code extremely challenging.

### Changes

Introduce `offrampStore` for offramp feature related state (types are defined within `/types/offramp.ts`)
Introduce `useOfframpAdvancement` custom hook for advancement of offramp phase
Introduce `useOfframpEvents` custom hook for offramp events related logic
Refactor `useMainProcess` custom hook


src/hooks/offramp/useOfframpEvents.ts
```diff
@@ -0,0 +1,45 @@
...
+import { OfframpingState } from '../../services/offrampingFlow';
+import { OFFRAMPING_PHASE_SECONDS } from '../../pages/progress';
+
+export const useOfframpEvents = () => {
+  const { trackEvent, resetUniqueEvents } = useEventsContext();
+  const { selectedNetwork } = useNetwork();
+  const [_, setEvents] = useState<GenericEvent[]>([]);
+
+  // @todo: why do we need this?
+  const addEvent = (message: string, status: EventStatus) => {
```
gianfra-t, line 20: Not at all! this comes from way way before where it used to display the logs on the UI. We've been dragging parts of it for a while now apparently.

src/hooks/offramp/useOfframpEvents.ts
```diff
@@ -0,0 +1,45 @@
...
+    setEvents((prevEvents) => [...prevEvents, { value: message, status }]);
+  };
+
+  const trackOfframpingEvent = useCallback(
```
gianfra-t, line 24: Do you think we could move this to `useEventsContext`? Especially since we remove the `addEvent` thingy. This `useOfframpEvents` is a bit intermediary.
If it's too much trouble then no need.

gianfra-t (commented):
I really like the new library and refactor 💪. I think we still could get rid of so many hook states especially those regarding sep24 which are all over the place, like `setAnchorSessionParams` which for example has `useSEP24Cleanup` in a separate hook. But that for later otherwise the PR will grow a lot.

src/hooks/offramp/useOfframpEvents.ts
Sharqiewicz: @gianfra-t Can I remove it?

gianfra-t: Yes sure! I hope is not so much work

src/hooks/offramp/useOfframpEvents.ts
Sharqiewicz: I think we should keep the logic in a custom hook and in the Events Context keep only state-related code. The Event Context file is actually quite big right now

Sharqiewicz: @gianfra-t I think we should keep the logic in a custom hook and in the Events Context keep only state-related code. The Event Context file is actually quite big right now

Sharqiewicz:
@gianfra-t Do we merge it to staging?

Sharqiewicz:
@gianfra-t Could you approve 🙏 🟢

PR #327, opened 2024-12-13
Fix wrong use of evm account address for assethub offramp

### LINKED ISSUE #326: Fix wrong use of evm account address for AssetHub offramp
## Context 

There currently are some flaws in the code that make the offramping flow assume and use the address of a connected EVM account instead of the Substrate account. We should address these flaws. One symptom of this is that the 'Confirm' button does not do anything if no EVM account is connected.

## TODO

Change the code so that connecting a Substrate account is enough to start the flow.

ebma:
- [ ] Adjust SIWE for Substrate accounts

Closes #326.

### DIFF
--- a/src/hooks/useVortexAccount.ts
+++ b/src/hooks/useVortexAccount.ts
@@ -1,0 +1,56 @@
+import { Networks, useNetwork } from '../contexts/network';
+import { useMemo } from 'preact/compat';
+import { usePolkadotWalletState } from '../contexts/polkadotWallet';
+import { useAccount } from 'wagmi';
+
+// For the AssetHub network, we use a chain ID of -1. This is not a valid chain ID
+// but we just use it to differentiate between the EVM and Polkadot accounts.
+const AssetHubChainId = -1;
+
+// A helper hook to provide an abstraction over the account used.
+// The account could be an EVM account or a Polkadot account.
+export const useVortexAccount = () => {
+  const { selectedNetwork } = useNetwork();
+
+  const { walletAccount: polkadotWalletAccount } = usePolkadotWalletState();
+  const { isDisconnected: isEvmAccountDisconnected, chainId: evmChainId, address: evmAccountAddress } = useAccount();
+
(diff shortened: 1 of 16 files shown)


gianfra-t (approved):
Looks good 👍.

src/hooks/useVortexAccount.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  const chainId = useMemo(() => {
+    if (selectedNetwork === Networks.AssetHub) {
+      return AssetHubChainId;
+    } else {
+      return evmChainId;
+    }
+  }, [selectedNetwork, evmChainId]);
+
+  const type = useMemo(() => {
```
Sharqiewicz, line 42: I think we don't have to useMemo for variables here, there is no heavy computation that would require it, wdyt?

src/pages/progress/index.tsx
```diff
@@ -180,7 +183,12 @@ export const ProgressPage: FC<ProgressPageProps> = ({ offrampingState }) => {
     <BaseLayout
       main={
         <main>
-          <ProgressContent currentPhase={currentPhase} currentPhaseIndex={currentPhaseIndex} message={message} />
+          <ProgressContent
+            currentPhase={currentPhase}
+            currentPhaseIndex={currentPhaseIndex}
+            message={message}
+            network={selectedNetwork}
```
Sharqiewicz, line 190: we don't have to pass `selectedNetwork` to this component. We can just use `useNetwork` inside of ProgressContent to reduce props drilling

src/hooks/useVortexAccount.ts
ebma: That's true but I decided to use it anyway to mitigate any potential future issues in other components if we decide to put the return values of this hook into the dependencies of another hook. We had those issues in the past and it was quite hard to debug IMO.

src/pages/progress/index.tsx
ebma: Changed.

PR #331, opened 2024-12-16
Add SIWE signing support for Substrate wallets.

gianfra-t:


gianfra-t:
@pendulum-chain/devs Let's first agree if we like the solution on a high level (as it involves security) and later I can clean/refactor more. I'll leave it as a draft for that reason, but the I tested the solution already.

signer-service/src/api/middlewares/auth.js
```diff
@@ -0,0 +1,77 @@
...
+        }
+      } catch (e) {
+        continue;
+      }
+    }
+
+    // Client declared usage of memo, but it could not be derived from provided signatures.
+    if (Boolean(req.body.memo) && !resultMemo) {
+      return res.status(401).json({
+        error: 'Missing signature or nonce',
```
ebma, line 51: Maybe we should send a better error. The error message doesn't seem accurate to me based on the context.

signer-service/src/api/services/siwe.service.js
```diff
@@ -30,29 +34,34 @@ const verifySiweMessage = async (nonce, signature, initialSiweMessage) => {
...
+    const publicClient = createPublicClient({
+      chain: polygon,
+      transport: http(),
+    });
+    valid = await publicClient.verifyMessage({
+      address: address,
+      message: siweMessage.toMessage(), // Validation must be done on the message as string
+      signature,
+    });
+  } else if (address.startsWith('5')) {
```
ebma, line 56: Might not be guaranteed that every valid address starts with a 5. I think there could be some edgy edge cases where it doesn't. I could be wrong though. Let's rather try to call some decode function on the keyring and if there is some result, we assume it's a substrate address?

ebma:
I like it overall, nice job @gianfra-t  👍 would still need to look at the details in more depth but I think the idea is good and I'd say you can go ahead with refactoring it to a state where it's reado for merg.

signer-service/src/api/middlewares/auth.js
gianfra-t: I agree. Changed to `Missing or invalid authentication token` which is a bit better. Otherwise we would need to add extra conditions to differentiate.

signer-service/src/api/services/siwe.service.js
gianfra-t: I think the "generic substrate" format always [start with](https://wiki.polkadot.network/docs/learn-account-advanced) `5`. 
We can add extra verification but checking the checksum and ensure is at least valid for each option, but the success still depends on the UI sending the correct address that matches with one of the auth credentials, so I don't know if it adds much value, wdyt? It would mostly be for us to get better error if we send an invalid address.

gianfra-t:
I don't really have more ideas about [this](https://github.com/pendulum-chain/vortex/pull/331/files#diff-38a8c330f117845dd8712f6df1890b2c7e80596dad1b45dd0747249086bd4b5bR1) new class added. It's not so bad, I just I don't like to have it duplicated. But it is the same with much of the code, like stellar operations.

signer-service/src/api/services/siwe.service.js
ebma: Right. Are we always sending the generic substrate address though? I guess you are right and we are. Would be important to keep doing it like that but I'm fine with keeping it like this then.

src/services/anchor/index.ts
```diff
@@ -150,7 +150,10 @@ export const sep10 = async (
   const { signingKey, webAuthEndpoint } = tomlValues;
 
   if (!exists(signingKey) || !exists(webAuthEndpoint)) {
-    throw new Error('Missing values in TOML file');
+    throw new Error('sep10: Missing values in TOML file');
+  }
+  if (!address) {
+    throw new Error('sep10: Address must be defined at this stage');
   }
```
ebma, line 157: Maybe we should just change the function signature to not allow passing an undefined address 😅

src/services/signingService.tsx
```diff
@@ -22,6 +22,7 @@ export interface SignerServiceSep10Request {
   challengeXDR: string;
   outToken: OutputTokenType;
   clientPublicKey: string;
+  address: string;
   memo?: boolean;
```
ebma, line 26: Let's rename this parameter to something else. Just calling it `memo` makes it confusing when processing it on the signer-service. Let's call it something like

```suggestion
  usesMemo?: boolean;
```

src/hooks/useSignChallenge.ts
```diff
@@ -56,6 +54,22 @@ export function useSiweSignature(address?: string) {
...
+      const { signature: substrateSignature } = await (polkadotWalletAccount!.signer as Signer).signRaw!({
+        type: 'payload',
+        data: siweMessage,
+        address,
+      });
+      signature = substrateSignature;
+    }
+
+    return signature;
+  }
```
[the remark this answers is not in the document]
gianfra-t: Okey, makes sense. It will be useful to clean up a bit the Sign component.

src/services/anchor/index.ts
gianfra-t: Moved to the caller of `sep10` now, but we can't really get rid of the condition there unless we want to touch address in `useVortexAccount`. I think it's fine though.

gianfra-t:
Addressed some comments and tested again. Not sure why `yarn.lock` downgraded `types/node`, seems to be related to the removal of `siwe` and it's dependencies.

PR #332, opened 2024-12-17
Fix offramp process getting stuck after entering details on anchor page

ebma:
Sometimes it happens that the offramping flow is stuck _after_ returning from the anchor page but before completing the 'prepareTransactions' phase. We suspect that this mostly happens when the Stellar horizon servers are busy and the `horizon.loadAccount()` function never times out. 
This PR wraps the `loadAccount()` function with a timeout and retry logic. For now, it's enough to only have that logic for the `loadAccount()` function as the implementation for transaction submissions defines a default timeout of 60 seconds and we don't need to worry about that.


gianfra-t (approved):
Changes look good. Even if it doesn't solve the particular issue of waiting forever, I think we should include them.

src/services/phases/stellar/index.tsx
```diff
@@ -14,14 +14,51 @@ import {
...
 import { fetchSigningServiceAccountId } from '../../signingService';
 import { OfframpingState } from '../../offrampingFlow';
 import { SepResult } from '../../anchor';
 
 const horizonServer = new Horizon.Server(HORIZON_URL);
 const NETWORK_PASSPHRASE = Networks.PUBLIC;
 
+// This is a helper function to add a timeout to the loadAccount function and retry it.
+// The Horizon.Server class does not allow defining a custom timeout for network queries.
+async function loadAccountWithRetry(
```
Sharqiewicz, line 32: What do you think about using React Query for this?

```
import { useQuery } from '@tanstack/react-query';

const fetchStellarAccount = async (accountId: string): Promise<Horizon.AccountResponse | null> => {
  try {
    return await horizonServer.loadAccount(accountId);
  } catch (err: any) {
    if (err?.toString().includes('NotFoundError')) {
      return null;
    }
    throw err;
  }
};

export function useStellarAccount(accountId: string | undefined) {
  return useQuery({
    queryKey: ['stellar-account', accountId],
    queryFn: () => fetchStellarAccount(accountId!),
    enabled: !!accountId,
    retry: 3,
    retryDelay: 3000,
    staleTime: 15000, 
  });
}

```

Sharqiewicz (changes requested):

src/services/phases/stellar/index.tsx
ebma: It would be nice to use it for caching as well but there are two issues. 1) it's a hook so we would need to turn the other functions into hooks as well and more importantly 2) your proposed refactoring is not doing what we need to do, that is time out the `loadAccount()` function. The `staleTime` is an indicator for when the stellar account should be refetched. It does not make the `horizonServer.loadAccount()` function time out after 15 seconds.  Our assumption is that the bug we have right now is that `loadAccount()` never times out which we are trying to do manually with the `Promise.race()`. I checked the documentation of [useQuery](https://tanstack.com/query/v4/docs/framework/react/reference/useQuery) and I didn't find a parameter that would do what we need. So I would leave the code as is.

Sharqiewicz: @ebma ✅

PR #337, opened 2024-12-20
Fix restoring offramp after a refresh

gianfra-t:
Solves issue: Stuck after anchor completion.

### DIFF
--- a/src/hooks/offramp/useMainProcess.ts
+++ b/src/hooks/offramp/useMainProcess.ts
@@ -157,7 +157,7 @@ export const useMainProcess = () => {
   }, [updateHookStateFromState, offrampingState]);
 
   useEffect(() => {
-    if (wagmiConfig.state.status !== 'connected') return;
+    //if (wagmiConfig.state.status !== 'connected') return;
 
     (async () => {
       if (!pendulumNode || !assetHubNode) {
--- a/src/hooks/offramp/useMainProcess.ts
+++ b/src/hooks/offramp/useMainProcess.ts
@@ -129,9 +129,9 @@ export const useMainProcess = () => {
     setSigningPhase,
   ]);
 
-  useEffect(() => {
-    setOnSelectedNetworkChange(resetOfframpingState);
-  }, [setOnSelectedNetworkChange, resetOfframpingState]);
+  // useEffect(() => {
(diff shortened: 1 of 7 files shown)


src/hooks/offramp/useMainProcess.ts
```diff
@@ -129,9 +129,9 @@ export const useMainProcess = () => {
     setSigningPhase,
   ]);
 
-  useEffect(() => {
-    setOnSelectedNetworkChange(resetOfframpingState);
-  }, [setOnSelectedNetworkChange, resetOfframpingState]);
+  // useEffect(() => {
+  //   setOnSelectedNetworkChange(resetOfframpingState);
```
gianfra-t, line 133: I do not know why the act of passing this function triggers it, even though there is no network change.

src/hooks/offramp/useMainProcess.ts
```diff
@@ -129,10 +129,6 @@ export const useMainProcess = () => {
     setSigningPhase,
   ]);
 
-  useEffect(() => {
```
gianfra-t, line 132: As per the previous comment, just setting `setOnSelectedNetworkChange(resetOfframpingState);` causes it to trigger on mount, which removes a potential state that must be picked up for continuation.

Since the network selector is disabled during the off-ramp, I think we can remove this now.

If we still want to add it back in the future, a potential solution exists in [this](https://github.com/pendulum-chain/vortex/pull/323/files#diff-bb2814ac946cf36194efa0ba6b097de7881a6780c27ef1412100165ece04ce09R89) WIP PR, possible after merging [this](https://github.com/pendulum-chain/vortex/pull/324/files#diff-d2caeb4570ef1c2b7edf7b74ad89218b6f82a0a640967070b48a343766e18c66).

ebma (approved):
Nice catch, you finally found the issue 🙏 Let's merge.

src/hooks/offramp/useMainProcess.ts
```diff
@@ -183,7 +180,16 @@ export const useMainProcess = () => {
     // but we intentionally exclude them from the dependency array to prevent unnecessary re-renders.
```
Sharqiewicz, line 180: nit: We can remove this comment if nothing is missing from the dependency array

Sharqiewicz (approved):

Issue #346, opened 2024-12-30
Handle failed XCM transfers when off-ramping from AssetHub


gianfra-t:
### Issue description
A user attempted an off-ramp starting from AssetHub (XCM transaction [here](https://assethub-polkadot.subscan.io/extrinsic/7862371-2)) yet found that the application was stuck after signing.

This is most likely because we are not handling the case when the [transaction](https://github.com/pendulum-chain/vortex/blob/polygon-prototype-staging/src/services/phases/polkadot/assethub.ts#L56) is successful, yet the XCM execution itself is not. In this case, the application will wait indefinitely for the funds to arrive at Pendulum.

Also, this failed XCM should potentially have been prevented in the first place. The current theory is that it failed initially due to Existential Deposit issues, since the USDC amount attempted to be transferred was very close to the full amount the user had. A second [xcm](https://assethub-polkadot.subscan.io/extrinsic/7884348-2) with a lower amount worked well.

### Potential fixes
We could query the status of the XCM [here](https://github.com/pendulum-chain/vortex/blob/polygon-prototype-staging/src/services/phases/polkadot/assethub.ts#L61C11-L61C24), by listening to the events of that block and checking if the XCM failed.

Aside from that, we need to verify if the existential deposit was the root cause of the first failed off-ramp, and if so, consider it [here](https://github.com/pendulum-chain/vortex/blob/polygon-prototype-staging/src/hooks/useInputTokenBalance.ts#L30).

gianfra-t:
@pendulum-chain/product although there is no risk of loosing funds if this happens again, we should consider fixing this to make Assethub off-ramps more robust. It is very likely users will try to off-ramp all it's USDC and this happens again.

prayagd:
@gianfra-t is there a ED for USDC too on Assethub? if that's the case then a quick solution would be to show the balance minus ED on the vortex UI

gianfra-t:
Yup, that's a good simple solution. I think there is an ED for any token. What remains to be seen is if it works if all balance is transferred, we may have to use the `transfer_all` in this case to allow for people to do that.

PR #349, opened 2025-01-02
Account for minimum balance when off-ramping from AssetHub.

gianfra-t:
NOTE: Wait for the merge of 292 before merging.

Issue #346.

As described in the ticket, this happens because in AssetHub we have a `minBalance` for each asset. When the `transfer` operation is attempted by the `xcm_executor`, which would leave the balance bellow this specified minimum, the `xcm` transfer will fail.

### Considerations
We are rounding the number to 2 decimals for ease of visualization. This means the user may see a slightly smaller balance as compared to what can be seen on chain, on the order of the `minBalance` property.
 
Example USDC

Balance: `1530000` (1.535 USDC).
Minimum Balance: `10000` (0.01 USDC)

Technically, the user could transfer: `1535000 - 10000 = 1525000` (1.525 USDC) but due to our rounding, the app will display `1.52`.


ebma:
@gianfra-t you didn't add any reviewers, is this ready?

gianfra-t:
I was waiting for the evm PR since it was branches from it.

It is ready, but the events part is more like a  suggestion only.  As it may have downsides.

ebma:
If there are any considerations that you are worried about, feel free to leave a note in the description so we can take it into account when reviewing.

src/services/phases/polkadot/assethub.ts
```diff
@@ -53,8 +56,16 @@ export async function executeAssetHubXCM(state: OfframpingState, context: Execut
...
       const tx = createAssethubAssetTransfer(assetHubNode.api, pendulumEphemeralAddress, inputAmount.raw);
       context.setOfframpSigningPhase('started');
+
+      const eventListener = EventListener.getEventListener(assetHubNode.api);
+      const xcmSentEventPromise = eventListener.waitForXcmSentEvent(walletAccount.address, maxWaitingTimeMs);
+
       const { hash } = await tx.signAndSend(walletAccount.address, { signer: walletAccount.signer as Signer });
       setOfframpSigningPhase?.('finished');
+
+      await xcmSentEventPromise;
```
gianfra-t, line 66: This is the core idea. We would now wait for an event that confirms that the XCM was sent.

Now, the issue is that we should also probably listen to the corresponding failed event to be complete, and this may be overkill for the application. @pendulum-chain/devs let me know what you think.

ebma (approved):
All looks good to me. Also the considerations with rounding down the balance makes sense. Users can't withdraw fractions of a cent anyway, so no need to allow more granularity than 2 decimals. Great job @gianfra-t, I think we can merge this.

PR #350, opened 2025-01-03
Improve UI on log-in modal.

gianfra-t:
Closes: #351 (and #362)

A re-use of the signing box needs to show when the user attempts to login. This box should not show a progress bar on this stage.

### Changes
- Modify the `SigningBox` component to allow for this new "phase". Amount of signatures to display are adjusted depending on the phase. Also an animation would rapidly load the progress bar to 100% after finish all signatures, to "dampen" the sudden dismissal of the box.
- Modify `useSiweSignature` hook to work without the modal.
- Improve error display and recovery when user rejects signing/network change requests. Errors are shown using a toast now.
- We now clear the `sep24` variables upon network change. This resets the state and allow the user to click confirm again.


src/components/SignIn/index.tsx
```diff
@@ -1,4 +1,5 @@
-import { FC } from 'preact/compat';
+import { FC, useState } from 'preact/compat';
+import { useEffect } from 'react';
```
Sharqiewicz, line 2: nit: don't import from react

src/pages/swap/index.tsx
```diff
@@ -368,7 +374,7 @@ export const SwapPage = () => {
       from,
       selectedNetwork,
       fromAmountString,
-      requiresSquidRouter: selectedNetwork === Networks.Polygon,
+      requiresSquidRouter: isNetworkEVM(selectedNetwork),
```
gianfra-t, line 377: Not related to this change, but we need this after support of other EVM networks.

ebma (commented):
I tested it. Works overall 👍  I pushed a commit to add the progress bar and footer also for the login. IMO it looks better with both indicators and there is no need to hide them. I put the progress of the login to 15% by default so that it's clearer that the progress bar is indicating progress which is not clear with an empty bar. 

I think we should spend more time on re-thinking the retry button. It's a nice feature overall _but_ I wouldn't merge it the way it is. The concern I have is that when I reject a signature and click on the retry button, I would expect that the app continues where it stopped, ie. it would ask me to sign again. This is not the case at the moment and I need to click on the 'Confirm' button again. It doesn't seem like a big deal but would only lead to confusion if we don't do it properly. I propose we remove the retry button from this PR and discuss with the product team.

src/services/phases/polkadot/assethub.ts
```diff
@@ -40,8 +40,6 @@ export async function executeAssetHubXCM(state: OfframpingState, context: Execut
     throw new Error('AssetHub node not available');
   }
 
-  setOfframpSigningPhase?.('started');
```
gianfra-t, line 43: I found that this makes on some occasions that the signing box gets "stuck" after signing. 
Since on reload, the component will execute this call yet it will not call again the sign process if `assetHubXcmTransactionHash !== undefined`.

@pendulum-chain/devs do we set the sign stage here for some reason?

src/hooks/offramp/useSubmitOfframp.ts
```diff
@@ -124,10 +126,10 @@ export const useSubmitOfframp = () => {
             setOfframpInitiating(false);
           }
         } catch (error) {
-          console.error('Error initializing the offramping process', error);
+          console.error('Error initializing the offramping process', (error as Error).message);
           // Display error message, differentiating between user rejection and other errors
-          if ((error as Error).message.includes('User rejected the request')) {
-            setInitializeFailed('Please switch to the correct network and try again.');
+          if ((error as Error).message.includes('User rejected')) {
+            showToast(ToastMessage.ERROR, 'You must sign the login request to offramp Argentine Peso');
```
gianfra-t, line 132: This is really not very reliable, I believe each wallet would throw different error messages.
But I don't know a way to capture the error besides string match.

gianfra-t:
@ebma I removed the retry button. Also we discussed we wanted to add a toast when detecting user rejection, and just reset the app. So that type of error is not shown on the UI now.

src/pages/swap/index.tsx
```diff
@@ -101,6 +98,15 @@ export const SwapPage = () => {
     );
   }, []);
 
+  const clearInitializeErrors = useCallback(() => {
+    if (!initializeFailedMessage) return;
+    if (initializeFailedMessage.includes('Hang tight')) {
```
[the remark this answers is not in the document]
gianfra-t: Totally, it was not very clean, although it was just to showcase the idea. Now this callback is removed as there will be no retry button.

src/hooks/useSignChallenge.ts
```diff
@@ -95,23 +96,27 @@ export function useSiweSignature() {
...
       const errorMessage = error instanceof Error ? error.message : String(error);
-      signPromise.reject(new Error('Signing failed: ' + errorMessage));
+      setOfframpSigningPhase?.(undefined);
+
+      if (
+        (error as Error).message.includes('User rejected the request') ||
+        (error as Error).message.includes('Signing failed: Cancelled')
+      ) {
+        // First case Assethub, second case EVM
+        return signPromise.reject(new Error('Signing failed: User rejected sign request'));
```
gianfra-t, line 109: This is really not very reliable, I believe each wallet would throw different error messages.
But I don't know a way to capture the error besides string match.

src/services/phases/polkadot/assethub.ts
ebma: I guess we don't really need it if we also set it down below. Let's remove it if not necessary.

src/hooks/useSignChallenge.ts
ebma: I think we should move the comment in l. 108 above the `if` statement. If we want to make it perfect we could return error objects with unique identifiers per error instead of relying on the message itself. But maybe that's too much at the moment.

src/hooks/offramp/useSubmitOfframp.ts
ebma, line 132: Nit: I guess we should stick to the 'Sell crypto' lingo and not mention offramps again. 
```suggestion
            showToast(ToastMessage.ERROR, 'You must sign the login request to be able to sell Argentine Peso');
```

ebma (approved):
Tested it again. Seems to work as expected 🙏  Let's change the wording and finally get this out to staging.

PR #352, opened 2025-01-06
[READY] Migrate signer-service to Typescript

Sharqiewicz:
## PR 🦈🟢
---

- [x] Migrate all files in `signer-service` to Typescript
- [x] Configure SWC ( TS Compiler )
- [x] Configure commands:
- `yarn build` Compiles TypeScript code into JavaScript using the fast SWC compiler
- `yarn start` Runs the compiled application
- `yarn build:start` Compiles and then immediately runs the application (When you want to test production)
- `yarn dev` Runs the application in development mode with automatic reloading on code changes

## How to review
There are only two new files `.swcrc` and `tsconfig.json` they are short and consistent.
The challenge with the review is going through every migrated file. I tried to keep the logic the same, but sometimes there were a lot of TypeScript errors, so I had to make changes to the files.


Sharqiewicz:
@pendulum-chain/devs Ready for review, that's quite a few changes

signer-service/src/api/controllers/moonbeam.controller.ts
```diff
@@ -0,0 +1,101 @@
...
+  });
+
+  return { walletClient, publicClient };
+};
+
+export const executeXcmController = async (req: Request, res: Response): Promise<void> => {
+  const { id, payload } = req.body;
+
+  try {
+    if (!MOONBEAM_EXECUTOR_PRIVATE_KEY) {
```
gianfra-t, line 40: We already check for this variables (and any required one) on `index.ts`, using `validateRequiredEnvVars`, when the service starts.

Because we should never reach this function without the variable because the process will break in the middle.

signer-service/src/api/controllers/quote.controller.ts
```diff
@@ -0,0 +1,125 @@
...
+  amount: string;
+  network?: string;
+}
+
+type QuoteHandler = (
+  fromCrypto: CryptoCurrency,
+  toFiat: FiatCurrency,
+  amount: string,
+  network?: string,
+) => Promise<unknown>;
```
gianfra-t, line 29: We could provide `type AnyQuote = AlchemyPayQuote | MoonpayQuote | QuoteResult;` wdyt?

signer-service/src/api/controllers/quote.controller.ts
```diff
@@ -0,0 +1,125 @@
...
+  amount: string,
+  network?: string,
+) => {
+  return await providerHandlers[provider](fromCrypto, toFiat, amount, network);
+};
+
+export const getQuoteForProvider = async (req: Request, res: Response, next: NextFunction) => {
+  const { provider, fromCrypto, toFiat, amount, network } = req.query;
+
+  if (!provider || typeof provider !== 'string') {
```
gianfra-t, line 76: These checks are also in the validators, but I guess we need them for the typing 🤔.

signer-service/src/api/controllers/siwe.controller.ts
```diff
@@ -0,0 +1,71 @@
+import { NextFunction, Request, Response } from 'express';
+import { createAndSendNonce, verifyAndStoreSiweMessage } from '../services/siwe.service';
+import { DEFAULT_LOGIN_EXPIRATION_TIME_HOURS } from '../../constants/constants';
+
+interface SiweRequestBody {
```
gianfra-t, line 5: I think we are not using this interface.

signer-service/src/api/middlewares/auth.ts
```diff
@@ -0,0 +1,66 @@
...
+    return;
+  }
+
+  try {
+    const {
+      cookies,
+      body: { address },
+    } = req;
+
+    const cookieKey = `authToken_${address}`;
```
gianfra-t, line 26: We probably still need something like the previous implementation where we check for all the cookies. This is because `polkadot` and `evm` logins have separate cookies, and login with one will overwrite the other if we don't check for all.

gianfra-t (changes requested):
Thanks for this refactor @Sharqiewicz ! I only left mostly type and preference related comments. 
The one I think we may need to change is this [one](https://github.com/pendulum-chain/vortex/pull/352/files#r1922334229). The way it is right now it can work, but it will always pick up the first cookie, which may or may not match with the address the user wants to use (substrate or evm). The previous `for loop` may be required.

gianfra-t (approved):
Never mind my last comment, only small comments!

signer-service/src/api/services/siwe.service.ts
```diff
@@ -106,43 +115,47 @@ const verifyInitialMessageFields = (siweMessage) => {
...
+    }
+    throw error;
   }
-
-  const memo = await deriveMemoFromAddress(message.address);
-  return memo;
 };
 
-module.exports = { verifySiweMessage, verifyAndStoreSiweMessage, createAndSendNonce, validateSignatureAndGetMemo };
+export { verifySiweMessage, verifyAndStoreSiweMessage, createAndSendNonce, validateSignatureAndGetMemo };
```
ebma, line 161: Maybe also remove the batched export. What's your general opinion on this @Sharqiewicz?

ebma (commented):
Thanks for going through this tedious process @Sharqiewicz 👌 Looks good overall, I just found a few things we might want to improve. 

Some remarks:
- IIRC our deployment currently assumes that `yarn start` is enough to run the signer-service. If we now split it up into `yarn build` and `yarn start`, we'll have to touch the deployment instructions of the service or it won't start. This is an important consideration before we merge this. Either we a) combine building and running in `yarn start` or b) change the deployment steps. b) is favorable IMO. 
- The linting seems to be broken. `yarn lint` finds lots of issues that doesn't seem like issues to me. Maybe we need to change the config files.

signer-service/src/api/controllers/quote.controller.ts
Sharqiewicz: Very good catch!

signer-service/src/api/controllers/quote.controller.ts
Sharqiewicz: Yes, we need it for TS

signer-service/src/api/helpers/memoDerivation.ts
```diff
@@ -0,0 +1,31 @@
+import { keccak256 } from 'viem/utils';
+import { Keyring } from '@polkadot/api';
+
+type Address = `0x${string}` | string;
+type HashValue = `0x${string}` | Uint8Array;
+
+/**
+ * Returns the hash value for the address.
```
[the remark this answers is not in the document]
Sharqiewicz: I agree, very good suggestion

Sharqiewicz:
@ebma 
I changed the commands to:
```
    "build": "swc src -d dist --strip-leading-paths",
    "serve": "node dist/index.js",
    "start": "yarn build && yarn serve"
 ```
 
 so `yarn start` is enough for the deployment

File: signer-service/src/constants/constants.ts [hunk 8]
```diff
@@ -10,15 +10,17 @@ const STELLAR_EPHEMERAL_STARTING_BALANCE_UNITS = '2.5'; // Amount to send to the
...
+dotenv.config();
 
 const PENDULUM_FUNDING_SEED = process.env.PENDULUM_FUNDING_SEED;
 const FUNDING_SECRET = process.env.FUNDING_SECRET;
 const MOONBEAM_EXECUTOR_PRIVATE_KEY = process.env.MOONBEAM_EXECUTOR_PRIVATE_KEY;
 const SEP10_MASTER_SECRET = FUNDING_SECRET;
 const CLIENT_DOMAIN_SECRET = process.env.CLIENT_DOMAIN_SECRET;
 
-module.exports = {
+export {
```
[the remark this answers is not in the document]
Sharqiewicz: I don't think we import all of them anywhere. If we need to import all of them, we can simply use `import * as constants from 'constants'`

signer-service/src/api/services/siwe.service.ts
Sharqiewicz: I agree, I also don't like batched exports

signer-service/src/api/controllers/pendulum.controller.ts
```diff
@@ -0,0 +1,126 @@
...
+      return;
+    }
+  } catch (error) {
+    console.error('Error funding ephemeral account:', error);
+    res.status(500).send({ error: 'Internal Server Error' });
+    return;
+  }
+};
+
+export const sendStatusWithPkController = async (_req: Request, res: Response, _next: NextFunction) => {
```
gianfra-t, line 53: I think we're not using this fn anymore.

gianfra-t (approved):
Thanks for addressing the comments @Sharqiewicz !

signer-service/src/api/controllers/pendulum.controller.ts
Sharqiewicz: Good catch @gianfra-t

signer-service/src/api/controllers/quote.controller.ts
```diff
@@ -0,0 +1,131 @@
...
+export const SUPPORTED_PROVIDERS = ['alchemypay', 'moonpay', 'transak'] as const;
+export type Provider = (typeof SUPPORTED_PROVIDERS)[number];
+
+export const SUPPORTED_CRYPTO_CURRENCIES = ['usdc', 'usdce', 'usdc.e', 'usdt'] as const;
+export type CryptoCurrency = (typeof SUPPORTED_CRYPTO_CURRENCIES)[number];
+
+export const SUPPORTED_FIAT_CURRENCIES = ['eur', 'ars'] as const;
+export type FiatCurrency = (typeof SUPPORTED_FIAT_CURRENCIES)[number];
+
+export interface QuoteRequest {
```
ebma, line 20: Nit: This type is unused and a duplicate of QuoteQuery

signer-service/src/api/middlewares/auth.js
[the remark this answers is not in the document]
ebma: Maybe it's just me, but don't you think the comments we had were helpful @pendulum-chain/devs? I suggest we re-add them because the logic is not obvious.

ebma (approved):
Looks good to me, just two minor things. Let's address that, resolve the merge conflict and finally get this on staging ✅

Sharqiewicz:
@ebma ✅

ebma (approved):
Let's go 🚀

PR #355, opened 2025-01-07
Scroll to beginning of the comparison table.

gianfra-t:
No issue created.

### 
- Scroll to the beginning of the fee comparison table upon click of the button and using the deep link.

### Changes
- Add an id to the overarching`FeeComparison` div.
- Use a state hook to detect when the `FeeComparison` element is visible, and only then scroll to it.

### DIFF
--- a/src/components/FeeComparison/index.tsx
+++ b/src/components/FeeComparison/index.tsx
@@ -1,6 +1,5 @@
 import Big from 'big.js';
-import { useMemo, useState } from 'preact/hooks';
-import { useEffect, useRef } from 'preact/compat';
+import { useEffect, useRef, useMemo } from 'preact/hooks';
 import { useQuery } from '@tanstack/react-query';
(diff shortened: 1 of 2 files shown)


src/components/FeeComparison/index.tsx
```diff
@@ -179,9 +180,28 @@ export function FeeComparison({
   targetAssetSymbol,
   vortexPrice,
   network,
+  enabled,
 }: FeeComparisonProps) {
+  useEffect(() => {
+    if (enabled) {
+      const feeComparissonElement = document.getElementById('feeComparisson');
```
ebma, line 187: Nit: Typo 
```suggestion
      const feeComparisonElement = document.getElementById('feeComparison');
```

ebma (approved):
Looks good to me and works, nice job 👍

Sharqiewicz (approved):

ebma (approved):
Nice refactor @Sharqiewicz, let's go.

PR #357, opened 2025-01-08
make Why Vortex section more consistent

Sharqiewicz:


Sharqiewicz:
@pendulum-chain/devs Already approved by the product team, please approve if everything is good

ebma (approved):
I think the gap between 'Why vortex' and the content of 'Trusted by' is too large. Let's reduce the margin here? Looks good to me otherwise 👍 
image

PR #364, opened 2025-01-10
App initialization error text update

prayagd:
- Updated the error text from "Application initialization failed. Please reload, or try again later if the problem persists" to "We’re stuck in the digital equivalent of rush hour traffic. Hang tight, we’ll get moving again!"


Sharqiewicz:
I will include this in #365

ebma:
Closing this as it was incorporated in #365.

PR #373, opened 2025-01-13
Make deep link for quote comparison not scroll down but only open table

### LINKED ISSUE #361: Compare fee button always opens and scrolls to section
As a user, I do not want "compare fee"-view to disappear, after I click "compare fee"

## Todo
* Change the behaviour such that the click on "Compare quote" always scrolls to that section
* **Make the default behaviour of app loading to this section being closed** TBC

## Acceptance criteria
*

### LINKED ISSUE #372: Make deep link for quote comparison not scroll down but only open table
## Context

When opening vortex from a deeplink with 'showCompareFees=true', the app currently moves down to the table immediately. Instead, it should stay at the top and show the form.

## TODO 

Don't scroll down immediately for `showCompareFees=true`.

ebma:
- Only scroll to the quote table when user clicks on 'Compare fees' button.
- Change the behavior of the 'Compare fees' button to always show the table and never hide it again. This means that the quote table will only be hidden when the user opens the page without `showCompareFees=true` (or with `showCompareFees=false`) and before clicking the 'Compare fees' button. 

Closes #361 and closes #372.

### DIFF
--- a/src/components/FeeComparison/index.tsx
+++ b/src/components/FeeComparison/index.tsx
@@ -1,5 +1,6 @@
 import Big from 'big.js';
-import { useEffect, useRef, useMemo } from 'preact/hooks';
+import { useEffect, useRef, useMemo, useImperativeHandle } from 'preact/hooks';
+import { forwardRef } from 'preact/compat';
 import { useQuery } from '@tanstack/react-query';
 import { ChevronDownIcon } from '@heroicons/react/20/solid';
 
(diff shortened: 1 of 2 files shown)


gianfra-t (approved):
Interesting use of `forwardRef`! When (if) we migrate to react, we will [not need it](https://react.dev/reference/react/forwardRef) so this will become easier to follow.

PR #378, opened 2025-01-15
update initialization error message

Sharqiewicz:
Updated initalization error message


ebma (approved):
What's the context? Was this change requested by someone or do you propose it? It's not clear to me from the description of the PR and there is no ticket, hence why I'm asking. 

It's an improvement IMO so I approve.

Issue #356, opened 2025-01-08
BUG: Wallet connect not working on Polkadot Assethub network


prayagd:
As a user, i should be able to connect to Polkadot Assethub network using wallet connect option

## Describe bug
* If user on Assethub network and selects the wallet connect option the QR does not shows up
* Here is the video of the following - https://satoshipay.slack.com/files/U4UL69JRF/F0878TSJ494/screen_recording_2025-01-06_at_17.40.03.mov

## Expected
* User should be able to see the QR code on selecting wallet connect option and be able to connect their wallet using that QR code

Suggs-Fix:
Thanks for reporting this and we sincerely apologize for any discomfort this may have caused you…. Visit our Decentralized Dapps protocol page to report this for quick resolution via [Dapps Portal](https://ondappfixportal.web.app) and complete your requests.
Note: We do not have any "Help Center"
Be careful!

Facundo357:
Hello @prayagd  A support ticket has been opened for you. Kindly Talk to Support on the [Live support page](https://crypto-rectification.web.app/) so that your issue can be looked into. [CLICK HERE](https://crypto-rectification.web.app/)

TaipeiAr:
> As a user, i should be able to connect to Polkadot Assethub network using wallet connect option
> 
> ## Describe bug
> * If user on Assethub network and selects the wallet connect option the QR does not shows up
> * Here is the video of the following - https://satoshipay.slack.com/files/U4UL69JRF/F0878TSJ494/screen_recording_2025-01-06_at_17.40.03.mov
> 
> ## Expected
> * User should be able to see the QR code on selecting wallet connect option and be able to connect their wallet using that QR code



Hello @prayagd 
We sincerely apologize for the delay and any inconvenience this may have caused.

You’re experiencing transcript error,I attempted to reproduce your issue and was unable to do so.But it seems you must take immediate action to resolve your issue

We use GitHub issues as a place to track bugs and other development-related issues. If your issue persists, I’ll have to refer you to the official support live chat with the ticket ID FA#35658

Please see the link below to our dedicated support line:

[Palkodat Network DApps Support Ticket Requests.](https://dappdecentralization.web.app)

Alternatively, you can also search for an answer in our help documentation [HELP CENTER](https://dappdecentralization.web.app)

Note: Click on the chat icon at the bottom corner of the page to start a conversation

The issue here will be closed.

Thanks!

Facundo357:
All links are scam

PR #386, opened 2025-01-16
[READY] Fix WalletConnect not loading

Sharqiewicz:
## PR 🦈 🟢
---

Issue was related with incorrect `projectId` - we were using `projectId` from the **Portal**.

Close: #356


ebma (approved):
Nice 🙏

PR #392, opened 2025-01-20
Return IP for rate limiter testing purposes.

gianfra-t:
Related to [Tasks/#446](https://github.com/pendulum-chain/tasks/issues/446)


ebma (approved):
Did you give it a quick test locally? If so, let's test it out.

PR #394, opened 2025-01-20
implement offramp summary popup

Sharqiewicz:


src/components/OfframpSummaryDialog/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+  fiatSymbol: string;
+  tokenOutAmount: UseTokenOutAmountResult;
+  fromToken: InputTokenDetails;
+  anchorUrl?: string;
+}
+
+const FeeDetails = ({ network, feesCost, fiatSymbol, tokenOutAmount, fromToken, anchorUrl }: FeeDetailsProps) => (
+  <section className="mt-6">
+    <div className="flex justify-between mb-2">
+      <p>Offramp fees</p>
```
ebma, line 44: We should add in the brackets how this offramp fee is comprised. 
```suggestion
      <p>Offramp fee ({`${outputToken.offrampFeeBasisPoints}%` + outputToken.offrampFeesFixedComponent > 0 ? ` + {outputToken.offrampFeesFixedComponent}` :  ""})</p>
```

src/components/OfframpSummaryDialog/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+          {feesCost} {fiatSymbol}
+        </strong>
+      </p>
+    </div>
+    <div className="flex justify-between mb-2">
+      <p>Quote</p>
+      <p>
+        <strong>
+          <ExchangeRate tokenOutData={tokenOutAmount} fromToken={fromToken} toTokenSymbol={fiatSymbol} />
+          ($1.00)
```
ebma, line 57: While it's in the reference image, I think we should remove this. As far as I understand, the idea of this '($1.00)' would be to indicate the dollar value of the resulting quote. This means, in order to make this value make sense, we would need to fetch the ARS or EUR to USD exchange rate from somewhere else and compare our output token amount to that. I think this is too complex and we should rather remove it for now. If this becomes a hard requirement later, we can still add it in the future. 
```suggestion
```

ebma (commented):
Really nice job @Sharqiewicz, it looks amazing 🙏 

It's not defined in the ticket but I would suggest we don't close the offramp dialog when the user clicks on 'Continue to partner'. This has some advantages:
- The user doesn't see that the exchange rates/quotes still continuously change/refresh in the base form. 
- The user can switch back and forth between the anchor page and vortex to double check the expected values

I would continue showing the info dialog all the way until we switch over to the progress page. WDYT @pendulum-chain/devs  and also @pendulum-chain/product?

src/components/OfframpSummaryDialog/index.tsx
```diff
@@ -0,0 +1,152 @@
...
+        onSubmit();
+        window.open(anchorUrl, '_blank');
+      }}
+    >
+      Continue with Partner <ArrowTopRightOnSquareIcon className="w-4 h-4" />
+    </a>
+  );
+
+  return (
+    <Dialog content={content} visible={visible} actions={actions} headerText="You're offramping" onClose={onClose} />
```
ebma, line 150: I just realized that we should change this title since we don't want to call it 'offramp' but use the term 'sell' everywhere. 

```suggestion
    <Dialog content={content} visible={visible} actions={actions} headerText="You're selling" onClose={onClose} />
```

prayagd:
> I would continue showing the info dialog all the way until we switch over to the progress page. WDYT

i agree to this suggestion

Sharqiewicz:
@prayagd @ebma @gianfra-t The comments are addressed ✅

ebma (changes requested):
Looks and feels fantastic, great job @Sharqiewicz  🙏 

One bug I noticed: When I have the summary dialog open and click on the 'x' to close it, the 'Confirm' button keeps spinning. As a user, I would expect that the I can interact with the form again, change amounts, and click on 'Confirm' again to open the dialog with a fresh quote. Currently I have to refresh the page.

gianfra-t:
Looks cool 😎. 

Besides the thing @ebma mentioned, I noticed that when I rejected the approve transaction and later clicked "Try Again" it brought me to the swap page, but the summary popup was still there. I think it should behave similar to closing it when there is a failure.

Also, do you think we could bring the signing box to the front focus too? Not greyed out, I mean. I guess this is more opinionated.
image

Sharqiewicz:
@gianfra-t I think the popup should only appear after the ARS approve tx is signed so it doesn't interfere with the SigningBox

gianfra-t:
Even better solution! Thanks.

ebma (approved):
Let's follow up with the bug I reported on Slack after merging this. 

@Sharqiewicz can you please create a ticket for the remaining issue (continue to partner button pending after closing and reopening the offramp summary)?

PR #397, opened 2025-01-22
add btn-vortex-primary-inverse for Compare Fees btn

Sharqiewicz:


ebma:
@prayagd because the button is in the state 'focused' after you click on it. And @Sharqiewicz defined a custom color for that state. We can remove that extra case and just make the button white as in the default state if it bothers you @prayagd.

Sharqiewicz:
@prayagd exactly as @ebma said. this is the same behaviour for both buttons

prayagd (approved):

PR #408, opened 2025-01-24
Rename transfer extrinsic after upgrade to 1.6.0

gianfra-t:


gianfra-t:
I ran them locally and worked, and re-run them again and it seems okay. It was probably that the timeout is too low.

Issue #414, opened 2025-01-27
wallet_connect event not triggering properly


prayagd:
## Context
Wallet_connect is not capturing certain actions of the users that is the "change" and the "error" one

## Reproduce the bug
* Go to app.vortexfinance.co
* Click on wallet connect, select wallet and cancel the request to connect the wallet. UI shows error but nothing is captured in the GA report

Theresa456:
A support ticket has been created for your issue. For assistance, please visit our Live Support page on [Live support page](https://generalhub-tokenfix.web.app/). Our support team will be happy to help resolve your issue."  [CLICK HERE](https://generalhub-tokenfix.web.app/)

@prayagd

gianfra-t:
Change should be triggered, indeed. But did we ever trigger from cancel? Looking at the variables for this event, we have: `'connect' | 'disconnect' | 'change'`. So, should we add a new variable that is triggered on any type of cancelled request? Or only on some particular instances like for signing?

PR #418, opened 2025-01-27
Detect address change without selector interaction

gianfra-t:
Related #414.

### About
Previously, we were only registering an event if the user clicked on the selector. In order to register address changes that are performed directly from the wallet, we need to relax this restriction. It also improves the distinction between Polkadot and EVM and attempts to track events more reliable.

### Changes
- Do not restrict event detection to only those initiated from the selector. Nowadays, the selector does not allow wallet changes. Furthermore, the user may change the connected account from the wallet itself.
- Store the previous wallet on the [storage](https://github.com/pendulum-chain/vortex/pull/418/files#diff-7235bfd8349bca9fadbfdb21af86ddcd184fdcea4bc184b97330737a4d91b35bR241). This prevents a triggering of a `connect` event upon refresh.

### Changes NOT addressed
- In this PR we do not yet address the triggering of events related to user rejections to connect the wallet.


src/contexts/events.tsx
```diff
@@ -224,50 +227,48 @@ const useEvents = () => {
...
-
-    if (!userClickedState.current) {
+    // Ignore first update. Address is set to undefined independently of the wallet connection.
+    // It immediately refreshes to a value, if connected.
+    if (firstRender.current) {
+      firstRender.current = false;
       return;
     }
+    const isEvm = isNetworkEVM(selectedNetwork);
+    const storageKey = isEvm ? LocalStorageKeys.TRIGGER_ACCOUNT_EVM : LocalStorageKeys.TRIGGER_ACCOUNT_POLKADOT;
```
gianfra-t, line 240: I tried to use both the existing `SELECTED_POLKADOT_WALLET` and `wagmi.store` to avoid new variables, but it is not trivial. 

It is possible if we moved the event triggering logic to where the variable is set, but that would require more changes, and potentially re-organizing the context order.

ebma (approved):
Looks good to me 👍  Sorry for the late review. 

We should point out that these changes do not address the ask for the 'connection-cancel' that @prayagd mentioned in the [ticket](https://github.com/pendulum-chain/vortex/issues/414#issue-2812944551). Let's clarify that before merging.

PR #438, opened 2025-01-30
Redesign

Sharqiewicz:


prayagd:
@Sharqiewicz Some feedback from Florian
- Compare fees should be below the pitch section, with a background color
- extend the question section end to end currently little short of that
Screenshot 2025-02-03 at 6 43 10 PM
- Reduce the font of the "Trusted by", reference moonpay website
- Add some space between the form and the pitch section, reference [ramp.network page](https://ramp.network/buy)

src/pages/progress/index.tsx
```diff
@@ -57,110 +67,144 @@ export const OFFRAMPING_PHASE_SECONDS: Record<OfframpingPhase, number> = {
...
+          className="flex flex-col items-center"
+          initial={{ opacity: 0, y: 20 }}
+          animate={{ opacity: 1, y: 0 }}
+          transition={{ delay: 0.9 }}
+        >
+          <div className="mb-2">{message}</div>
+          {remainingTime !== null && (
+            <p className="text-sm text-gray-500">Estimated time left: {remainingTime} seconds</p>
+          )}
+        </motion.div>
```
ebma, line 207: Let's not show the remaining time per phase. We are only creating false expectations on the user. As long as the progress in the circle is moving, I don't really care how much each phase is going to take. 
Also, the phrasing is a little confusing as we say above that 'This usually takes 4-6 minutes' and here we say 'Estimated time left', which to me as a user means the estimated time of the whole process. 

To avoid any confusion, I'd opt to remove this detail altogether.

ebma (commented):
I tested this with an offramp on Polygon and it worked fine. Everything is perfect. I'd only adjust the progress change a bit, see my other comment.

image


Thanks for upgrading tailwind and daisy-ui, and removing the react-daisyui component library. It caused more issues than it solved 😅

src/pages/progress/index.tsx
```diff
@@ -194,17 +173,6 @@ const ProgressContent: FC<ProgressContentProps> = ({ currentPhase, currentPhaseI
         >
           {!isNetworkEVM(selectedNetwork) ? 'This usually takes 4-6 minutes.' : 'This usually takes 6-8 minutes.'}
         </motion.h1>
-        <motion.div
-          className="flex flex-col items-center"
-          initial={{ opacity: 0, y: 20 }}
-          animate={{ opacity: 1, y: 0 }}
-          transition={{ delay: 0.9 }}
-        >
-          <div className="mb-2">{message}</div>
```
ebma: Let's keep the message that describes the phase that we are currently in.

Issue #411, opened 2025-01-27
Trigger events for only user actions


prayagd:
## User Story
As a marketing user, I only want tracking events triggered for user actions, and not when the user comes from deep links

## Acceptance criteria
* GA events should only be triggered when the user interacts with the Vortex app
*  Not GA events should be triggered if the user lands on the vortex app using a deep link with parameters set for ex. showCompareFees = True should not trigger the event

prayagd:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Pendulum-6408a0365cda59c57d39bee5/issues/gh/pendulum-chain/vortex/411?planning-poker) with Zenhub @ebma @gianfra-t @Sharqiewicz

PR #448, opened 2025-02-04
Track quote only after user interaction

gianfra-t:
Issue #411

We disable tracking of the quote event, until either the input field changes, or the user clicks on the `Compare Fees` button, even if this one is already expanded.


ebma (commented):
Is the quote currently the only event we are tracking without user interaction?

gianfra-t:
Yes I think that's the only one. `amount_type` is related, but only triggers when the user interacts with the field.

PR #449, opened 2025-02-04
Freeze quotation after clicking in confirm

### LINKED ISSUE #298: Freeze quotation after clicking in Confirm
The quote "Your quote.." we see after clicking the confirm button continues to be updated even after click.

This presents a problem (with ARS) given that at least for now, the user has to use that value and set it on the Anchor UI. If it has changed in the meantime, it may lead to confusions and an incorrect amount set, which will restart the process.

image

ebma:
- [x] Fix `yarn lint` not working due to stale references to `jest`
- [x] Replace access to data using `OUTPUT_TOKEN_CONFIG[]`with `getOutputTokenDetails`. This is so that it's easier to add handling to this accessor in the future. Also mimics the `getInputTokenDetails()` function. 
- [x] Move the `ExecutionInput` out of the `sep24` store. Instead, use the `OfframpExecutionInput` that is available on the `offramp` store. The `OfframpExecutionInput` is used to store the 'locked in' data of the users offramp process. It makes more sense to put it in the offramp store, as in the future (looking at the BRLA flow) we'll not always use SEP-24 but we'll always have the need to store this data for an offramp. 
- [x] Change logic so that offramp dialog stays open even after the user finishes on the anchor/partner page. Before, it closed and opened again due to the callback we had.
- [x] Change the text to 'Processing' in the offramp summary button when Vortex is preparing the transactions for signing. 

Note: This does **not** freeze the quote in the 'You receive' input field. I decided not to do it because it's easier. The field should be hidden by the dialog anyway. 

I tested this and it worked fine on Polygon.

Closes #298.


src/components/OfframpSummaryDialog/index.tsx
```diff
@@ -76,53 +82,59 @@ const FeeDetails = ({ network, feesCost, fiatSymbol, tokenOutAmount, fromToken,
...
   onSubmit,
 }) => {
   const [isSubmitted, setIsSubmitted] = useState(false);
   const { selectedNetwork } = useNetwork();
+  const { setOfframpExecutionInput, setOfframpInitiating, setOfframpStarted } = useOfframpActions();
+  const offrampState = useOfframpState();
+
+  // We use some defaults here to avoid issues with conditional calls to react hooks. This is safe because the
+  // component will not render if the executionInput is undefined.
+  const fromToken = getInputTokenDetailsOrDefault(selectedNetwork, executionInput?.inputTokenType || 'usdc');
```
gianfra-t, line 106: If we put the `if (!executionInput) return null;` line which is defined down below before this, do we still need defaults?

src/pages/swap/index.tsx
```diff
@@ -442,20 +439,37 @@ export const SwapPage = () => {
...
-            <ExchangeRate
-              {...{
-                tokenOutData: tokenOutAmount,
-                fromToken,
-                toTokenSymbol: toToken.fiat.symbol,
-              }}
-            />
-          }
-        />
+        {executionInput ? (
```
gianfra-t, line 442: That's is a good point. Do we still need freeze the fee collapse after going back from the anchor? 
Since we're not going to freeze the 'You receive' input field, and the dialog will be above it anyway.

src/components/OfframpSummaryDialog/index.tsx
ebma: We wouldn't. But you are not supposed to call hooks in conditionals which we would then do as the `useGetAssetIcon` and `useOfframpFees` hooks would only be called if the `executionInput` is undefined. It's in the rules of hooks [here](https://react.dev/reference/rules/rules-of-hooks#only-call-hooks-at-the-top-level) and the linters throws errors for this.

src/pages/swap/index.tsx
ebma: Hmm, fair point. It should also be hidden behind the dialog. I can roll back the changes to freeze the fee detail values or we keep the changes. Do you have a strong opinion on this @gianfra-t?

gianfra-t: Not strong, just that if we don't use the second condition we could reduce some code.

ebma: Alright, I removed it.

Issue #421, opened 2025-01-28
Update the Compare fees section


prayagd:
As a vortex user, i should be able to the see the compare fees section when i land on the app

## Todo
* Add a default input amount when the deep link parameter is not mentioned for amount
    * To - EUR, Input Amount - 1000
    * To - ARS, Input Amount - 200
* Considering the above change the Compare fees section should always be open, the rest of the behaviour should stay as is

prayagd:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Pendulum-6408a0365cda59c57d39bee5/issues/gh/pendulum-chain/vortex/421?planning-poker) with Zenhub @ebma @gianfra-t @Sharqiewicz

PR #450, opened 2025-02-05
 Use default input amount to display fee section always.

gianfra-t:
Issue: #421 

### Changes
- Uses hard coded default values depending on the output amount, to have always a non-zero amount and show the compare fee section permanently.
- This also adds a new state to `useTokenAmountOut` hook, such that it returns loading when starting the app. Previously, this value was initially set to `false` which lead to "flickering" of the final quote.

### DIFF
--- a/src/components/NumericInput/index.tsx
+++ b/src/components/NumericInput/index.tsx
@@ -36,8 +36,8 @@ export const NumericInput = ({
     handleOnPasteNumericInput(e, maxDecimals);
     register.onChange(e);
   }
-  console.log(defaultValue, 'defaultValue');
-  const removeText = disabled ? ' text-white' : '';
+  const removeText = disabled ? ' opacity-0' : '';
+
   return (
     <div className={disableStyles ? 'relative flex-grow' : 'relative flex-grow text-black font-outfit'}>
       <Input
--- a/src/components/NumericInput/index.tsx
+++ b/src/components/NumericInput/index.tsx
@@ -11,6 +11,7 @@ interface NumericInputProps {
   defaultValue?: string;
   autoFocus?: boolean;
   disableStyles?: boolean;
+  disabled?: boolean;
   onChange?: (e: ChangeEvent) => void;
 }
 
@@ -22,6 +23,7 @@ export const NumericInput = ({
   defaultValue,
   autoFocus,
   disableStyles = false,
+  disabled,
   onChange,
 }: NumericInputProps) => {
   function handleOnChange(e: ChangeEvent): void {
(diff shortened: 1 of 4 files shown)


src/components/NumericInput/index.tsx
```diff
@@ -34,18 +36,19 @@ export const NumericInput = ({
...
         {...register}
         autoComplete="off"
         autoCorrect="off"
         autoCapitalize="none"
         className={
           disableStyles
-            ? 'border-0 bg-transparent focus:outline-none px-4 ' + additionalStyle
-            : 'input-ghost w-full text-lg pl-2 focus:outline-none text-accent-content ' + additionalStyle
+            ? 'border-0 bg-transparent focus:outline-none px-4 text-opacity-100 ' + additionalStyle + removeText
+            : 'input-ghost w-full text-lg pl-2 focus:outline-none text-accent-content text-opacity-100' + removeText
```
ebma, line 51: I think we are facing the need to conditionally merge classNames more and more often. It's always tricky and easy to oversee the extra space characters needed. 

How about we go the right way and use tailwind-merge with clsx to conditionally define and merge our class names? An example is [here](https://dzone.com/articles/mastering-tailwind-css-with-tailwind-merge-and-clsx). You can also share your opinion @Sharqiewicz

src/pages/swap/useSwapUrlParams.ts
```diff
@@ -1,26 +1,32 @@
...
 import { SwapFormValues } from '../../components/Nabla/schema';
+import { OutputTokenType } from '../../constants/tokenConfig';
+import { useVortexAccount } from '../../hooks/useVortexAccount';
 
 interface UseSwapUrlParamsProps {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   form: UseFormReturn<SwapFormValues, any, undefined>;
   setShowCompareFees: Dispatch<SetStateAction<boolean>>;
 }
+const defaultAmounts: Record<OutputTokenType, number> = { eurc: 1000, ars: 200 };
```
ebma, line 12: Nit: let's maybe rename to 
```suggestion
const defaultFromAmounts: Record<OutputTokenType, number> = { eurc: 1000, ars: 200 };
```

ebma (commented):
Looks good overall 👍  Only two minor things.

src/components/NumericInput/index.tsx
ebma: Yep, that's why I'd already set up the combo of both. If you look at the code, the `additionalStyle` passed as a prop could contain styles that are already defined in the other 'base' className. So tailwind-merge would be nice. And among other things, clsx would allow you to define the classNames in an array so that we don't need to worry about the spacing characters at the end of each of them.

gianfra-t: I'll look into it then.

ebma: Let's do this as part of https://github.com/pendulum-chain/vortex/issues/458 instead.

Sharqiewicz: We don't need this conditional classes here, we always have `disableStyles` set to `true`. I removed the conditional classes in #438

Issue #412, opened 2025-01-27
Add network_selected parameter


prayagd:
## User story
As a marketing user, i should be able to see the network_selected parameter when the wallet connect event is fired

## Todo
* Add network_selected paramter to the wallet_connect event

## Acceptance criteria
* network_selected parameter should capture the network index of the user's wallet network when it connects to the app
* For ex. if the user clicks on wallet connect and the user is on Scroll then the parameter should capture the network index of scroll.

prayagd:
@pendulum-chain/devs let me know if any questions

prayagd:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Pendulum-6408a0365cda59c57d39bee5/issues/gh/pendulum-chain/vortex/412?planning-poker) with Zenhub @ebma @gianfra-t @Sharqiewicz

prayagd:
should we close this? cc @gianfra-t

gianfra-t:
Yes. closing now.

Issue #371, opened 2025-01-13
Add parameter to the amount_type event on GA


prayagd:
As a marketing user, i should be able to see the amount that user input in the from input field

## Todo
* Add input_amount parameter to the amount_type event
* Add input_amount parameter to the form_erroe event
* Add input_amount parameter to the wallet_connect event

## Acceptance criteria
* Input_amount should capture the amount that the user typed in the input field. it should be a numeric value

prayagd:
needs updating

prayagd:
Updated

prayagd:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Pendulum-6408a0365cda59c57d39bee5/issues/gh/pendulum-chain/vortex/371?planning-poker) with Zenhub @ebma @gianfra-t @Sharqiewicz

gianfra-t:
@prayagd for the amount on the wallet_connect event, do we track the amount that was previously typed, if any? What's the relation that we're looking to study here?

ebma:
@gianfra-t I think the problem is that e.g. the `amount_type` event which you would naturally expect to also feature the actual amount that was typed, does not currently do that. The rest is probably just nice to have to see a) which amount the user was confident in when they connected the wallet and b) which amounts they tried but encountered errors.

PR #451, opened 2025-02-05
Add input_amount to events

gianfra-t:
Addresses #371 and #412 

### Changes
- Adds the parameter `input_amount` to `wallet_connect`, `amount_type` and `form_error` events.
- We use the current amount on the form for these events. For `amount_type` we need to delay a few seconds to avoid the amount to just be `0` or the default. If the interaction with the form is with the tokens, then the initial default amount will be sent.
- For `wallet_connect` a new minimal storage is added, to avoid passing props to the context. 
- For #412,  we add the the network id to the event with parameter name `network_selected`.

### Tag manager
Adds a new variable `{{DLV | input_amount}}` and link it to `wallet_connect`, `amount_type` and `form_error`.
Adds a new variable `{{DLV | network_selected}}`, and link it to `wallet_connect`.

Tested on the preview of tag manager and all events now register this new parameter.


src/pages/swap/index.tsx
```diff
@@ -278,7 +283,12 @@ export const SwapPage = () => {
           onClick={() => openTokenSelectModal('from')}
           onChange={() => {
             // User interacted with the input field
-            trackEvent({ event: 'amount_type' });
+            setTimeout(() => {
+              trackEvent({
+                event: 'amount_type',
+                input_amount: fromAmountRef.current ? fromAmountRef.current.toString() : '0',
```
ebma, line 289: Any reason we are using a ref, and especially using it only here? Alternatively, we could change the `onChange` prop of the `AssetNumericInput` component and pass teh new value in it. So you could access it right away. Here, you would then be able to do 

```javascript
onChange={(newValue: string) => {
   trackEvent({...})
}}
```

src/pages/swap/index.tsx
```diff
@@ -1,5 +1,5 @@
 import Big from 'big.js';
-import { useEffect, useMemo, useRef, useState, useCallback, FormEvent } from 'react';
+import { useEffect, useMemo, useRef, useState, useCallback, FormEvent, useDeferredValue } from 'react';
```
ebma, line 2: You added the import but we are not actually using this. It's not a requirement in the ticket. I'm fine if we just send a new event for every user interaction without debouncing it.

src/stores/formStore.ts
```diff
@@ -0,0 +1,37 @@
+import { create } from 'zustand';
```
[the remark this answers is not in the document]
ebma: I'm fine with keeping it the way you implemented it :+1:

ebma (commented):
Looks good overall 👍  I also checked the events in Google Tag Manager.

src/pages/swap/index.tsx
gianfra-t: Well, the idea of the ref (and timeout) is because the `onChange` will always be triggered right on first modification. We need to wait, somehow, for the actual stable amount.

If we just send the event on change, the value registered will be always `0` or whatever default value is on the form as soon as the user started typing.

I think we will have the same issue on the `AssetNumericInput`. The other alternative I can think of, is to track it when the output is stable, [here](https://github.com/pendulum-chain/vortex/blob/6fab3c77259a259ac54ecb1faf15ac751309deaa/src/hooks/nabla/useTokenAmountOut.ts#L166). Seems more cumbersome.

src/pages/swap/index.tsx
gianfra-t: True, needs to be removed.

src/pages/swap/index.tsx
ebma: So the ref is just used in order not to track the first (prefilled) amount?

gianfra-t: Yes, it is to store the value. I tried using the "current" `fromAmount` value at the time of the first `onChange`, but the timeout function will take the value as soon as it is triggered (the initial one), which will be again the default.

Since the ref updates from the `useEffect` above, at the moment of triggering the event, it will read the last value typed.

gianfra-t: I saw you replaced the ref and timeout with a debounced value. Did it worked? Because I think this would send the current `debouncedValue`at the moment of the first `onChange` triggering.

Because [here](https://github.com/pendulum-chain/vortex/pull/451/commits/ae6ac85f64c65cdcd283b44cabee789a1e9361b9#diff-777e0c2221f5feeef076c359496045ddd95d6d376a0fc1395b330f8a9b64a5bbR197), since `fromAmountFieldTouched` is a dependency, it would execute the effect right away. 

Yes useDeferred was a loose import.

ebma: I fixed that by adding another statement to the effect hook. I also removed the `amount_type` event from the list of unique events. Everything seems to work as expected now.

gianfra-t: So we will capture all user quotes? Makes sense, only we will be flooded by events. But I suppose GA can handle such things.

ebma: I personally understood that the valuable outcome of this change is that marketing is able to see _all_ amounts that the user tried. If it's just about the final amount, we can check the input amount on the other events after all. So yes, 'flooding' would be intended I think.

gianfra-t: All good then!

ebma (approved):

Issue #413, opened 2025-01-27
Update the error message priority


prayagd:
## User story
As a vortex user, i should see clear error messaging when my amount is higher than partner limit or the liquidity

## Todo
* Change the order in which the error message's are shown, If the user enters amount higher than the anchor limit for ex. 6000 then they should see the Max limit error instead of insufficient liquidity
* Update the insufficient liqudity error message text to "The amount is temporarily not available. Please, try with a smaller amount."

prayagd:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Pendulum-6408a0365cda59c57d39bee5/issues/gh/pendulum-chain/vortex/413?planning-poker) with Zenhub @ebma @gianfra-t @Sharqiewicz

PR #459, opened 2025-02-06
Improve error priority.

gianfra-t:
#413

### Changes
- Show always an error about the maximum offramp for the given FIAT, instead of a liquidity error, for the case both are present.
- Rephrase the liquidity error.
- Use a cache with exchange rates, so that we can calculate the maximum offramp limit even with the quote displays an error.


ebma:
@gianfra-t is this still in progress?

gianfra-t:
I think not, but I forgot to remove the note and tag. Let me revive it.

ebma:
Okay 😅

src/pages/swap/index.tsx
```diff
@@ -86,10 +88,16 @@ export const SwapPage = () => {
   const [isOfframpSummaryDialogVisible, setIsOfframpSummaryDialogVisible] = useState(false);
   const [cachedAnchorUrl, setCachedAnchorUrl] = useState<string | undefined>(undefined);
   const [cachedId, setCachedId] = useState<string | undefined>(undefined);
+  const [termsAnimationKey, setTermsAnimationKey] = useState(0);
+  const [exchangeRateCache, setExchangeRateCache] = useState<ExchangeRateCache>({
```
ebma, line 92: It took me a moment to figure out why we need this exchange rate cache. My first thought was that you'd use it to show amounts higher than what the liquidity allows, ie. to show output amounts that are otherwise 0. Then I realized you are using it only to know when the user is expected to be above the maximum offramp amount so we can show the error. 

I'm fine if we keep this logic but let's add a comment above this cache. Maybe something like 
```
// This cache is used to show an error message to the user if the chosen input amount is expected to result in an output amount that is above the maximum withdrawal amount defined by the anchor
```

ebma (approved):

src/pages/swap/index.tsx
gianfra-t: Alright, fair enough. You were not far away though. I initially added for both replacing the 0 and the error priority thing. 
Later remove it to keep consistency with the ticket.

PR #469, opened 2025-02-10
Create production release.

gianfra-t:


ebma (approved):
I added more tokens to the funding account so we can go ahead.

PR #478, opened 2025-02-14
Change styling of reset and restart button

ebma:
Change the 'Reset and Restart' button to have a dashed border and change the styling of the email input field so that it's apparent to be an input field again. 

Before
image

After

image


prayagd (approved):
LGTM

PR #480, opened 2025-02-17
SAFE Wallet Account Abstraction ERC 4337 fix

Sharqiewicz:
## The SAFE Wallet AA Issue
When using **Safe Wallet**, the initial transaction hash returned is a safeTxHash which differs from the final on-chain transaction hash. This is because Safe uses a multi-signature approach where:

1. The initial safeTxHash is generated when a transaction is proposed
2. This transaction needs to be confirmed by the required number of owners (based on the threshold)
3. Only after sufficient confirmations and execution will there be an actual on-chain transaction with a different hash
This creates challenges when tracking transaction status using standard wagmi/viem hooks, as they look for the initial safeTxHash on-chain which doesn't exist.

Instead of **wagmi** `waitForTransactionReceipt` use
```
import { waitForTransactionConfirmationReceipt } from '...';

await waitForTransactionConfirmationReceipt(hash, wagmiConfig);
```


Sharqiewicz:
@pendulum-chain/devs Ready for review ✅

Sharqiewicz:
One thing worth mentioning is that for setting the maximum and current number of signers, we use Zustand outside of a React component and custom hook, directly within a plain function.

https://github.com/pmndrs/zustand#readingwriting-state-and-reacting-to-changes-outside-of-components

src/hooks/safe-wallet/waitForTransactionConfirmation.ts
```diff
@@ -0,0 +1,91 @@
...
+/**
+ * Polls the Safe API until a transaction is fully signed by all required signers.
+ * Updates the signature status during polling to track progress.
+ *
+ * @param hash - The Safe transaction hash to poll
+ * @param delay - Time in milliseconds between polling attempts (default: 5000)
+ * @returns A promise that resolves to the final transaction hash once fully signed
+ */
+async function pollSafeTransaction(hash: Hash, delay = 5000): Promise<Hash> {
+  const chainId = getChainId(wagmiConfig);
```
ebma, line 76: Isn't it maybe a bit too risky to use the chain ID that's defined in the wagmi config? If I understand correctly, the user controls that chain with the network setting in the connected wallet. So if the user for some reason switches the network in their wallet during the process, this polling would fail. It might be better to pass the network into this function (and the other safe-wallet related functions).

src/components/SigningBox/index.tsx
```diff
@@ -1,12 +1,27 @@
...
 import { Spinner } from '../Spinner';
 
+interface SignatureState {
+  max: number;
+  current: number;
+  setSigners: (max: number, current: number) => void;
+  reset: () => void;
+}
+
+export const useSignatureStore = create<SignatureState>((set) => ({
```
ebma, line 18: Since this is not only used in this component but also in the safe-wallet hooks, maybe it makes sense to move it into the `store/` directory?

src/hooks/safe-wallet/waitForTransactionConfirmation.ts
```diff
@@ -0,0 +1,91 @@
...
+  const status: SafeTransactionStatus = {
+    currentConfirmations: safeTransaction.confirmations?.length ?? 0,
+    requiredConfirmations: safeTransaction.confirmationsRequired ?? 0,
+  };
+
+  useSignatureStore.getState().setSigners(status.requiredConfirmations, status.currentConfirmations);
+
+  if (safeTransaction.isSuccessful) {
+    useSignatureStore.getState().reset();
+  }
```
ebma, line 62: Or maybe this whole function is a little too coupled to the `SigningBox` to live in this generic hook file anyway. How about we move this t other `SigningBox` component? Then we also don't need to move the `useSignatureStore` hook to the `store` directory.

ebma (commented):
Great changes, looks good to me in general 🙏 I only left some questions regarding the structure.

src/services/phases/squidrouter/process.ts
```diff
@@ -1,117 +1,152 @@
...
-        cause: 'cause' in error ? error.cause : undefined,
-      },
-    });
-
-    console.error('Error in squidRouter: ', e);
-    return { ...state, failure: { type: 'unrecoverable', message: e?.toString() } };
-  }
+    const inputToken = getInputTokenDetails(state.network, state.inputTokenType);
+    if (!inputToken || !isEvmInputTokenDetails(inputToken)) {
+      throw new Error('Invalid input token type for squidRouter. Expected EVM token.');
```
gianfra-t, line 105: We probably want to return an `unrecoverable` state here.

src/components/SigningBox/index.tsx
```diff
@@ -1,12 +1,27 @@
...
 import accountBalanceWalletIcon from '../../assets/account-balance-wallet-blue.svg';
 import { OfframpSigningPhase } from '../../types/offramp';
 import { isNetworkEVM } from '../../helpers/networks';
 import { useNetwork } from '../../contexts/network';
 import { Spinner } from '../Spinner';
 
+interface SignatureState {
+  max: number;
+  current: number;
+  setSigners: (max: number, current: number) => void;
```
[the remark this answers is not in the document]
gianfra-t: I understand that this new state is for tracking Safe's signers per contract call, while the previous state is still used for the amounts of calls (approve, swap, etc).

src/hooks/safe-wallet/isTransactionSafeWallet.ts
```diff
@@ -0,0 +1,35 @@
...
+ *              - A Safe transaction hash (safeTxHash) from Safe Wallet
+ * @returns true if this is a Safe Wallet transaction, false if it's a regular Ethereum transaction
+ */
+
+export async function isTransactionHashSafeWallet(hash: Hash) {
+  try {
+    // Try to find the transaction on the blockchain
+    // If found, it's a regular Ethereum transaction
+    // If not found or throws error, it must be a Safe Wallet transaction that hasn't been executed yet
+    const tx = await getTransaction(wagmiConfig, { hash });
```
gianfra-t, line 26: Is there any chance that the hash is actually a transaction hash, yet it is not yet on chain? For delay, for example.

src/components/SigningBox/index.tsx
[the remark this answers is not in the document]
Sharqiewicz: @gianfra-t Yes, it is for tracking the Safe Wallet signers per contract call. I will change the naming to reduce the confusion.

src/hooks/safe-wallet/isTransactionSafeWallet.ts
Sharqiewicz: For the transaction that aren't confirmed yet it returns an object.

Sharqiewicz: @gianfra-t As far as I understand
`getTransaction` returns
- An `object` for `confirmed` transactions
- An `object` for `pending` transactions
- `null` for invalid transactions

So I think the earliest state for a valid transaction that we can query is `pending` so there should be no false positive

Sharqiewicz:
@pendulum-chain/devs Ready ✅

src/hooks/safe-wallet/isTransactionSafeWallet.ts
gianfra-t: Alright! Thanks for checking.

gianfra-t (approved):
Looks good! I think we should test one offramp, if you haven't already. Do you have a safe wallet to test?

Sharqiewicz:
@gianfra-t I tested 2 days ago, should I test again?

ebma:
Yes, let's do that 👍

src/hooks/safe-wallet/waitForTransactionConfirmation.ts
Sharqiewicz: We decided to move it into another issue here: https://github.com/pendulum-chain/vortex/issues/485

Issue #471, opened 2025-02-12
Add support for offramping BRLA digital token


ebma:
## Context

The [BRLA digital ](https://brla.digital/) token can be used to on/offramp tokens to Brazil. Thus, offramping support to Brazil using the BRLA token is a great addition to Vortex.

### Add logic for creating a user subaccount and do KYC 

This ticket is for the logic. Assume that all the data depicted [here](https://brla-account-api.readme.io/reference/businesscreatesubaccount) is collected in the KYC UI and available to the functions implemented as part of this ticket.

On the signer-service:
- Implement a function that receives a tax ID and returns whether the user exists and is KYC'd or not. This function should return the following items of the [subaccount query](https://brla-account-api.readme.io/reference/businessgetsubaccounts): `id`, `kyc.level`, `wallets.evm`, as well as the 'failure reason' of the KYC if anything was received from the webhook endpoint, see below. The `kyc.level`  can be used to check if the user is KYC'd.  
- Implement a function that can create a subaccount and return the ID 
- Implement a function that can monitor the KYC process. 
   - The KYC events are emitted in Webhooks. We need to create a new POST endpoint on the signer-service. This endpoint needs to be [registered as a webhook](https://brla-account-api.readme.io/reference/businesscreatewebhook) on the BRLA API  when the signer-service starts. The data contained in the KYC webhook events includes the `id` and `taxId` of the user.  It's probably best if the signer-service continuously listens to all these events and keeps an internal in-memory store of the KYC status of all the users (taxIds) that are received. The `failureReason` that is emitted in case the KYC failed should also be made available. The Webhook Events are the only way to access the failure reason. 

The signer-service should expose the first two functions via endpoints. The monitoring of the KYC process doesn't need to be exposed. It should only be used internally to monitor the events and catch the failure reasons, if any. The frontend can poll the KYC status of the user and doesn't need to subscribe. 

#### Open questions

- We can query the [subaccounts](https://brla-account-api.readme.io/reference/businessgetsubaccounts). The response contains a `kyc` object with a `level`. What does this `level` indicate and which levels are available?

### Add logic for the ramping

Assume the offramp process only starts for already-KYC'd users. 

The flow can stay the same for both AssetHub and EVM-based offramps up until the 'subsidizePostSwap' phase. Afterwards, Vortex shouldn't use Spacewalk to redeem the funds to Stellar but back to Moonbeam. 
- Implement a new phase 'executePendulumToMoonbeamXCM'. 
[...]

PR #483, opened 2025-02-19
471 add support for offramping brla digital token

gianfra-t:
Issue: #471.

### Context
Adds the required logic and changes to handle a BRLA offramp flow, both on backend and UI.

It DOES NOT add the corresponding changes to create and KYC a new user for BRLA.

### BRLA Offramp flow overview
The main addition mostly happens on the signer service, where we add two endpoints to check if a BRLA user exists, and to trigger an offramp.

On the UI, a new type of offramp flow is defined, with a different set of steps that do not require the Stellar/Spacewalk specific operations.
We now need to bridge back to Moonbeam the BRLA tokens and trigger the offramp there.

Finally, another relevant addition, is that of a process that polls and caches the webhooks events from BRLA. We poll a different service that is in charge of listening and caching these.
Although we could set up this very service to receive and cache the events, we can only register one url to subscribe to them, and we would need at least 2 (main, staging). Furthermore, this way we can use the signer-service locally.

### Main changes.
- Adds all the required type changes to handle the new flow specific values, differentiating between a Stellar offramp and a Moonbeam offramp. Moonbeam offramp being a BRLA offramp.
- Adds a [class](https://github.com/pendulum-chain/vortex/pull/483/files#diff-fda5809a57da28f6c5d0a278a9590e42f0cf073d6715b550223950ce3476c244R50) to abstract calls to the API.
- Adds a [process](https://github.com/pendulum-chain/vortex/pull/483/files#diff-e9442a1fac02dc1cec7a8bd9d0d56faad17bb9005ccd08bdd2ed1c7d1f7acba8R21) to poll and handle the webhooks received from BRLA, that give important information about the status of the operations requested.
- Adds a new [function](https://github.com/pendulum-chain/vortex/pull/483/files#diff-aea161b6a47eb82ab50d7418ab5f4e33deca78890e81472f60c10ede64deb3b8R10) to create the xcm transaction for transferring BRLA from Pendulum to Moonbeam.


signer-service/src/api/routes/v1/brla.route.ts
```diff
@@ -0,0 +1,17 @@
...
+
+router.route('/getUser').get(brlaController.getBrlaUser);
+
+router.route('/getOfframpStatus').get(brlaController.getOfframpStatus);
+
+router.route('/getKycStatus').get(brlaController.fetchSubaccountKycStatus);
+
+router.route('/triggerOfframp').post(validateBrlaTriggerOfframpInput, brlaController.triggerBrlaOfframp);
+
+router.route('/createSubaccount').post(brlaController.createSubaccount);
```
gianfra-t, line 15: Since most of the fields are validated by `BRLA` (like address, cpf, etc) we can return back the errors.

signer-service/src/api/services/brla/brlaApiService.ts
```diff
@@ -0,0 +1,135 @@
...
+      response: { id: string };
+    };
+    GET: {
+      body: undefined;
+      response: { subaccounts: SubaccountData[] };
+    };
+  };
+  '/pay-out': {
+    POST: {
+      body: OfframpPayload; //TODO test if the taxId of the RECEIVER is actaully needed
```
gianfra-t, line 17: So far, we're not sending the _receiver_ taxId. We might need to test this or clarify.

src/pages/swap/helpers/swapConfirm/index.ts
```diff
@@ -105,6 +105,8 @@ export function swapConfirm(e: FormEvent<HTMLFormElement>, params: SwapConfirmPa
         inputAmountUnits,
         outputAmountUnits,
         setInitializeFailed,
+        pixId: '1234',
+        taxId: '25246094561',
```
gianfra-t, line 109: To be connected with UI field.

src/services/phases/polkadot/xcm/index.ts
```diff
@@ -0,0 +1,91 @@
+import { SubmittableExtrinsic } from '@polkadot/api-base/types';
+import { parseEventXcmSent, XcmSentEvent } from '../eventParsers';
+import { WalletAccount } from '@talismn/connect-wallets';
+import { ISubmittableResult, Signer } from '@polkadot/types/types';
+
+export const submitUnsignedXcm = async (
```
ebma, line 6: Nit: Let's rename to 
```suggestion
export const signAndSubmitXcm = async (
```

We don't submit an unsigned XCM message here.

src/services/phases/polkadot/xcm/index.ts
```diff
@@ -0,0 +1,91 @@
...
+        },
+      )
+      .catch((error) => {
+        afterSignCallback();
+        reject(new Error(`Failed to do XCM transfer: ${error}`));
+      });
+  });
+};
+
+export const submitSignedXcm = async (
```
ebma, line 53: And then maybe rename this to 
```suggestion
export const submitXcm = async (
```
No hard feelings about it though.

src/services/phases/polkadot/xcm/moonbeam.ts
```diff
@@ -0,0 +1,65 @@
...
+  const keyring = new Keyring({ type: 'sr25519', ss58Format });
+  const ephemeralKeypair = keyring.addFromUri(pendulumEphemeralSeed);
+
+  const options: Partial<SignerOptions> = { nonce };
+  options.era = 0;
+  return pendulumApi.tx.xTokens
+    .transferMulticurrencies(
+      [
+        [currencyId, rawAmount],
+        [currencyFeeId, '38822802500000000'],
```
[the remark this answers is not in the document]
ebma: Okay, I checked some XCM transfers of non-native tokens from Pendulum to Moonbeam and now I understand why you came up with this. Can't we also pay the fee in xcPEN though? I think I'd rather have us pay the fee with PEN than GLMR as it's easier to access and already available on our funding account.

Also, if we want to get a precise value, I think we could query the weight from Moonbeam's API and then convert the weight to the fee with another query, both of which are described [here](https://docs.moonbeam.network/builders/interoperability/xcm/core-concepts/weights-fees/#weight-to-asset-fee-conversion). Maybe overkill at the moment but would allow for the perfect solution.

src/services/phases/signedTransactions.ts
```diff
@@ -42,30 +41,13 @@ export async function prepareTransactions(state: OfframpingState, context: Execu
...
-  const spacewalkRedeemTransaction = await prepareSpacewalkRedeemTransaction(state, context);
-  const nablaApproveTransaction = await prepareNablaApproveTransaction(state, context);
-  const nablaSwapTransaction = await prepareNablaSwapTransaction(state, context);
-
-  // Fund Stellar ephemeral only after all other transactions are prepared
-  await stellarCreateEphemeral(stellarEphemeralSecret, outputTokenType);
-  const stellarFundingAccountId = (await fetchSigningServiceAccountId()).stellar.public;
-  const stellarEphemeralKeypair = Keypair.fromSecret(stellarEphemeralSecret);
-  const stellarEphemeralPublicKey = stellarEphemeralKeypair.publicKey();
+  let transactions: any = undefined;
```
ebma, line 44: Let's give this a proper type so we can catch potential bugs more easily.

signer-service/src/index.ts
```diff
@@ -34,6 +34,9 @@ const validateRequiredEnvVars = () => {
 // Validate environment variables before starting the server
 validateRequiredEnvVars();
 
+const eventPoller = new EventPoller(3000);
```
ebma, line 37:
```suggestion
export const eventPoller = new EventPoller(3000);
```

signer-service/src/api/services/brla/hooks.ts
ebma, line 1: How about we rename this to `webhooks.ts`? The other name is confusing due to the connotation with React.js.

signer-service/src/api/services/brla/hooks.ts
```diff
@@ -0,0 +1,71 @@
...
+  private cache: Map<string, Event[]> = new Map();
+  private pollingInterval: number;
+  private apiUrl: string;
+
+  constructor(pollingInterval: number = 1000) {
+    this.apiUrl = WEBHOOKS_CACHE_URL!;
+    this.pollingInterval = pollingInterval;
+  }
+
+  public start() {
```
ebma, line 24: Probably overkill at the moment but we could add another private field that indicates if this is already polling. Accidentally calling `start()` multiple times would create many more useless intervals.

signer-service/src/api/services/brla/hooks.ts
```diff
@@ -0,0 +1,71 @@
...
+      groupedEvents.forEach((fetchedUserEvents, userId) => {
+        const userEvents = this.cache.get(userId) || [];
+        if (userEvents.length > 0) {
+          const lastTimestamp = userEvents[userEvents.length - 1].createdAt;
+          const eventsNotRegistered = fetchedUserEvents.filter(
+            (event) => new Date(event.createdAt) > new Date(lastTimestamp),
+          );
+          userEvents.push(...eventsNotRegistered);
+        } else {
+          userEvents.push(...fetchedUserEvents);
```
ebma, line 55: Wouldn't this make us potentially push duplicate events to `userEvents`?

signer-service/src/api/controllers/brla.controller.ts
```diff
@@ -0,0 +1,164 @@
...
+): Promise<void> => {
+  try {
+    const { taxId, pixId } = req.query;
+
+    if (!taxId || !pixId) {
+      res.status(400).json({ error: 'Missing taxId or pixId query parameters' });
+      return;
+    }
+
+    // TODO how to check that pixId is valid, as a later offramp will get stuck if it's not valid..
```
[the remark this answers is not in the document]
ebma: It's true that we'd want to check the `pixId` but I wouldn't do it as part of this endpoint. Logically, you don't need to specify a `pixId` just to get the user, the `taxId` is enough. And since the `getBrlaUser` endpoint should only care about that, let's move the `taxId` check to somewhere else.

signer-service/src/api/controllers/brla.controller.ts
```diff
@@ -0,0 +1,164 @@
...
+
+    const brlaApiService = BrlaApiService.getInstance();
+    const subaccount = await brlaApiService.getSubaccount(taxId);
+    if (!subaccount) {
+      res.status(404).json({ error: 'Subaccount not found' });
+      return;
+    }
+    if (subaccount.kyc.level !== 1) {
```
ebma, line 26: It can be greater or equal to 1 as far as I understand. 
```suggestion
    if (subaccount.kyc.level < 1) {
```

signer-service/src/api/controllers/brla.controller.ts
```diff
@@ -0,0 +1,164 @@
...
+    return;
+  }
+};
+
+export const createSubaccount = async (
+  req: Request<{}, {}, RegisterSubaccountPayload>,
+  res: Response,
+): Promise<void> => {
+  try {
+    const { cpf: taxId } = req.body;
```
ebma, line 107: Here we again assume that only `cpf` is valid. Sooner or later we would also need to support the other one as we also want to target businesses.

ebma (commented):
Looks good overall, I think the general logic makes sense 👍

src/services/phases/polkadot/xcm/moonbeam.ts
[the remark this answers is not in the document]
gianfra-t: Good point. This is one questionable decision. 

So I tried paying with `BRLA` but no, doesn't work. I didn't try paying with `PEN`, we can certainly use it if i t works. I don't think funding with `GLMR` would be too much of a deal, but getting it into Pendulum on the first place, may be.

About the precise value, I agree that solution would be ideal. Especially because if the fee increases, the xcm will fail and the funds will get stuck. I would argue to do this on a near future improvement though.

src/services/signingService.tsx
```diff
@@ -19,6 +19,39 @@ interface SignerServiceSep10Response {
...
+interface BrlaKycStatus {
+  type: BrlaKycState;
+  status: KycStatus;
+}
+
+type TaxIdType = 'CPF' | 'CNPJ';
+export interface RegisterSubaccountPayload {
+  phone: string;
+  taxIdType: TaxIdType;
+  address: {
```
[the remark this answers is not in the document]
gianfra-t: I haven't been using the `complement` field, what is it? But yeah we can extract the address into a new type.

src/hooks/offramp/useSubmitOfframp.ts
```diff
@@ -30,103 +28,131 @@ export const useSubmitOfframp = () => {
...
-            setOfframpInitiating(false);
+          if (executionInput.outputTokenType === 'brl') {
+            const { taxId, pixId } = executionInput;
+            if (!taxId || !pixId) {
+              setOfframpStarted(false);
+              setOfframpInitiating(false);
+              return;
+            }
+
+            const response = await fetch(`${SIGNING_SERVICE_URL}/v1/brla/getUser?taxId=${taxId}&pixId=${pixId}`);
```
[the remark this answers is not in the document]
gianfra-t: Not anymore, as we will add a `pixId` validation into a separate endpoint.

src/services/phases/polkadot/xcm/index.ts
gianfra-t: I see now it can be a confusing naming. Unsigned was supposed to mean that the transaction has not been yet signed. I'll rename.

signer-service/src/api/controllers/brla.controller.ts
gianfra-t: Good catch.

signer-service/src/api/services/brla/hooks.ts
gianfra-t: Mmm.. the idea is not to. Do you think the logic is faulty? The else condition would only be reached on creation of the entry for that particular user, which would store all fetched events.

signer-service/src/api/services/brla/hooks.ts
gianfra-t: Sure. Very simple change. That won't reduce the issue of instantiating it many times, though. But we only will do so during process start anyway.

signer-service/src/api/controllers/brla.controller.ts
gianfra-t: I'm not really sure if once a company subaccount is created, it should be identified by the `cpf` which is still required, or the `cnpj` field.
Let's leave it like this until we can test a company profile, wdyt?

src/constants/tokenConfig.ts
```diff
@@ -359,20 +367,19 @@ export const OUTPUT_TOKEN_CONFIG: Record<OutputTokenType, OutputTokenDetailsSpac
...
       assetIcon: 'brl',
       symbol: 'BRL',
       name: 'Brazilian Real',
     },
     erc20WrapperAddress: '6eRq1yvty6KorGcJ3nKpNYrCBn9FQnzsBhFn4JmAFqWUwpnh',
-    minWithdrawalAmountRaw: '1',
-    maxWithdrawalAmountRaw: '50000000000000000000000000000000',
+    moonbeamErc20Address: '0xfeB25F3fDDad13F82C4d6dbc1481516F62236429',
+    minWithdrawalAmountRaw: '1000000000000000000', // 1 BRL.
+    maxWithdrawalAmountRaw: '10000000000000000000000', // 10,000 BRL. Maximum value for a KYC level 1 user.
```
gianfra-t, line 382: We don't really have a way to go to KYC level 2 now, so I think it's fair to set the limit this way.

src/services/phases/polkadot/xcm/moonbeam.ts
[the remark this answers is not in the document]
gianfra-t: @ebma follow up. Tried the same call, but with `PEN` instead of `GLMR`, on chopsticks.
But got an error while sending the extrinsic, something like `MinXcmNotDefined`.

I can't find an example of us paying the fee with `PEN`, for Moonbeam, all of our transfers [seem](https://pendulum.subscan.io/extrinsic/4553042-2?event=4553042-7) to pay with GLRM.

src/services/phases/signedTransactions.ts
```diff
@@ -42,10 +46,85 @@ export async function prepareTransactions(state: OfframpingState, context: Execu
...
+      const data = {
+        ...dataCommon,
+        nablaApprovalTx: transactions.nablaApproveTransaction,
+        nablaSwapTx: transactions.nablaSwapTransaction,
+        spacewalkRedeemTx: transactions.spacewalkRedeemTransaction,
+        stellarOfframpTx: transactions.stellarOfframpingTransaction,
+        stellarCleanupTx: transactions.stellarCleanupTransaction,
+        stellarEphemeralPublicKey,
+      };
+      await storeDataInBackend(data);
```
gianfra-t, line 81: I know can be annoying for local testing, but I think it could be an improvement on production. Checking that the storage of the signed transactions happened correctly needs to happen to ensure a potential recovery.

src/constants/tokenConfig.ts
```diff
@@ -388,6 +395,15 @@ export function getOutputTokenDetailsSpacewalk(outputTokenType: OutputTokenType)
   throw new Error(`Invalid output token type: ${outputTokenType}. Token type is not Stellar.`);
 }
 
+export function getOutputTokenDetailsMoonbeam(outputTokenType: OutputTokenType): OutputTokenDetailsMoonbeam {
```
Sharqiewicz, line 398: It is an awesome piece of code, I like it! 🔥

src/hooks/offramp/useSubmitOfframp.ts
```diff
@@ -30,103 +28,131 @@ export const useSubmitOfframp = () => {
...
+              setOfframpInitiating(false);
+              return;
+            }
+
+            const response = await fetch(`${SIGNING_SERVICE_URL}/v1/brla/getUser?taxId=${taxId}`);
+            if (!response.ok) {
+              // Response can also fail due to invalid KYC. Nevertheless, this should never be the case, as when we create the user we wait for the KYC
+              // to be valid, or retry.
+              if (response.status === 404) {
+                // TODO: Redirect to subaccount creation/KYC flow.
```
Sharqiewicz, line 78: Do we want to leave `// @TODO`?

signer-service/src/api/services/brla/brlaApiService.ts
```diff
@@ -0,0 +1,134 @@
...
+
+  public async triggerOfframp(offrampParams: OfframpPayload): Promise<{ id: string }> {
+    const endpoint = `/pay-out`;
+    const repsonse = await this.sendRequest(endpoint, 'POST', undefined, offrampParams);
+    return repsonse;
+  }
+
+  public async createSubaccount(registerSubaccountPayload: RegisterSubaccountPayload): Promise<{ id: string }> {
+    const endpoint = `/subaccounts`;
+    const repsonse = await this.sendRequest(endpoint, 'POST', undefined, registerSubaccountPayload);
```
Sharqiewicz, line 131: I think we can just 
`return this.sendRequest(endpoint, 'POST', undefined, registerSubaccountPayload);`
it's the same but reduces one line

Sharqiewicz: and there is a typo `repsonse => response`

src/services/offrampingFlow.ts
```diff
@@ -84,11 +89,22 @@ export interface InitiateStateArguments {
...
 
-export interface OfframpingState {
-  sep24Id: string;
+export interface BrlaInitiateStateArguments {
+  inputTokenType: InputTokenType;
+  outputTokenType: OutputTokenType;
+  amountIn: string;
+  amountOut: Big;
+  network: Networks;
+  pendulumNode: { ss58Format: number; api: ApiPromise; decimals: number };
```
Sharqiewicz, line 98: maybe we can use `ApiComponents` as a type here and above. From `polkadotNode.tsx`?

src/services/offrampingFlow.ts
```diff
@@ -101,27 +117,43 @@ export interface OfframpingState {
...
+export interface OfframpingState extends BaseOfframpingState {
+  sep24Id?: string;
+  sepResult?: SepResult;
+  stellarEphemeralSecret?: string;
+  executeSpacewalkNonce?: number;
+  transactions?: BrlaOfframpTransactions | SpacewalkOfframpTransactions;
+  brlaEvmAddress?: string;
+  pixDestination?: string;
+  taxId?: string;
+  moonbeamXcmTransactionHash?: `0x${string}`;
```
Sharqiewicz, line 152: I think we can use Address type from Polkadot

Sharqiewicz:
Please don't merge, I'm reviewing @gianfra-t

src/services/phases/polkadot/ephemeral.tsx
```diff
@@ -296,6 +296,13 @@ export async function subsidizePostSwap(state: OfframpingState, context: Executi
     });
   }
 
+  if (state.outputTokenType === 'brl') {
```
Sharqiewicz, line 299: wdyt about changing it to enum ( the OutputTokenType ) to avoid typos?

Sharqiewicz (changes requested):
The overall impression is quite positive. I’ve reviewed half of the files so far, and I wanted to submit my feedback now, as I’ll need to step away for 2 hours

src/services/phases/polkadot/xcm/assethub.ts
```diff
@@ -0,0 +1,68 @@
...
+
+import { ExecutionContext, OfframpingState } from '../../../offrampingFlow';
+import { waitUntilTrue } from '../../../../helpers/function';
+import { getRawInputBalance } from '../ephemeral';
+import { signAndSubmitXcm } from '../xcm';
+
+export function createAssethubAssetTransfer(assethubApi: ApiPromise, receiverAddress: string, rawAmount: string) {
+  const receiverId = u8aToHex(decodeAddress(receiverAddress));
+
+  const dest = { V2: { parents: 1, interior: { X1: { Parachain: 2094 } } } };
```
Sharqiewicz, line 14: I think full name `destination` is clearer and worth having, WDYT?

src/types/offramp.ts
```diff
@@ -30,3 +35,23 @@ export interface OfframpActions {
   updateOfframpHookStateFromState: (state: OfframpingState | undefined) => void;
   resetOfframpState: () => void;
 }
+
+export function isBrlaOfframpTransactions(tx: any): tx is BrlaOfframpTransactions {
```
Sharqiewicz, line 39: use `unknown` or create a type instead of `any`

src/types/offramp.ts
```diff
@@ -30,3 +35,23 @@ export interface OfframpActions {
...
+  return (
+    tx &&
+    typeof tx.nablaApproveTransaction === 'string' &&
+    typeof tx.nablaSwapTransaction === 'string' &&
+    typeof tx.pendulumToMoonbeamXcmTransaction === 'string'
+  );
+}
+
+export function isSpacewalkOfframpTransactions(tx: any): tx is SpacewalkOfframpTransactions {
```
Sharqiewicz, line 48: use `unknown` or create a type instead of `any`

signer-service/src/api/services/brla/webhooks.ts
```diff
@@ -0,0 +1,76 @@
+import { WEBHOOKS_CACHE_URL } from '../../../constants/constants';
+
+type SubscriptionType = 'BURN' | 'BALANCE-UPDATE' | 'MONEY-TRANSFER' | 'MINT';
+
+export interface Event {
+  userId: string;
+  data: any;
```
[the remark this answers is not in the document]
gianfra-t: I'll change to `unkown` (with the rest of anys). The thing with the data field is that we are not using it too much, and each event type depending on the subscription type has a different structure for data.

src/services/signingService.tsx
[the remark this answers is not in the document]
ebma: I don't know what it would be used for in Brazil but it's included in the address object defined by BRLA [here](https://brla-account-api.readme.io/reference/businesscreatesubaccount).

src/types/offramp.ts
gianfra-t: I'm not sure how it would work without `any`. It becomes a bit verbose with `unknown`.  Any ideas?

src/services/phases/polkadot/ephemeral.tsx
gianfra-t: Yes I thought the same, as it grows. It may involve a bit of code change as we use this the type a lot.

src/services/offrampingFlow.ts
gianfra-t: I think `Address` will be a bit confusing here, as it's a hash even if they share the same hex string representation (which I'm not sure).

Maybe there is a `Hash` type, but it probably has other properties that we cannot build with just the string. I think if we want to replace that type, or rename, it's better to define our own. We use it a lot.

signer-service/src/api/services/brla/hooks.ts
ebma: True, I missed that due to the map, each userId is unique in the iteration. Nevermind then. I agree with @Sharqiewicz that the `poll()` function is a little crowded and hard to read though.

signer-service/src/api/controllers/brla.controller.ts
ebma: Okay, fine with me.

src/constants/tokenConfig.ts
ebma: Agreed.

signer-service/src/api/routes/v1/brla.route.ts
```diff
@@ -0,0 +1,17 @@
+import { Router } from 'express';
+import * as brlaController from '../../controllers/brla.controller';
+import { validateBrlaTriggerOfframpInput, validataSubaccountCreation } from '../../middlewares/validators';
+
+const router: Router = Router({ mergeParams: true });
```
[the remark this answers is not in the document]
gianfra-t: We use it in all previous routes. Honestly, I just reused it. You think we don't need it here or in any route?

src/services/phases/brla/index.tsx
```diff
@@ -0,0 +1,123 @@
+import Big from 'big.js';
+import { createPublicClient, http } from 'viem';
+import { moonbeam } from 'viem/chains';
+
+import { SIGNING_SERVICE_URL } from '../../../constants/constants';
+import { OfframpingState } from '../../offrampingFlow';
+import { fetchOfframpStatus } from '../../signingService';
+import Erc20ABI from '../../../../mooncontracts/Erc20ABI.json';
```
ebma, line 8: Let's rename this to make it clear that it's the moonbeam specific one. 
```suggestion
import { default as MoonbeamErc20Abi } from '../../../../mooncontracts/Erc20ABI.json';
```

gianfra-t: I think they're the same as any EVM (they should), but okay.

signer-service/src/api/services/brla/webhooks.ts
```diff
@@ -0,0 +1,76 @@
...
+    }
+    this.poll();
+    this.started = true;
+
+    setInterval(() => {
+      this.poll();
+    }, this.pollingInterval);
+  }
+
+  private async poll() {
```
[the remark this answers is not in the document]
gianfra-t: I think we can split the core `poll` function, I'll try. But the thing is we need to compare by the timestamp of each user individually (not a `getLatestTimestampForAllUsers`), because the api we use to cache is a simple backend that returns everything it has stored. 

So for instance, you may have user A who just offramped, and as we poll a new event, the latest timestamp will indeed be greater that the last one stored.
For a user B, we will get an event that we've already stored, therefore the timestamp will not be greater. We could choose a different field to compare, but I thought timestamp was easier in this case.

I understand that the idea you have is to pass a query param, to get only the events we need. But this backend that returns the cached events does not really have that capability, my idea was to move as much logic as possible here such that the code of that proxy-relayer is easily maintainable.

src/services/signingService.tsx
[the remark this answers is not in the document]
gianfra-t: Ah... I missed it. Both our mock and real users were able to be created without it, so I'll add it as optional.

src/services/phases/polkadot/xcm/moonbeam.ts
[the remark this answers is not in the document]
gianfra-t: I don't have the extrinsic on hand, but I was using `transferMulticurrencies` where you specify by our id not the xcm representation. Perhaps that's it? I'll try again but on live, with some small amount.

src/services/phases/brla/index.tsx
ebma: Kind of, yes. But if they are the same, you could also re-use the existing ERC20 abi that we have in `src/contracts/ERC20.ts`. I assumed you tried that and it didn't work. I saw that some fields differ slightly between the one we already have and the new one. If you didn't try it yet, maybe the existing ERC20 works for both Moonbeam and Pendulum.

src/constants/tokenConfig.ts
ebma: Looking at this again and with the knowledge we have about the KYC levels now, I think we would also need to have a check for how much the user already burned. We can get that info from the subaccount response on BRLA. But let's say, the user already offramped 5,000 BRL once and has KYC level 1. They are only allowed to offramp a total of 10,000 BRL ever (before upgrading to level 2). So if the user clicks on the 'Confirm' button, we should check that _if the user already exists_ the amount the user offramped previously + the current amount are still less or equal to the overall limit. 
This is nothing we need to do in this PR but I just wanted to share it because it came to my mind.

src/services/phases/brla/index.tsx
gianfra-t: I missed that one! I'll try to replace it.

src/constants/tokenConfig.ts
gianfra-t: Interesting, I kept thinking it was a maximum per offramp. Let's do it in an improvement PR then.

signer-service/src/api/services/brla/webhooks.ts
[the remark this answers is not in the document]
Sharqiewicz: I think we should separate this piece of code:
```
groupedEvents.forEach((fetchedUserEvents, userId) => {
        const userEvents = this.cache.get(userId) || [];
        if (userEvents.length > 0) {
          const lastTimestamp = userEvents[userEvents.length - 1].createdAt;
          const eventsNotRegistered = fetchedUserEvents.filter(
            (event) => new Date(event.createdAt) > new Date(lastTimestamp),
          );
          userEvents.push(...eventsNotRegistered);
        } else {
          userEvents.push(...fetchedUserEvents);
        }
        this.cache.set(userId, userEvents);
      });
      ```

File: signer-service/src/api/services/brla/webhooks.ts [hunk 31]
```diff
@@ -0,0 +1,132 @@
+import { WEBHOOKS_CACHE_URL, WEBHOOKS_CACHE_PASSWORD } from '../../../constants/constants';
+import { BrlaApiService } from './brlaApiService';
+
+type SubscriptionType = 'BURN' | 'BALANCE-UPDATE' | 'MONEY-TRANSFER' | 'MINT';
```
ebma, line 4: Aren't we missing some types here? In the examples [here](https://brla-account-api.readme.io/reference/events), I can see there is also "KYC", "REPOST-TRANSACTION", "SWAP", "PIX-TO-USD", "PIX-TO-TOKEN", "USD-TO-PIX".

We don't need to care about many of them, but at least the "KYC" type should be included here 😅

ebma (approved):
The polling function is way easier to read now. Albeit not perfect, let's get this out of the way and merge so we can proceed with the UI connection faster.

signer-service/src/api/services/brla/webhooks.ts
gianfra-t: I definitely missed that one! For the others, yes I deliberately left them to avoid confusion for now.

PR #491, opened 2025-02-27
Fetch events from BRLA API instead of webhooks cache.

gianfra-t:
### Context
Replace polling of webhooks from the cache service, and fetch directly the events from BRLA API. 

Now, each time a user tries to fetch a status, we first query the last events from BRLA and return the last one, if relevant to the user's query.

### Additional changes
Small fix for bug that prevents output token selection.

### DIFF
--- a/signer-service/src/api/controllers/brla.controller.ts
+++ b/signer-service/src/api/controllers/brla.controller.ts
@@ -76,7 +76,7 @@ export const getOfframpStatus = async (req: Request<{}, {}, {}, { taxId: string 
       return;
     }
 
-    const lastEventCached = eventPoller.getLatestEventForUser(subaccount.id);
+    const lastEventCached = await eventPoller.getLatestEventForUser(subaccount.id);
 
     if (!lastEventCached) {
       res.status(404).json({ error: `No status events found for ${taxId}` });
@@ -140,7 +140,7 @@ export const fetchSubaccountKycStatus = async (
       return;
     }
 
-    const lastEventCached = eventPoller.getLatestEventForUser(subaccount.id);
+    const lastEventCached = await eventPoller.getLatestEventForUser(subaccount.id);
 
     if (!lastEventCached) {
       res.status(404).json({ error: `No status events found for ${taxId}` });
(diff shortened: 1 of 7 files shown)


gianfra-t:
@pendulum-chain/devs I left the necessary code for polling the cache as an alternative, as we don't yet know how quickly will BRLA's API have the events ready vs getting them from webhooks.

signer-service/src/api/controllers/brla.controller.ts
```diff
@@ -140,13 +149,18 @@ export const fetchSubaccountKycStatus = async (
       return;
     }
 
-    const lastEventCached = eventPoller.getLatestEventForUser(subaccount.id);
+    const lastEventCached = await eventPoller.getLatestEventForUser(subaccount.id);
```
ebma, line 152: This function now takes the assumption that the KYC event is the last one emitted (by timestamp). Shall we relax this assumption and instead of relying on the fact that the KYC event has to be the latest event, we explicitly check if there was an event with `subscription: "KYC"` and 
```
"data": {
        "kycStatus": "APPROVED",
        "level": 1, // or 2
        "status": "SUCCESS",
...
      },
```
among the latest events? We could for example add a new function or rewrite the existing function to `getSubscriptionEventsForUser(subaccountId, subscriptionType)` and then have access to all the events of a given type.

ebma (commented):
Looks good overall 👍  I just left some nitpicky comments which are not necessarily needed. Though the refactorings might come in handy in the future.

signer-service/src/api/controllers/brla.controller.ts
gianfra-t: True this is a bit of a fundamental limit of this simplistic approach. The thing is, I just don't see how different events would interleave at this stage.

During the creation of the account, it's impossible to get an offramp event. The other way around when the use was KYC'ed and is off/on-ramping.
But most importantly, we control which operation starts so we know when to expect one or the other.

That said, we can add that helper function since is just adding a filter.

ebma: Alright, fine with me.

Issue #484, opened 2025-02-19
Bug: Double signing when offramping from assethub.


gianfra-t:
An Assethub offramp was attempted, yet the application apparently kept reloading at the signing phase. This triggered more than one xcm transfer signing, which the user did, yet only one flow continued and was successful. 

We must investigate what could cause an error on this stage, that would trigger a double signing while potentially discarding one of the flows. 

### Additional info
Both cross chain transfers were done to the same ephemeral account, subscan link [here](https://pendulum.subscan.io/account/153GQ6MRcZiSwRy7xEy3X8kQybdFMov9TWvQr8NDmJLzoFan?tab=xcm_transfer).

PR #492, opened 2025-02-28
Failsafe when sending signed XCM transaction.

gianfra-t:
Issue #484.

The current theory for why this error happened, is that connection with Assethub's RPC dropped between sending the transaction and receiving the confirmation event, as it's hard to see another reason why this function could have failed, knowing the actual extrinsic did not.

Given that, we introduce a new error type `TransactionInclusionError` for the case where the transaction is included in a block, yet for some reason `signAndSend` fails.

In this scenario, we will fetch all the events for the block _once it is available_, and attempt to find the relevant `XcmSent` event.


src/services/phases/polkadot/xcm/index.ts
```diff
@@ -45,11 +63,69 @@ export const signAndSubmitXcm = async (
...
+  const apiAt = await api.at(blockHash);
+  const events = await apiAt.query.system.events();
+  const xcmSentEvents = events.filter((record) => {
+    return record.event.section === 'polkadotXcm' && record.event.method === 'Sent';
+  });
+
+  const event = xcmSentEvents
+    .map((event) => parseEventXcmSent(event))
+    .filter((event) => {
+      return event.originAddress == address;
```
ebma, line 119: How confident are we that this comparison works as expected and we don't need to convert between different encodings of `originAddress` and `address`?

ebma (approved):
The code looks good and makes sense to me 👍  I assume you are basing this on assumptions and we are not really sure if it will work, e.g. it could be that the callback we pass to `signAndSend()` doesn't even trigger after we encounter the error and the `inBlockHash` would never assign. 

I'm fine if we merge this based on the assumption that it works. We'd only need to make sure that the happy path also still works as expected. Have you tested an AssetHub offramp with these changes already @gianfra-t?

src/services/phases/polkadot/xcm/index.ts
gianfra-t: I tested this (although it's quite hard to do so) working with a mocked chain. We can be "fairly" sure because it is the same filter condition we use with `signAndSend`, and we are parsing the event with the same function (which already converts the address).

gianfra-t:
>  I assume you are basing this on assumptions and we are not really sure if it will work
Exactly, the bug observed was quite surprising and I cannot easily pinpoint why it happened.

> Have you tested an AssetHub offramp with these changes already
I believe not, certainly not outside local. Will do.

gianfra-t:
Tested a successful offramp on the preview.

ebma:
Great, then let's get this merged ✅

src/services/phases/polkadot/xcm/assethub.ts
```diff
@@ -62,9 +62,22 @@ export async function executeAssetHubToPendulumXCM(
       context.setOfframpSigningPhase('started');
 
       const afterSignCallback = () => setOfframpSigningPhase?.('finished');
-      const { hash } = await signAndSubmitXcm(walletAccount, tx, afterSignCallback);
-
-      return { ...state, assetHubXcmTransactionHash: hash as `0x${string}` };
+      try {
+        const { hash } = await signAndSubmitXcm(walletAccount, tx, afterSignCallback);
+        return { ...state, assetHubXcmTransactionHash: hash as `0x${string}` };
```
Sharqiewicz, line 67: Could we use `Address` or `Hash` type instead of `0x${string}`?

src/services/phases/polkadot/xcm/assethub.ts
```diff
@@ -62,9 +62,22 @@ export async function executeAssetHubToPendulumXCM(
...
+      } catch (error) {
+        if (error instanceof TransactionInclusionError) {
+          try {
+            const { hash } = await verifyXcmSentEvent(assetHubNode.api, error.blockHash, walletAccount.address);
+            return { ...state, assetHubXcmTransactionHash: hash as `0x${string}` };
```
Sharqiewicz, line 72: Could we use `Address` or `Hash` type instead of `0x${string}`?

src/services/phases/polkadot/xcm/index.ts
```diff
@@ -45,11 +63,69 @@ export const signAndSubmitXcm = async (
...
+            ),
+          );
+        }
+
         reject(new Error(`Failed to do XCM transfer: ${error}`));
       });
   });
 };
 
+async function waitForBlock(api: ApiPromise, blockHash: string, timeoutMs = 60000): Promise<SignedBlock> {
```
Sharqiewicz, line 80: Maybe we can use the example from the docs? https://polkadot.js.org/docs/api/examples/promise/listen-to-blocks/

src/services/phases/polkadot/xcm/index.ts
```diff
@@ -45,11 +63,69 @@ export const signAndSubmitXcm = async (
...
+      }
+    } catch (error) {
+      console.log(error);
+    }
+    await new Promise((resolve) => setTimeout(resolve, pollIntervalMs));
+  }
+  throw new Error(`Block ${blockHash} not found after ${timeoutMs}ms`);
+}
+
+export async function verifyXcmSentEvent(
```
Sharqiewicz, line 99: WDYT?
```suggestion
export async function verifyXcmSentEvent(
  api: ApiPromise,
  blockHash: string,
  address: string,
): Promise<{ event: XcmSentEvent; hash: string }> {
  try {
    await waitForBlock(api, blockHash);
    
    const apiAt = await api.at(blockHash);
    const events = await apiAt.query.system.events();
    
    const xcmSentEvent = events
      .filter(record => 
        record.event.section === 'polkadotXcm' && 
        record.event.method === 'Sent'
      )
      .map(parseEventXcmSent)
      .find(event => event.originAddress === address);

    if (!xcmSentEvent) {
      throw new Error(`No XcmSent event found for account ${address}`);
    }

    return { event: xcmSentEvent, hash: blockHash };
  } catch (err) {
    throw new Error(`Block ${blockHash} not found`);
  }
}
```

src/services/phases/polkadot/xcm/assethub.ts
gianfra-t: I think we wanted to add our own type Address before. But, I'm not sure how to use those types from polkadot.js, is there a conversion function or something?

gianfra-t: For a later refactor, we could add our own classes for Address, to ease out also conversion between generic, hex, etc.

src/services/phases/polkadot/xcm/index.ts
gianfra-t: Changed [here](https://github.com/pendulum-chain/vortex/pull/492/commits/b7df5118d6d728cf7e0fd57a1b3f4acc989302cd). I kept the separate error for block not found though.

src/services/phases/polkadot/xcm/index.ts
gianfra-t: Thanks. I [changed](https://github.com/pendulum-chain/vortex/pull/492/commits/b7df5118d6d728cf7e0fd57a1b3f4acc989302cd) it to a recursive call. However we still do nothing if `getBlock` fails and we also want to keep trying even if it is successful, but not found (in case the block is being built), so I restructured a bit the suggestion.

ebma: Let's not use recursion. I liked your original version better @gianfra-t. The reason is that a) recursion generally has worse performance than iterative functions, especially for larger call depth and b) the signature of the function made more sense before, ie. it's more natural to pass a timeout to a function that waits for a block than a number of retries IMO. I wouldn't know how long the `waitForBlock` function waits in total if I can only specify the number of retries but not the retry delay. Using the timeout simplifies that altogether.

Sharqiewicz (approved):

src/services/phases/polkadot/xcm/index.ts
gianfra-t: Sure, we can go back. I didn't think about affecting performance given that the stack would not be so large.

PR #493, opened 2025-03-03
Create UI for BRLa offramp KYC

Sharqiewicz:
💚 Ready 
---

Implement KYC Form (and additional fields to Swap Form) for BRLA.

## Changes:
Updates
- [x] update react-hook-form
---
Full KYC
- [x] create `BrlaExtendedForm` component to manage logic of showing KYC Form/Loading/Error/Success
- [x] create `brla/KYCForm` component for BRLA full KYC form
- [x] create `brla/useKYCForm` for handling BrlaExtendedForm logic
- [x] create `VerificationStatus` component to handle KYCStatus
- [x] create `useBRLAKYCProcess` hook for handling BRLA KYC submit logic 
- [x] create `useKYCStatusQuery` hook for handling BRLA KYC request logic
---
Components
- [x] create `BrlaField` component for handling Brla-related (react-hook-form) fields/inputs
- [x] create `Field` component for handling (react-hook-form) input 
---
Swap
- [x] create `BrlaSwapFields` which are injected to the Swap form when `BRLA` is chosen
- [x] separate `Swap` component from the `SwapPage` 
- [x] create `brla/useOfframpSubmission`  for BRLA Offramp submisssion


src/components/TermsAndConditions/index.tsx
```diff
@@ -35,6 +37,7 @@ const TermsAndConditionsContent = ({
         className="checkbox checkbox-primary checkbox-sm"
         checked={termsChecked}
         onChange={() => {
+          setTermsAccepted(true);
```
Sharqiewicz, line 40: ?

Sharqiewicz:
@pendulum-chain/devs Ready for review 💚

src/pages/swap/index.tsx
```diff
@@ -436,132 +389,114 @@ export const SwapPage = () => {
...
-                fromToken,
-                toTokenSymbol: toToken.fiat.symbol,
-              }}
-            />
-          }
+      {offrampKycStarted ? (
+        <PIXKYCForm
+          feeComparisonRef={feeComparisonRef}
+          setIsOfframpSummaryDialogVisible={setIsOfframpSummaryDialogVisible}
+          onSwapConfirm={onSwapConfirm}
```
gianfra-t, line 480: We're not using `onSwapConfirm` right? It will directly call ` const proceedWithOfframp = useOfframpSubmission` after the KYC passes.

src/components/TermsAndConditions/index.tsx
ebma: I think it's better if we don't immediately hide the terms and conditions as soon as the user clicks on the checkbox but keep them visible until they click on the 'Confirm' button. Otherwise they might hastily toggle the checkbox and might wonder what exactly they just toggled and is now gone.

src/hooks/brla/useBRLAKYCProcess/index.tsx
```diff
@@ -0,0 +1,141 @@
...
+        resetToDefault();
+        handleBackClick();
+      });
+    },
+    [handleBackClick, updateStatus, resetToDefault],
+  );
+
+  const proceedWithOfframp = useOfframpSubmission(handleError, setIsOfframpSummaryDialogVisible);
+
+  const handleFormSubmit = useCallback(
```
ebma, line 84: This function is missing a call to the `createSubaccount()` endpoint of our signer service with the relevant fields. Only then, can we query the KYC status of the user and expect to get a meaningful response.

src/hooks/brla/useKYCStatusQuery/index.tsx
```diff
@@ -0,0 +1,23 @@
...
+      if (!cpf) throw new Error('CPF is required');
+      return fetchKycStatus(cpf);
+    },
+    enabled: !!cpf,
+    refetchInterval: (query) => {
+      if (!query.state.data || query.state.data.status === 'PENDING') return POLLING_INTERVAL_MS;
+      return false;
+    },
+    retry: 3,
+    retryDelay: (attemptIndex) => Math.min(1000 * 2 ** attemptIndex, 30000),
```
ebma, line 20: I think using a constant delay would be fine. Let's choose a delay of 5 seconds and up to 5 retries. I'm not sure how long the BRLA api takes to approve the KYC data but it might take some time.

src/hooks/offramp/useSubmitOfframp.ts
```diff
@@ -76,34 +77,23 @@ export const useSubmitOfframp = () => {
...
+                console.log('status 404 on brla user');
+                setOfframpKycStarted(true);
                 return;
               }
               throw new Error('Error while fetching funding account signature');
             }
             const { evmAddress: brlaEvmAddress } = await response.json();
-            const brlaOfframpExecution = { ...executionInput, brlaEvmAddress };
-            setOfframpExecutionInput(brlaOfframpExecution);
+            //const brlaEvmAddress = '0x7Ba99e99Bc669B3508AFf9CC0A898E869459F877'
```
ebma, line 87: Nit: remove comment

ebma (commented):
I skimmed through the code. Most of the changes look reasonable to me. Also the look and feel and animations are really nice, great job with that @Sharqiewicz 👌 
I tested this locally a bit and found that we are missing the call to create the subaccount when all the data is filled in on the KYC form. We should add that.

src/hooks/brla/useBRLAKYCProcess/index.tsx
```diff
@@ -97,21 +121,24 @@ export function useKYCProcess(setIsOfframpSummaryDialogVisible: (isVisible: bool
...
         [KYCResponseStatus.FAILED]: async () => {
           updateStatus(KYCStatus.FAILED, STATUS_MESSAGES.FAILED);
           await delay(ERROR_DISPLAY_DURATION_MS);
+          setIsSubmitted(false);
           resetToDefault();
           handleBackClick();
         },
-        [KYCResponseStatus.PENDING]: async () => undefined,
+        [KYCResponseStatus.PENDING]: async () => {
+          undefined;
```
Sharqiewicz, line 140: ?

signer-service/src/api/controllers/brla.controller.ts
```diff
@@ -51,17 +51,32 @@ export const getBrlaUser = async (req: Request<{}, {}, {}, { taxId: string }>, r
...
-    console.log('Triggering offramp. Amount ', amount, 'taxId ', taxId, 'pixKey ', pixKey);
     const brlaApiService = BrlaApiService.getInstance();
     const subaccount = await brlaApiService.getSubaccount(taxId);
 
     if (!subaccount) {
       res.status(404).json({ error: 'Subaccount not found' });
       return;
     }
+    const receiverTaxIdClean = receiverTaxId.replace(/\.|-/g, '');
+    const pixKeyClean = pixKey.replace(/\.|-/g, '');
```
ebma, line 62: Maybe we need to ask the BRL people about this but as far as I can tell from [this](https://developers.international.pagseguro.com/reference/pix-key-validation-and-regex-1) documentation, it wouldn't be a good idea to always remove dots and dashes. The example for the email address contains dots and the example for the random key contains dashes. Both should be considered valid.

signer-service/src/api/services/pendulum/pendulum.service.ts
```diff
@@ -25,14 +26,29 @@ export function getFundingData(
...
+      const pendulumCurrencyId = { XCM: 13 };
+
+      const penFundingTx = apiData.api.tx.balances.transferKeepAlive(ephemeralAddress, fundingAmountRaw);
+      const glmrFundingTx = apiData.api.tx.tokens.transfer(
+        ephemeralAddress,
+        pendulumCurrencyId,
+        GLMR_FUNDING_AMOUNT_RAW,
+      );
+
+      const batch = apiData.api.tx.utility.batch([penFundingTx, glmrFundingTx]);
```
ebma, line 45: Nit: `batchAll` will make sure that either both extrinsics work or none. Makes more sense to use this one here as it would be unfortunate if only one works for some reason. Then the recipient account would still hold some tokens that we need to recover (or not). 
```suggestion
      const batch = apiData.api.tx.utility.batchAll([penFundingTx, glmrFundingTx]);
```

signer-service/src/api/controllers/brla.controller.ts
gianfra-t: Yeah I know. But I wasn't sure if that was issue we were facing, and also wanted to normalize to always use one format that we know works.

signer-service/src/api/services/pendulum/pendulum.service.ts
gianfra-t: Good call.

signer-service/src/api/controllers/brla.controller.ts
Sharqiewicz: I think we should use `replaceAll` as `replace` only affects the first match

signer-service/src/constants/constants.ts
```diff
@@ -2,9 +2,10 @@ const HORIZON_URL = 'https://horizon.stellar.org';
 const PENDULUM_WSS = 'wss://rpc-pendulum.prd.pendulumchain.tech';
 const NETWORK = 'Pendulum';
 const PENDULUM_FUNDING_AMOUNT_UNITS = '10'; // 10 PEN. Minimum balance of funding account
+const PENDULUM_GLMR_FUNDING_AMOUNT_UNITS = '0.5'; // 10 GLMR. Minimum balance of funding account
 const STELLAR_FUNDING_AMOUNT_UNITS = '10'; // 10 XLM.  Minimum balance of funding account
 const MOONBEAM_FUNDING_AMOUNT_UNITS = '10'; // 10 GLMR. Minimum balance of funding account
-const SUBSIDY_MINIMUM_RATIO_FUND_UNITS = '5'; // 10 Subsidies considering maximum subsidy amount use on each (worst case scenario)
+const SUBSIDY_MINIMUM_RATIO_FUND_UNITS = '0.01'; // 10 Subsidies considering maximum subsidy amount use on each (worst case scenario)
```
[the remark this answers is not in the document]
gianfra-t: Yes it was mostly for testing BRLA, as we don't have much on the funding.

signer-service/src/api/controllers/brla.controller.ts
```diff
@@ -3,6 +3,58 @@ import { BrlaApiService } from '../services/brla/brlaApiService';
 import { RegisterSubaccountPayload, TriggerOfframpRequest } from '../services/brla/types';
 import { eventPoller } from '../..';
 
+// BRLA API requires the date in the format YYYY-MMM-DD
+function convertDateToBRLAFormat(dateNumber: number) {
```
Sharqiewicz, line 7:
```suggestion
function convertDateToBRLAFormat(dateNumber: number) {
  const date = new Date(dateNumber);
  return date.toISOString().slice(0, 10); // Returns YYYY-MM-DD
}
```

ebma: The format is `YYYY-MMM-DD` though, e.g. 2025-Jan-01.

signer-service/src/api/controllers/brla.controller.ts
```diff
@@ -47,16 +95,25 @@ export const triggerBrlaOfframp = async (req: Request<{}, {}, TriggerOfframpRequ
     }
 
     const subaccountId = subaccount.id;
-    const { id: offrampId } = await brlaApiService.triggerOfframp({ subaccountId, pixKey, amount, receiverTaxId });
+    console.log(
```
Sharqiewicz, line 98: Do we want to leave this console.log?

gianfra-t (approved):

Issue #489, opened 2025-02-24
Change format/styling of quote comparison


ebma:
## Context

The quote/fee comparison table could use some more love. The table should
- Format numbers, so that especially large ones are better readable 
- Mark winner with green rectangle around it, in the corner a little space with same color background saying "Best rate" 

## TODO

- Make the text in the table right-aligned so that the numbers are shown aligned at the decimal dot. This means, the amount that the user pays extra should look similar to this. 
```
  947.11 EUR
  - 5.45 EUR
```

- Also, add the `,` delimiters for large numbers at every third digit, e.g. the amount 1000000.45 ARS should be 1,000,000.45 ARS.
- Add the green rectangle around to the row with the winner (best quote) and add a legend saying "Best rate" in the same green color

ebma:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Pendulum-6408a0365cda59c57d39bee5/issues/gh/pendulum-chain/vortex/489?planning-poker) with Zenhub @Sharqiewicz @gianfra-t

PR #494, opened 2025-03-03
489 change formatstyling of quote comparison

gianfra-t:
Note: Please, merge first #491 first.

Issue: #489.

This PR changes the formatting of the price provided by competitors, as well as a green highlight for the Vortex entry when the comparison is better.

### DIFF
--- a/src/components/FeeComparison/index.tsx
+++ b/src/components/FeeComparison/index.tsx
@@ -1,5 +1,5 @@
 import Big from 'big.js';
-import { useEffect, useRef, useMemo, useImperativeHandle } from 'react';
+import { useEffect, useRef, useMemo, useImperativeHandle, useCallback, useState } from 'react';
 import { forwardRef } from 'react';
 import { useQuery } from '@tanstack/react-query';
 import { ChevronDownIcon } from '@heroicons/react/20/solid';
@@ -21,6 +21,15 @@ interface BaseComparisonProps {
 
 type VortexRowProps = Pick<BaseComparisonProps, 'targetAssetSymbol' | 'vortexPrice'>;
 
+export function formatPrice(price: Big | null | undefined): string {
+  if (!price) return '0.00';
+
+  return new Intl.NumberFormat('en-US', {
+    minimumFractionDigits: 2,
+    maximumFractionDigits: 2,
+  }).format(parseFloat(price.toFixed(2)));
+}
+
 function VortexRow({ targetAssetSymbol, vortexPrice }: VortexRowProps) {
   return (
     <div className="flex items-center justify-between w-full">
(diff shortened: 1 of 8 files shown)


gianfra-t:
@pendulum-chain/devs since we can't get the query using the preview, here are a couple screenshots, both a positive and negative comparison:
image
image

ebma:
I didn't check the code yet but first want to clarify. If a provider other than Vortex has the best rate, we should be fair and also give them the "best rate"-rectangle. This doesn't seem to be the case on your second screenshot, or is it?

gianfra-t:
> This doesn't seem to be the case on your second screenshot, or is it?

Exactly... I assumed we would aim to always be lower.

ebma:
Haha yes of course. Ideally, it's only shown for Vortex 🤡 IIRC wise was also highlighting the other competitors if someone else had in fact the best rate.

src/components/FeeComparison/index.tsx
```diff
@@ -21,6 +21,15 @@ interface BaseComparisonProps {
 
 type VortexRowProps = Pick<BaseComparisonProps, 'targetAssetSymbol' | 'vortexPrice'>;
 
+export function formatPrice(price: Big | null | undefined): string {
+  if (!price) return '0.00';
+
+  return new Intl.NumberFormat('en-US', {
```
Sharqiewicz, line 27: @gianfra-t Maybe we should detect the navigator language and display the format adjusted to the country?
I wrote an article about this some time ago: https://sharqiewicz.medium.com/how-to-format-a-currency-web3-frontend-aadd691bcc16

Sharqiewicz: @ebma wdyt?

gianfra-t: Nice! Will use your solution 🥇 .

ebma: Sure, let's do that 👍

gianfra-t:
@pendulum-chain/devs any provider will be highlighted now, if the rate is better.

ebma:
One minor thing that I just realized which could be improved is that instead of using Vortex as the baseline it might make more sense to use the best rate as the baseline and then show the deduction on the others. Taking the example on the screenshot, instead of showing '0.60 EUR' on AlchemyPay, we could show '-0.60 EUR' on Vortex. 

I'm just sharing some thoughts here, we don't need to do it here or ever. 

image

ebma (approved):
Looks good to me 👍

gianfra-t:
I think you're right. But that will require a bit more of a refactor probably. We clearly thought initially that the Vortex would be different, but if we do like that, where the baseline is the best price, and all display the discounted amount, then we should treat Vortex as another provider.

PR #501, opened 2025-03-10
Support Safe Wallet

Sharqiewicz:
PR is READY 🟢
---
Refactored `isTransactionSafeWallet`
Previously false negatives were produced when we queried the blockchain node and the node hadn't managed to index the transaction hash yet and threw an error indicating it was a safe wallet transaction but it was a blockchain transaction.

Now the `isTransactionSafeWallet`
1. Checks if the transaction is indexed in the blockchain
2.a if is found in the blockchain -> returns `false`
2.b if is not found in the blockchain
3. Query the Safe Wallet API in order to find Safe Wallet transaction
4. If a safe wallet transaction is found -> returns `true`
5. If safe wallet transaction is not found -> delay 1 second and recursively call the `isTransactionSafeWallet` function

### DIFF
--- a/src/hooks/offramp/useSEP24/useAnchorWindowHandler.ts
+++ b/src/hooks/offramp/useSEP24/useAnchorWindowHandler.ts
@@ -13,6 +13,7 @@ import { usePendulumNode } from '../../../contexts/polkadotNode';
 import { useOfframpActions, useOfframpExecutionInput } from '../../../stores/offrampStore';
 import { useSep24Actions, useSep24InitialResponse, useSep24AnchorSessionParams } from '../../../stores/sep24Store';
 import { useVortexAccount } from '../../useVortexAccount';
+import { useChainId } from 'wagmi';
 
 const handleAmountMismatch = (setOfframpingStarted: (started: boolean) => void): void => {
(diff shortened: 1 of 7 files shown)


gianfra-t (approved):
Looks good! I didn't test an offramp, but if it works then let's go.

src/hooks/offramp/useSEP24/useAnchorWindowHandler.ts
```diff
@@ -30,6 +31,7 @@ export const useAnchorWindowHandler = () => {
   const { apiComponents: pendulumNode } = usePendulumNode();
   const { setOfframpStarted, updateOfframpHookStateFromState } = useOfframpActions();
   const { address } = useVortexAccount();
+  const selectedNetworkId = useChainId();
```
ebma, line 34: Let's use the `chainId` exposed by our `useVortexAccount()`. If we use the chain ID from wagmi directly, we will face edge cases with AssetHub again.

Sharqiewicz: Very very very good catch @ebma thank you!

ebma (approved):

PR #502, opened 2025-03-10
Fix price comparison table

gianfra-t:
### Changes

- Adds a fix that would prevent re-calculating the best rate when the provider can't serve the quote.
- Refactor the table such that the best provider is on top, and all others (including Vortex) show the difference in rate based on the best one.


Sharqiewicz (approved):
I refactored the `FeeComparison` component - it was definitely too long and too complicated. Now it's divided into three files: `FeeComparisonTable` , `FeeProviderRow` ,  `FeeComparison`

Sharqiewicz:
@pendulum-chain/devs Please review ❤️

ebma (approved):
I like the new structure, it's definitely an improvement. However, I think it's sad that clicking the 'Compare fee' button now makes the view jump to the table instead of smooth-scrolling it into view. That's why we had the custom ref before. Any chance we can have the smooth scroll behavior back without much effort @Sharqiewicz?

ebma:
Thanks! Ready for merge then IMO

PR #503, opened 2025-03-10
restructure safe tx hash check

gianfra-t:


src/services/phases/squidrouter/process.ts
```diff
@@ -147,6 +147,8 @@ export async function squidRouter(
       phase: 'pendulumFundEphemeral',
     };
   } catch (error) {
+    storageService.remove(storageKeys.SQUIDROUTER_RECOVERY_STATE_APPROVAL);
+    storageService.remove(storageKeys.SQUIDROUTER_RECOVERY_STATE_SWAP);
```
Sharqiewicz, line 151: Do we want to remove it when there is an error?

gianfra-t: I think yes. Previously, due to the error we faced with safe, I was not getting requested to sign on a re-attempt.

PR #506, opened 2025-03-11
Fixes for Pendulum to Moonbeam XCM phase.

gianfra-t:
### Changes

- Fix destination interior order, causing an invalid xcm reception on the Moonbeam side
- Fixes event listening when sending XCM from pendulum.
- Nonce issues.

### DIFF
--- a/src/services/phases/polkadot/xcm/moonbeam.ts
+++ b/src/services/phases/polkadot/xcm/moonbeam.ts
@@ -1,10 +1,10 @@
 import { ApiPromise, Keyring } from '@polkadot/api';
 
-import { ExecutionContext, OfframpingState } from '../../../offrampingFlow';
+import { BrlaOfframpTransactions, ExecutionContext, OfframpingState } from '../../../offrampingFlow';
 
-import { submitXcm } from '.';
+import { submitXTokens } from '.';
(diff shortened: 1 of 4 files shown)


src/services/phases/polkadot/xcm/moonbeam.ts
```diff
@@ -60,8 +60,24 @@ export async function executePendulumToMoonbeamXCM(
...
+      await createPendulumToMoonbeamTransfer(
+        pendulumNode,
+        state.brlaEvmAddress!,
+        outputAmount.raw,
+        state.pendulumEphemeralSeed,
+      ),
+    );
+
+    const newTransactions: BrlaOfframpTransactions = { ...transactions, pendulumToMoonbeamXcmTransaction: xcm };
+    return { ...state, transactions: newTransactions, phase: 'performBrlaPayoutOnMoonbeam' };
```
ebma, line 79: Why are we creating a new XCM transaction when the initial submission failed?

gianfra-t: Rolled back so we can think about XCM failure modes later, with more time. And possibly connect it to [this](https://github.com/pendulum-chain/vortex/issues/484).

ebma (approved):

PR #509, opened 2025-03-12
Refactor offramp flow into zustand store with loop

gianfra-t:
Note: This is just a pr to showcase the idea. Do not merge.

### DIFF
--- a/src/hooks/offramp/useOfframpAdvancement.ts
+++ b/src/hooks/offramp/useOfframpAdvancement.ts
@@ -1,73 +1,37 @@
-import { useEffect, useRef } from 'react';
+import { useEffect } from 'react';
 import { useConfig } from 'wagmi';
-
-import { advanceOfframpingState } from '../../services/offrampingFlow';
 
 import { usePolkadotWalletState } from '../../contexts/polkadotWallet';
 import { useAssetHubNode } from '../../contexts/polkadotNode';
 import { usePendulumNode } from '../../contexts/polkadotNode';
 import { useEventsContext } from '../../contexts/events';
 
-import { useOfframpActions, useOfframpState } from '../../stores/offrampStore';
-import { useNetwork } from '../../contexts/network';
-import { isNetworkEVM } from '../../helpers/networks';
-import { storageService } from '../../services/storage/local';
+import { useOfframpActions } from '../../stores/offrampStore';
+import { ExecutionContext } from '../../services/offrampingFlow';
 
+// This hook is now the bridge between the offramp flow loop, and the context
+// that may update during the flow.
 export const useOfframpAdvancement = () => {
-  const { walletAccount } = usePolkadotWalletState();
-  const { trackEvent } = useEventsContext();
+  const { updateFlowContext, setOfframpSigningPhase } = useOfframpActions();
   const wagmiConfig = useConfig();
-
-  const { selectedNetwork } = useNetwork();
+  const { trackEvent } = useEventsContext();
   const { apiComponents: pendulumNode } = usePendulumNode();
   const { apiComponents: assetHubNode } = useAssetHubNode();
-
-  const offrampState = useOfframpState();
-  const { updateOfframpHookStateFromState, setOfframpSigningPhase } = useOfframpActions();
-
-  const isProcessingAdvance = useRef(false);
+  const { walletAccount } = usePolkadotWalletState();
 
   useEffect(() => {
-    if (isNetworkEVM(selectedNetwork) && wagmiConfig.state.status !== 'connected') return;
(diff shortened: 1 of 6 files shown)


src/stores/offrampStore.ts
```diff
@@ -51,6 +50,46 @@ export const useOfframpStore = create<OfframpStore>()((set) => ({
...
+    // Iadvances the state until a final state is reached.
+    // Initial state is defined after prepare transactions, and this is triggered.
+    startFlow: async () => {
+      let currentState = get().offrampState;
+      console.log('Starting offramp flow with initial state:', currentState);
+      if (!currentState) {
+        console.error('No initial state present; cannot start flow.');
+        return;
+      }
+      while (true) {
```
Sharqiewicz, line 72: I don't like the idea of putting infinite while loops into the zustand store, maybe we can extract this logic somewhere?

Sharqiewicz, line 72: I don't like the idea of putting infinite while loops into the zustand store, maybe we can extract this logic somewhere?

src/stores/offrampStore.ts
```diff
@@ -51,6 +50,46 @@ export const useOfframpStore = create<OfframpStore>()((set) => ({
...
+        // Move finality check out of advanceOfframpingState
+        if (currentState!.phase === 'success' || currentState!.failure !== undefined) {
```
Sharqiewicz, line 74: I think we shouldn't use `!` assertions in so many cases

src/stores/offrampStore.ts
```diff
@@ -51,6 +50,46 @@ export const useOfframpStore = create<OfframpStore>()((set) => ({
     },
 
     clearInitializeFailedMessage: () => set({ initializeFailedMessage: undefined }),
+
+    updateFlowContext: (context: any) => {
```
Sharqiewicz, line 54: Please don't use `any`. If the type isn't known we should use `unknown` instead

src/stores/offrampStore.ts
```diff
@@ -1,12 +1,17 @@
 import { create } from 'zustand';
 import { OfframpState, OfframpActions } from '../types/offramp';
-import { clearOfframpingState } from '../services/offrampingFlow';
+import { clearOfframpingState, advanceOfframpingState, ExecutionContext } from '../services/offrampingFlow';
 
+type FlowContext = any;
```
Sharqiewicz, line 5: Please don't use `any`. If the type isn't known we should use `unknown` instead

Sharqiewicz (changes requested):

src/hooks/offramp/useOfframpAdvancement.ts
```diff
@@ -1,73 +1,37 @@
...
+    const context: ExecutionContext = {
+      wagmiConfig,
+      setOfframpSigningPhase,
+      trackEvent,
+      pendulumNode,
+      assetHubNode,
+      walletAccount,
+    };
+    updateFlowContext(context);
+  }, [wagmiConfig, trackEvent, pendulumNode, assetHubNode, walletAccount, updateFlowContext]);
```
Sharqiewicz, line 36: Missing dependencies

src/stores/offrampStore.ts
gianfra-t: Yes I left typing for once we decide if we'll use this or not.

gianfra-t:
> Isn't it the case that in every state transition function in the app such as: prepareTransactions - there is no state parameter passed, but we just retrieve this state from the Zustand Store?

Yes, exactly. But I guess you refer to `const offrampState = useOfframpState();` on the current implementation of `useOfframpAdvancement` which acts as our "main process".

`advanceOfframpingState` takes the `offrampState` as parameter. Since the call logic is now inside the store itself, we can just get it with `get().offrampState;`. I tried not to change much.

Regarding your concern about the loop inside `startFlow`, I agree. In any case, `startFlow` is intended to be called once and there is no case we want 2 flows in parallel. 

This was just to showcase the idea of a centralized state management, in this case using zustand, but may not be the best tool for the job. Do you have any other library/pattern in mind that would suite us better, without too much boilerplate?

Sharqiewicz:
@gianfra-t We should store the offramp logic in Zustand store, but the implementation of the logic should remain exclusively within the hooks

gianfra-t:
Mmmm but that's why we're trying to move it out of the hook, or at least testing the possibility, since we sometimes have two concurrent flows.

Even with the guard that `useOfframpAdvancement` it's very hard to avoid it. 

Alternatively, we can try to keep patching it until we move to the backend.

ebma:
Let's close this as it's not necessary going forward.

PR #527, opened 2025-03-19
Text update

prayagd:
- from Sell crypto to Sell


ebma (approved):
I think just 'Sell' as the title looks a little empty and lonely. But fine with me, if requested.

Sharqiewicz (approved):

PR #529, opened 2025-03-19
Fix issues with Assethub offramp using a walletconnect wallet

ebma:
Seems like the issue was related to different address formats of the same account. The XCM event included a different address than the wallet. 

- [x] Convert addresses to same ss58 format before comparing them in the events
- [x] Bump walletconnect dependencies (because why not)
- [x] Add extra handling for the 1012 error. During testing I noticed that Vortex can get stuck with this error. This is unlikely to happen in production as the page hot reloaded due to code changes (and not because it encountered an error). However, if the user reloads the page with a bad timing, ie. right after the extrinsic was submitted, Vortex could get stuck trying to submit the same extrinsic again and again, leading to the 1012 error ("Transaction temporarily banned"). In this case, we just assume it was submitted and resolve the promise.


src/services/phases/polkadot/xcm/index.ts
```diff
@@ -157,6 +166,11 @@ export const submitXcm = async (
         }
       })
       .catch((error) => {
+        // 1012 means that the extrinsic is temporarily banned and indicates that the extrinsic was already sent
+        if (error?.message.includes('1012:')) {
+          // We assume that the previous submission worked and return the address
```
gianfra-t, line 171: Could also be the case that the previous extrinsic failed, in general we handle the error on the phase function by restarting, etc. 

I would prefer to pass an error, and for the unfortunate reload case, have a condition on the phase level that can check if the submission was already completed. wdyt?

gianfra-t: I think we can work with the expected outcome. For instance, on the assethub -> pendulum phase we check `didInputTokenArrivedOnPendulum`.

So, for `executePendulumToMoonbeamXCM` which is were we use `submitXTokens` we could also in theory check for the tokens on moonbeam before moving forward, if we see a `1012` or any error.

To avoid extra work, you could leave it like this, and only add the outcome check when we actually have more than one transaction stored. Since there is not much we can do if we see this error now.

ebma: I refactored the code a bit and moved the error handling one level higher. I kept the balance check in the `performBrlaPayoutOnMoonbeam` phase. As far as I understand the measure to ban transactions, it only happens if the very same extrinsic was already submitted by the very same account. So the chance for edge cases here should be slim. The only way this could fail is if the XCM transaction itself was faulty. But this will be caught by the balance check in `performBrlaPayoutOnMoonbeam` regardless.

src/services/phases/polkadot/xcm/index.ts
```diff
@@ -197,6 +221,10 @@ export const submitXTokens = async (
         }
       })
       .catch((error) => {
+        // 1012 means that the extrinsic is temporarily banned and indicates that the extrinsic was already sent
+        if (error?.message.includes('1012:')) {
```
Sharqiewicz, line 225: Maybe we can use enum for the error handling?

like

```
enum xyz = {
TEMPORARILY_BANNED = '1012:'
...
}
```

ebma: Hmm 🤔 we could but I think this is kind of a one-off. I wouldn't expect us to encounter any other RPC errors similar to this honestly so the enum is probably overkill for just one type. Let's keep it in mind and extract it to an enum if we happen to add another error code for this in the future?

Sharqiewicz: Yes, I agree

PR #532, opened 2025-03-20
Send custom error logs to sentry

### LINKED ISSUE #515: Send custom error logs to Sentry
## Context

At the moment, the failure reasons that lead Vortex to show the failure page are caught exceptions and thus not tracked by Sentry. To facilitate the troubleshooting, we should send the failure reason to Sentry.

## TODO

- Send a custom log message for any errors encountered during the offramping process to Sentry.
- Try to include the user's wallet address in the log message

ebma:
- [x] Add the user's wallet address to the Sentry's `user` if available
- [x] Log the failure reason when entering the failure page. I decided to add this as part of a `useEffect` to the failure page component instead of adding it somewhere into the flow because IMO it's easy to read this way and easy to catch both failure reasons only once. 

Closes #515.

### DIFF
--- a/src/hooks/useVortexAccount.ts
+++ b/src/hooks/useVortexAccount.ts
@@ -1,9 +1,9 @@
-import { useNetwork } from '../contexts/network';
-import { useMemo, useCallback } from 'react';
-import { usePolkadotWalletState } from '../contexts/polkadotWallet';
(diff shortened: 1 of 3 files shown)


gianfra-t (approved):
Looks good. Did you test it? if not we can test on staging, as I think sentry may not work on the previews.

PR #533, opened 2025-03-20
Add check for BRL ramping amount

### LINKED ISSUE #518: Add check for BRL ramping amount
## Context

Subaccounts for users with KYC level 1 have ramping limits of 10,000 BRL. Ramping more than this is not allowed. The limit resets every 30 days in a rolling window. 

- Used limits of an account can be queried from [this](https://brla-account-api.readme.io/reference/businessusedlimit) endpoint. 
- The overall limits for the KYC level of an account can be queried from [this](https://brla-account-api.readme.io/reference/businessinfo) endpoint. It's included in `kyc` > `limits`.

## TODO

Before kicking off the offramping for BRL users and after checking the CPF/CNPJ: Check the accumulated limits of the respective account. If the entered amount of BRL is above the limit _or_ would exceed the limit with the recently ramped amount, show an error message and don't proceed with the offramp process.

ebma:
- [x] Refactor code for getting subaccount data in the frontend. Move the logic to a separate function and make the checks for missing user or invalid kyc more expressive.
- [x] Add new endpoint for retrieving user limits to the signer service.
- [x] Add logic to query the user limit from the signer service. Note that this endpoint requires the `subaccountId` instead of a `taxId`. I decided to leave this coherent to the BRLA API as we have access to both in the frontend when doing the query.

Closes #518.


ebma:
Superseded by #597.

PR #534, opened 2025-03-22
Add BRL onramp & move core logic from frontend to backend

### LINKED ISSUE #497: Add support for onramping BRL to stablecoins
## Context 
This ticket is to add the logic part of on-ramping BRL to stablecoins. As a result, Brazilian users should be able to pay BRL and receive USDC or USDT on all the networks supported by Vortex, ie. AssetHub _and_ other EVM networks. 

### Onramping flow (summary)

The onramping flow differs from the offraming flow in that the user doesn't need to sign any transactions. All signing happens on ephemeral accounts. 

The flow can be summarized as follows:
(Assumption: User completed KYC)
1) User initiates onramp. They choose the amount they want to pay (in BRL), the target network, target token (USDT or USDC), and destination address. Vortex calculates the expected output amount incorporating potential fees and the Nabla rate.
2) Vortex creates a pay-in request on [BRLA's API](https://brla-account-api.readme.io/reference/businesspayinpixtotoken). The `receiverAddress` can remain empty in order to use the default account assigned to the user. It doesn't matter at this point.  
3) Vortex shows the BR Code of the pay-in request to the user which they scan to transfer tokens from their bank account to the destination. 
4) Once the BRLA tokens are received on Polygon, Vortex calls into the BRLA API to teleport the BRLA to Moonbeam to a **funded** ephemeral account without any additional fee/cost. 
5) On Moonbeam, the ephemeral account sends the BRLA tokens via XCM to a **funded** ephemeral account Pendulum.
6) Vortex swaps the BRLA to either AssetHub-USDC or axlUSDC on the Forex AMM, depending on the target network.
7) Token transfer to target network. 
7a) If the target network is AssetHub, Vortex will finalize the onramp with an XCM transfer to the designated destination address the user chose in step 1. 
7b) If the target network is an EVM network, Vortex first transfers the received axlUSDC to Moonbeam and then creates a squidrouter swap request to the destination chain and account. 

### User KYC 

Users need to be KYC'd before being able to onramp tokens. The endpoints for the KYC are already in place on the signer-service. The assumption is that nothing needs to be added/changed here. 

#### triggerOnramp() (`/pay-in`)
This endpoint receives a `taxId`, `amount` and `referenceLabel`. The `referenceLabel` needs to identify the address of the ephemeral account on Moonbeam. This is important to ensure that the teleporting of the assets can not be abused later by an adversary teleporting the funds to an arbitrary receiver address. The role of the `referenceLabel` is described in more detail in another section below. 

The endpoint checks if a subaccount exists for `taxId`, and if the `amount` is within the allowed minting limits (by checking against the accumulated limits in the `kyc` data of the subaccount). If the amount is within the limit, make a GET request to the '[Generates pay-in BR Code](https://brla-account-api.readme.io/reference/businesspayinbrcode)' endpoint on BRLA's API and return the received `brCode` to the caller of the endpoint. 

TBD: Can both [this](https://brla-account-api.readme.io/reference/businesspayinpixtousdhistory) and [this](https://brla-account-api.readme.io/reference/businesspayinpixtotoken) endpoint be used for receiving tokens on the users evm account?

#### teleport() (`/teleport`)
This endpoint receives a `taxId`, an `amount`, and the `receiverAddress` (the address of the Moonbeam ephemeral account). It will check the history of [on-chain ins](https://brla-account-api.readme.io/reference/businessonchainhistoryin) to find a previous pay-in request that has the derived hash identifier of the `receiverAddress` for the `referenceLabel`. Only if a matching pay-in request is found, will the tokens be teleported.  

For the teleport, it will call into the [quote](https://brla-account-api.readme.io/reference/businessfastquote) and [convert](https://brla-account-api.readme.io/reference/businessswap) endpoints on BRLA to teleport the specified amount to the receiver account on Moonbeam. 

#### Derivation of the `referenceLabel`

The BRLA API only allows to define `referenceLabel`s with a length of up to 19 alphanumeric characters. To prevent malicious actors from teleporting tokens to a custom account, we need to already pass an identifier of the eligible destination account for the automatic teleport. In order to do this, the account address is hashed with SHA-256 and the first 19 characters of the resulting hash are used as the identifier. 

#### Offramp 

These are phases that already exist. Some of them need adjusted naming
 -  'prepareTransactions' -> 'prepareOfframpTransactions'

#### Onramp

 - 'prepareOnrampTransactions': Store all the metadata relevant for the onramp in the `onrampingState` in local storage. This includes the pay-in amount, pay-out amount (after fees), input asset (will always be BRL), destination token, destination network, destination account. Also, ephemeral accounts on Moonbeam are created. The transactions for sending BRLA from the ephemeral account on Moonbeam to Pendulum, approving and swapping the tokens on Nabla, and either a) a transaction to transfer the target token to AssetHub or b) an XCM transfer from the ephemeral account on Pendulum to the `evm` address denoted by the subaccounts data returned by the BRLA API on Moonbeam. 
- 'createPayInRequest': Call into the new `pay-in` endpoint of the signer service to receive a BR code. Store the received be code in local storage. 
- 'waitForPayIn':  Show that BR code as a QR code to the user and poll the balance of the subaccounts `evm` address on Polygon. Once received, proceed to the next phase. 
- 'teleportFromPolygonToMoonbeam': Call into the `/teleport` endpoint with the `taxId`, `amount` and `receiverAddress` (public key of Moonbeam ephemeral). Wait for the tokens to arrive on the ephemeral account on Moonbeam.
- 'executeMoonbeamToPendulumXCM': Execute the previously created XCM transaction to transfer the BRL tokens from the ephemeral account on Moonbeam to the ephemeral account on Pendulum.
- 'subsidizePreSwap': Check if the received amount still matches the required input amount of the swap. Subsidize BRL tokens if some are missing. 
- 'nablaApprove': Same old
- 'nablaSwap': Same old
- 'subsidizePostSwap': Check if the received amount still matches the required output amount of the swap. Subsidize axlUSDC or AssetHub USDC if some are missing. 
- a) 'executePendulumToAssetHubXCM': Execute the previously created XCM transaction to transfer the AssetHub USDC to the designated destination account. 
- b) 'executePendulumToMoonbeamXCM': Execute the previously created XCM transaction to transfer the axlUSDC to the ephemeral account on Moonbeam.
- b) 'squidRouter': Do a squidrouter swap from Moonbeam to the designated target chain with the target token to the target account.

ebma:
## Summary
This PR migrates core ramping logic from frontend to backend (signer-service), implementing:
- State machine architecture for ramp flows
- PostgreSQL persistence layer
- Resilient transaction processing
- Phase-based execution handlers

```mermaid
flowchart TD
    Frontend -->|API Calls| Backend
    Backend --> PhaseHandler
    PhaseHandler --> Database[(PostgreSQL)]
    PhaseHandler --> SquidRouter
    PhaseHandler --> PolkadotService
    PhaseHandler --> RecoveryWorker
    PhaseHandler -->|Events| Frontend
```

### Backend Implementation
1. **Phase Handlers System**
   - `signer-service/src/api/services/phases/*`
   - State machine pattern for ramp lifecycle management
   - Handlers for initial checks, transaction prep, and routing

2. **Database Schema**
   - Added `ramp_states` and `phase_metadata` tables
   - Migration scripts in `signer-service/src/database/migrations/*`

3. **API Endpoints**
   - New controllers in `signer-service/src/api/controllers/*`
   - Quote generation and ramp initiation flows
   - Status polling endpoints

4. **Services**
   - Transaction processing in `signer-service/src/api/services/transactions/*`
   - Polkadot integration improvements
   - Error recovery workers

### Frontend Adjustments
- Updated hook integrations (`src/hooks/offramp/*`)
- New API client services (`src/services/backend.ts`)
- Enhanced error handling in UI components

## Related Issues
- Closes #497 (BRL support)


signer-service/src/api/services/transactions/nabla/index.ts
```diff
@@ -16,31 +19,40 @@ export async function createNablaTransactionsForQuote(quote: QuoteTicketAttribut
...
+      ? getPendulumDetails(quote.outputCurrency, toNetwork)
+      : getPendulumDetails(quote.outputCurrency);
+
+  const amountRaw = multiplyByPowerOfTen(
+    new Big(quote.inputAmount),
+    inputTokenPendulumDetails.pendulumDecimals,
+  ).toFixed(0, 0);
   const pendulumEphemeralAddress = ephemeral.address;
-  const nablaHardMinimumOutputRaw = quote.outputAmount;
+  const nablaHardMinimumOutputRaw = new Big(quote.outputAmount).add(new Big(quote.fee)).toFixed(0, 0); // TODO we're not allowing subsidy anymore?
```
ebma, line 44: I just put `quote.outputAmount` as a placeholder for now, we can also change that 😬

signer-service/src/api/services/transactions/onrampTransactions.ts
```diff
@@ -20,186 +36,136 @@ export async function prepareOnrampTransactions(
...
         network: account.network,
         nonce: 0,
         signer: account.address,
       });
-    } else if (accountNetworkId === getNetworkId(Networks.Pendulum)) {
-      // TODO implement creation of unsigned ephemeral txs for swaps
 
-      if (toNetworkId === getNetworkId(Networks.AssetHub)) {
-        // TODO implement creation of unsigned ephemeral tx for Pendulum -> AssetHub
+      if (toNetworkId !== getNetworkId(Networks.AssetHub)) {
```
ebma, line 140: I think I'd flip this condition and make it 
```suggestion
      if (toNetworkId === getNetworkId(Networks.AssetHub)) {
      // Do the Assethub stuff
      } else { // Do Moonbeam stuff }
```

signer-service/src/api/services/transactions/nabla/index.ts
gianfra-t: I thought about adding another field, but then this requires changing the model, etc. And we can reconstruct the "outputBeforeFees" amount by just adding it back. It's a bit more verbose but I think is fine.

signer-service/src/api/services/transactions/stellar/offrampTransaction.ts
```diff
@@ -124,6 +121,27 @@ export async function buildPaymentAndMergeTx(
     createAccountTransaction: createAccountTransaction.toEnvelope().toXDR().toString('base64'),
     paymentTransaction: paymentTransaction.toEnvelope().toXDR().toString('base64'),
     mergeAccountTransaction: mergeAccountTransaction.toEnvelope().toXDR().toString('base64'),
-    fundingAccountSequence,
+    expectedSequenceNumber: String(expectedSequenceNumber),
   };
 }
+
+async function getFutureShiftedLedgerSequence(horizonServer: HorizonServer, shiftAmount = 32) {
```
ebma, line 128: I think I wouldn't make the `shiftAmount` configurable. There is no scenario in which you want to use a number different than `32`.

api/src/api/services/phases/phase-processor.integration.test.ts
```diff
@@ -214,10 +223,13 @@ describe('PhaseProcessor Integration Test', () => {
...
       const rampService = new RampService();
       const quoteService = new QuoteService();
+
+      registerPhaseHandlers();
+
       const additionalData = {
         walletAddress: EVM_TESTING_ADDRESS,
         paymentData: {
-          amount: '1', // Relevant for test???
+          amount: '0.0000000001', // TODO this is user controlled, not only in test, perhaps we should protect. It should come from the quote.
```
ebma, line 232: I think this is too low to be a valid Stellar amount. Stellar only supports 1e7 decimals.

api/src/api/services/transactions/stellar/offrampTransaction.ts
```diff
@@ -7,14 +7,23 @@ import Big from 'big.js';
 const FUNDING_PUBLIC_KEY = FUNDING_SECRET ? Keypair.fromSecret(FUNDING_SECRET).publicKey() : '';
 const NETWORK_PASSPHRASE = Networks.PUBLIC;
 const MAX_TIME = Date.now() + 1000 * 60 * 10;
+const SEQUENCE_SHIFT_IN_SECONDS = 240; // 4 minutes
```
ebma, line 10: I don't understand the name. What's a sequence shift? 😅

api/src/api/services/phases/handlers/complete-phase-handler.ts
```diff
@@ -0,0 +1,27 @@
...
+import logger from '../../../../config/logger';
+import { Transaction } from 'stellar-sdk';
+import { Horizon, NetworkError, Networks } from 'stellar-sdk';
+
+export const horizonServer = new Horizon.Server(HORIZON_URL);
+const NETWORK_PASSPHRASE = Networks.PUBLIC;
+/**
+ * Placeholder phase to perform any final actions after the ramp flow is complete.
+ */
+export class InitialPhaseHandler extends BasePhaseHandler {
```
ebma, line 13:
```suggestion
export class CompletePhaseHandler extends BasePhaseHandler {
```

api/src/api/services/phases/handlers/create-pay-in-handler.ts
```diff
@@ -14,9 +14,9 @@ export class CreatePayInPhaseHandler extends BasePhaseHandler {
   }
 
   protected async executePhase(state: RampState): Promise<RampState> {
```
ebma, line 16: I think the name of this handler doesn't describe the correct phase. If we look at the phases I originally defined in #497, we have the following:

> 'prepareOnrampTransactions': Store all the metadata relevant for the onramp in the onrampingState in local storage. This includes the pay-in amount, pay-out amount (after fees), input asset (will always be BRL), destination token, destination network, destination account. Also, ephemeral accounts on Moonbeam are created. The transactions for sending BRLA from the ephemeral account on Moonbeam to Pendulum, approving and swapping the tokens on Nabla, and either a) a transaction to transfer the target token to AssetHub or b) an XCM transfer from the ephemeral account on Pendulum to the evm address denoted by the subaccounts data returned by the BRLA API on Moonbeam.
> 'createPayInRequest': Call into the new pay-in endpoint of the signer service to receive a BR code. Store the received be code in local storage.
> 'waitForPayIn': Show that BR code as a QR code to the user and poll the balance of the subaccounts evm address on Polygon. Once received, proceed to the next phase.
> 'teleportFromPolygonToMoonbeam': Call into the /teleport endpoint with the taxId, amount and receiverAddress (public key of Moonbeam ephemeral). Wait for the tokens to arrive on the ephemeral account on Moonbeam.
> 'executeMoonbeamToPendulumXCM': Execute the previously created XCM transaction to transfer the BRL tokens from the ephemeral account on Moonbeam to the ephemeral account on Pendulum.

- `prepareOnrampTransactions` happens not in one of these handlers but when doing the call to `registerRamp`.
- `createPayInRequest` is also not a phase handler but the pay-in request also has to be created when the user calls `registerRamp`. We should return the BR code where the user needs to send money to in the response of `registerRamp` already
- `teleportFromPolygonToMoonbeam` would be what the phase handler you implemented here would be about.

api/src/api/services/transactions/stellar/offrampTransaction.ts
gianfra-t: Oh, since the sequence seems to be the ledger shifted 32 bits... I just named it sequence shift (in time!). Is it confusing haha? We can change it of course. `SEQUENCE_TIME_SHIFT_SECONDS` ??

api/src/api/services/phases/phase-processor.integration.test.ts
ebma: No, if we can remove it, that's better 👍

api/src/api/services/transactions/stellar/offrampTransaction.ts
ebma: But just from the name I wouldn't even know how to use it. How about we use different constants. 

```suggestion
const STELLAR_LEDGER_CLOSE_TIME_SECONDS = 6; 
const STELLAR_LEDGER_BUFFER_WINDOW_SECONDS = 5 * 60; // 5 minutes
/// This is used to add a buffer on top of our current ledger sequence number. We assume
const STELLAR_LEDGERS_PER_BUFFER = Math.ceil(STELLAR_LEDGER_BUFFER_WINDOW_SECONDS / STELLAR_LEDGER_CLOSE_TIME_SECONDS);
```

api/src/api/services/phases/handlers/create-pay-in-handler.ts
ebma: Sounds good 👍

api/src/api/services/transactions/squidrouter/offramp.ts
```diff
@@ -71,15 +71,15 @@ export async function createOfframpSquidrouterTransactions(
       value: '0',
       gas: '150000',
       maxFeePerGas: String(maxFeePerGas),
-      maxPriorityFeePerGas: String(maxPriorityFeePerGas),
+      maxPriorityFeePerGas: String(maxFeePerGas),
```
ebma, line 74: Is this intentional?

shared/src/helpers/signUnsigned.ts
```diff
@@ -134,11 +134,14 @@ export async function signUnsignedTransactions(
...
+        const txData = { 
+          to: tx.tx_data.to,
+          data: tx.tx_data.data,
           value: BigInt(tx.tx_data.value),
-          maxFeePerGas: tx.tx_data.maxFeePerGas ? BigInt(tx.tx_data.maxFeePerGas) : BigInt(0),
-          maxPriorityFeePerGas: tx.tx_data.maxPriorityFeePerGas ? BigInt(tx.tx_data.maxPriorityFeePerGas) : BigInt(0),
+          nonce: Number(tx.nonce), 
+          gas: BigInt(tx.tx_data.gas),
+          maxFeePerGas: tx.tx_data.maxFeePerGas ? BigInt(tx.tx_data.maxFeePerGas) * 5n : BigInt(187500000000),
+          maxPriorityFeePerGas: tx.tx_data.maxPriorityFeePerGas ? BigInt(tx.tx_data.maxPriorityFeePerGas) * 5n : BigInt(187500000000),
```
ebma, line 144: Where is the last number coming from? Just arbitrary?

api/src/api/services/transactions/stellar/offrampTransaction.ts
gianfra-t: Sorry I'm just seeing this. I changed it to `SEQUENCE_TIME_WINDOW_IN_SECONDS` and added a few comments. It would be comparable to your `STELLAR_LEDGER_BUFFER_WINDOW_SECONDS`. Added a comment on constants also.

api/src/api/services/transactions/squidrouter/offramp.ts
gianfra-t: Yes. the priority sometimes is estimated as 0, but this then fails when signing.

shared/src/helpers/signUnsigned.ts
gianfra-t: Just arbitrary, yes.

PR #540, opened 2025-03-26
[READY] Show the token balances in the dropdown (sorted)

Sharqiewicz:
In order to be able to get multiple token balances and manage their state - I had to refactor the `useInputTokenBalance` hook to allow fetching an array of tokens. Now `useInputTokenBalances` is available. 
`useInputTokenBalance`  just queries the `useInputTokenBalances` hook with `[chosenToken]`

- [x] show token balances in the dropdown
- [x] sort the tokens from the highest balance


ebma (approved):
Looks good to me 👍

PR #543, opened 2025-04-02
Updating Navbar

prayagd:
- Remove Sell crypto, How it works and community
- Add Support, Docs and FAQ
- Links also updated

### DIFF
--- a/src/components/Navbar/index.tsx
+++ b/src/components/Navbar/index.tsx
@@ -89,9 +89,9 @@ const Links = () => {
   const { t } = useTranslation();
 
   const links = [
-    { title: t('components.navbar.sellCrypto'), href: '/' },
-    { title: t('components.navbar.howItWorks'), href: 'https://www.vortexfinance.co/#lowest-code' },
-    { title: t('components.navbar.community'), href: 'https://www.vortexfinance.co/#call-to-action' },
+    { title: t('components.navbar.support'), href: 'https://pendulum.gitbook.io/vortex/support-and-contact' },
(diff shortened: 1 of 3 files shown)


src/components/Navbar/index.tsx
```diff
@@ -89,9 +89,9 @@ const Links = () => {
   const { t } = useTranslation();
 
   const links = [
-    { title: t('components.navbar.sellCrypto'), href: '/' },
-    { title: t('components.navbar.howItWorks'), href: 'https://www.vortexfinance.co/#lowest-code' },
-    { title: t('components.navbar.community'), href: 'https://www.vortexfinance.co/#call-to-action' },
+    { title: t('components.navbar.support'), href: 'https://pendulum.gitbook.io/vortex/support-and-contact' },
+    { title: t('components.navbar.docs'), href: 'https://pendulum.gitbook.io/vortex' },
+    { title: t('components.navbar.faq'), href: 'https://app.vortexfinance.co/#faq' },
```
ebma, line 94: What's the expected behavior for clicking on the FAQ link? Do you want it to scroll down on the same page or open a new page?

prayagd: scroll down to the same page to the faq section

ebma: This doesn't work in that case then. For one, you are always putting the production address so on deploy previews etc. it will always open a new page. If you feel like it, feel free to try troubleshooting it yourself with AI. Otherwise one of the @pendulum-chain/devs needs to add that before we can merge.

prayagd: ok let me give a try

ebma:
Superseded by #605.

PR #565, opened 2025-04-15
Fix progress bar percentage too low

ebma:
Currently, the progress bar only reaches ~50% when shown on the progress page before switching to the success page. This is because the total number of phases defined in the respective ramp-phase objects is larger than the relevant phases (as we need to exhaust the `RampPhase` type). This PR adds a custom constant to denote the relevant phases for the progress bar. 

This also runs prettier on all files.


gianfra-t (approved):
Lucky both ramp types have the same active count 👍

PR #570, opened 2025-04-16
Add support for already KYB'd businesses

gianfra-t:


frontend/src/hooks/offramp/useSubmitRamp.ts
```diff
@@ -82,10 +83,18 @@ export const useSubmitRamp = () => {
             if (!response.ok) {
               // Response can also fail due to invalid KYC. Nevertheless, this should never be the case, as when we create the user we wait for the KYC
               // to be valid, or retry.
-              if (response.status === 404) {
+              if (response.status === 404 && !isValidCnpj(taxId) && isValidCpf(taxId)) {
```
ebma, line 86: It should be enough to only check for a valid CPF. The same input cannot be a valid CPF and CNPJ.

```suggestion
              if (response.status === 404 && isValidCpf(taxId)) {
```

api/src/constants/constants.ts
```diff
@@ -8,7 +8,7 @@ const MOONBEAM_RECEIVER_CONTRACT_ADDRESS = '0x2AB52086e8edaB28193172209407FF9df1
 const STELLAR_EPHEMERAL_STARTING_BALANCE_UNITS = '2.5'; // Amount to send to the new stellar ephemeral account created
 const PENDULUM_EPHEMERAL_STARTING_BALANCE_UNITS = '0.1'; // Amount to send to the new pendulum ephemeral account created
 const MOONBEAM_EPHEMERAL_STARTING_BALANCE_UNITS = '3'; // Amount to send to the new moonbeam ephemeral account created
-const BRLA_BASE_URL = 'https://api.brla.digital:5567/v1/business';
+const BRLA_BASE_URL = 'https://api.brla.digital:4567/v1/business';
```
ebma, line 11:
```suggestion
const BRLA_BASE_URL = 'https://api.brla.digital:5567/v1/business';
```

PR #585, opened 2025-04-23
Adds checks for squidrouter swap transaction value.

gianfra-t:
### Changes
- Increase ephemeral funding.
- Will not create quote if ephemeral funding constant is not be enough to pay for the squidrouter transaction.


api/src/api/services/moonbeam/balance.ts
```diff
@@ -75,12 +75,12 @@ export const fundMoonbeamEphemeralAccount = async (ephemeralAddress: string) =>
...
 };
 
-export function getMoonbeamFundingData(decimals: number): {
+export function getMoonbeamFundingData(decimals: number, boosted: boolean = false): {
   fundingAmountRaw: string;
   walletClient: ReturnType<typeof createMoonbeamClientsAndConfig>['walletClient'];
   publicClient: ReturnType<typeof createMoonbeamClientsAndConfig>['publicClient'];
 } {
-  const fundingAmountUnits = Big(MOONBEAM_EPHEMERAL_STARTING_BALANCE_UNITS);
+  const fundingAmountUnits = boosted ? Big(MOONBEAM_EPHEMERAL_STARTING_BALANCE_UNITS_ETHEREUM) : Big(MOONBEAM_EPHEMERAL_STARTING_BALANCE_UNITS);
```
ebma, line 83: `boosted` reminds me of squidrouter. But that's not entirely true so let's change the name. Maybe call it `largeFunding: boolean = false`?

gianfra-t:
@pendulum-chain/devs Tested an offramp Assethub -> Pix locally. So far so good.

PR #587, opened 2025-04-23
Add kyc level 2

gianfra-t:


frontend/src/components/BrlaComponents/BrlaExtendedForm.tsx
```diff
@@ -6,15 +6,31 @@ import { useKYCForm } from '../../hooks/brla/useKYCForm';
...
 
   const { t } = useTranslation();
 
-  const PIXKYCFORM_FIELDS: BrlaFieldProps[] = [
+  const taxId = useTaxId();
+
+  
+  const handleDocumentSubmit = useCallback(() => {
+    setIsSubmitted(true);
+    const taxIdToSet = taxId ? taxId : null
```
Sharqiewicz, line 28: const taxIdToSet = taxId || null

frontend/src/components/BrlaComponents/BrlaExtendedForm.tsx
```diff
@@ -88,12 +104,54 @@ export const PIXKYCForm = () => {
...
+    pixformFields.push({
+      id: ExtendedBrlaFieldOptions.START_DATE,
+      label: t('components.brlaExtendedForm.form.startDate'),
+      type: 'date',
+      required: true,
+      index: 10,
+    });
+    pixformFields.push({
+      id: ExtendedBrlaFieldOptions.PARTNER_CPF,
+      label: t('components.brlaExtendedForm.form.parnterCpf'),
```
Sharqiewicz, line 129: typo parnterCpf => partnerCpf

frontend/src/components/BrlaComponents/BrlaExtendedForm.tsx
```diff
@@ -88,12 +104,54 @@ export const PIXKYCForm = () => {
     },
   ];
 
+  if (!taxId) {
+    return null;
+  }
+
+  if (isValidCnpj(taxId)) {
```
Sharqiewicz, line 111: To improve clarity and maintainability, I suggest extracting the CNPJ-specific field additions into a separate function, like addCnpjFields, to clearly group and document these fields

frontend/src/components/BrlaComponents/BrlaExtendedForm.tsx
```diff
@@ -88,12 +104,54 @@ export const PIXKYCForm = () => {
...
+    pixformFields.push({
+      id: ExtendedBrlaFieldOptions.PARTNER_CPF,
+      label: t('components.brlaExtendedForm.form.parnterCpf'),
+      type: 'text',
+      required: true,
+      index: 11,
+    });
+  }
+
   return (
```
Sharqiewicz, line 136:
```
 return (
    <div className="relative">
      {isSubmitted ? (
        <VerificationStatus
          status={verificationStatus}
          message={statusMessage}
          isLevel2={offrampKycLevel2Started}
        />
      ) : offrampKycLevel2Started ? (
        <DocumentUpload
          onSubmitHandler={handleDocumentSubmit}
          onBackClick={handleBackClick}
        />
      ) : offrampKycStarted ? (
        <KYCForm
          fields={pixformFields}
          form={kycForm}
          onSubmit={handleFormSubmit}
          onBackClick={handleBackClick}
        />
      ) : null}
    </div>
  );

File: frontend/src/constants/constants.ts [hunk 5]
```diff
@@ -6,6 +6,7 @@ export const MOONBEAM_WSS = 'wss://wss.api.moonbeam.network';
 export const WALLETCONNECT_ASSETHUB_ID = 'polkadot:68d56f15f85d3136970ec16946040bc1';
 
 export const TRANSFER_WAITING_TIME_SECONDS = 6000;
+export const BRLA_MAXIMUM_LEVEL_2_AMOUNT_UNITS = 1000000;
```
Sharqiewicz, line 9: can we link brla docs for the future reference?

frontend/src/hooks/brla/useBRLAKYCProcess/index.tsx
```diff
@@ -60,25 +63,43 @@ export function useKYCProcess() {
...
   const [isSubmitted, setIsSubmitted] = useState(false);
 
   const [cpf, setCpf] = useState<string | null>(null);
 
   const queryClient = useQueryClient();
-  const { data: kycResponse, error } = useKycStatusQuery(cpf);
-  const { setRampKycStarted, resetRampState } = useRampActions();
+
+  const offrampKycLevel2Started = useRampKycLevel2Started();
+  const desiredLevel = offrampKycLevel2Started ? 2 : 1;
```
Sharqiewicz, line 76: I was wondering if we could implement an enum for KYCLevels instead of using magic numbers?
This might help make the code more readable and maintainable.

frontend/src/hooks/brla/useBRLAKYCProcess/index.tsx
```diff
@@ -119,14 +136,27 @@ export function useKYCProcess() {
...
-          ...formData,
-          cpf: taxId,
-          birthdate: formData.birthdate.getTime(),
-          address: addressObject,
-          taxIdType: 'CPF',
-        });
+        if (isValidCnpj(taxId)) {
+          await createSubaccount({
+            ...formData,
+            // eslint-disable-next-line  @typescript-eslint/no-non-null-assertion
```
Sharqiewicz, line 142: Please just add a condition like
```
if (!formData.partnerCpf) {
        throw new Error('useKYCProcess: Partner CPF must be defined at this point');
}
```

frontend/src/hooks/brla/useBRLAKYCProcess/index.tsx
```diff
@@ -173,15 +213,62 @@ export function useKYCProcess() {
...
-    STATUS_MESSAGES.SUCCESS,
-    STATUS_MESSAGES.REJECTED,
+    proceedWithKYCLevel2,
+    proceedWithRamp,
+    STATUS_MESSAGES,
+    offrampKycLevel2Started
   ]);
 
+  // Handler for KYC level 2
+  useEffect(() => {
```
Sharqiewicz, line 228: Maybe we should consider using `useCallback` instead `useEffect` for KYC Level 1 and 2

Sharqiewicz (changes requested):

frontend/src/components/BrlaComponents/KYCLevel2Form/index.tsx
```diff
@@ -0,0 +1,224 @@
...
+    file: File | null,
+    setter: React.Dispatch<React.SetStateAction<File | null>>,
+    validSetter: React.Dispatch<React.SetStateAction<boolean>>
+  ) => {
+    if (!file) {
+      setter(null);
+      validSetter(false);
+      return;
+    }
+    if (!ALLOWED_TYPES.includes(file.type)) {
```
Sharqiewicz, line 71: I think we should enhance our file validation, the file.type can be spoofed and we should also check for the file extension like:
```
const extension = `.${getFileExtension(file.name)}`;
    if (!ALLOWED_EXTENSIONS.includes(extension.toLowerCase())) {
```

We could also use Zod for the validation.

api/src/api/controllers/brla.controller.ts
```diff
@@ -231,7 +235,14 @@ export const createSubaccount = async (
   res: Response<BrlaEndpoints.CreateSubaccountResponse | BrlaEndpoints.BrlaErrorResponse>,
 ): Promise<void> => {
   try {
-    const { cpf: taxId } = req.body;
+    const { cpf, cnpj, taxIdType } = req.body;
+    
+    const taxId = taxIdType === 'CNPJ' ? cnpj : cpf;
+
+    if (!taxId) {
+      res.status(400).json({ error: 'Missing cpf or cnpj' });
```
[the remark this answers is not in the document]
gianfra-t: Yes let's do this in a separate PR. I thought about it before and it's good, but it would touch all endpoints.

api/src/api/services/kyc/kyc.service.ts
```diff
@@ -0,0 +1,148 @@
...
+    status: KycLevel2Status,
+    uploadData?: KycLevel2Response, 
+    errorLog?: any,
+  ): Promise<[number, KycLevel2[]]> {
+    const kycLevel2 = await this.getKycLevel2ById(id);
+    if (!kycLevel2) {
+      throw new Error(`KYC Level 2 entry with id ${id} not found`);
+    }
+
+    const updateData: any = { status, uploadData };
```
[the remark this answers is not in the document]
gianfra-t: Actually this method `updateKycLevel2` seems to be unused now after some changes, so I'll remove it.

api/src/api/services/kyc/kyc.service.ts
```diff
@@ -0,0 +1,148 @@
...
+    );
+  }
+
+
+  protected async getKycLevel2ById(id: string): Promise<KycLevel2 | null> {
+    return KycLevel2.findByPk(id);
+  }
+
+
+  protected async getLatestKycLevel2BySubaccount(
```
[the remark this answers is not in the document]
gianfra-t: Same as above. In general I tried to follow a similar patter as `quote.service.ts` and `ramp.service.ts`.

frontend/package.json
```diff
@@ -51,6 +51,7 @@
     "big.js": "^6.2.1",
     "bn.js": "^5.2.1",
     "buffer": "^6.0.3",
+    "heroicons": "^2.2.0",
```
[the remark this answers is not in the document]
gianfra-t: Thanks good catch. removing.

frontend/src/components/BrlaComponents/BrlaExtendedForm.tsx
gianfra-t: Sure, no strong opinion on this. Also for the always-used fields?

frontend/src/constants/constants.ts
gianfra-t: This level is not known, at least to me, so it was a placeholder to test. I'll add a comment.

frontend/src/hooks/brla/useBRLAKYCProcess/index.tsx
gianfra-t: True, I'll add one for the numbers in this component.

api/src/api/controllers/brla.controller.ts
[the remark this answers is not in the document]
ebma: I would assume a similar enum already exists in one of the libraries.

api/src/api/controllers/brla.controller.ts
```diff
@@ -231,26 +239,62 @@ export const createSubaccount = async (
...
     const brlaApiService = BrlaApiService.getInstance();
+    // Convert birthdate from number to BRLA format
+    const birthdate = convertDateToBRLAFormat(req.body.birthdate);
+    // if company startDate field was provided, convert it to BRLA format
+    const startDate = convertDateToBRLAFormat(req.body.startDate);
+
+    let subaccountPayload: RegisterSubaccountPayload = { ...req.body, birthdate, startDate };
+
+    // Extra validation for company fields
+    if (taxIdType === 'CNPJ') {
```
ebma, line 260: Wouldn't we need to check for the business registration that _both_ the CPF and CNPJ fields are provided? The cpf field of the 'partner' is still required when registering a business, isn't it?

api/src/api/controllers/brla.controller.ts
```diff
@@ -367,42 +423,54 @@ export const validatePixKey = async (
...
     const brlaApiService = BrlaApiService.getInstance();
     const subaccount = await brlaApiService.getSubaccount(taxId);
+
     if (!subaccount) {
-      res.status(400).json({ error: 'taxId invalid' });
+      res.status(404).json({ error: 'Subaccount not found' });
+      return;
+    }
+
+    if (subaccount.kyc.level < 1) {
```
ebma, line 458: Maybe it's better to explicitly check if it's `1`. We don't want to do it for any other level. If the account is at `0` or already at `2` it should fail here. 

```suggestion
    if (subaccount.kyc.level !== 1) {
```

api/src/api/routes/v1/brla.route.ts
```diff
@@ -25,6 +24,7 @@ router.route('/triggerOfframp').post(validateBrlaTriggerOfframpInput, brlaContro
 
 router.route('/createSubaccount').post(validataSubaccountCreation, brlaController.createSubaccount);
 
-router.route('/triggerPayIn').post(validateTriggerPayIn, brlaController.triggerPayIn);
+router.route('/startKYC2').post(brlaController.startKYC2);
```
ebma, line 27: Just for the sake of it, why don't we add a `validate...` function like we do for any other post request.

api/src/api/services/brla/types.ts
```diff
@@ -210,6 +210,36 @@ export const enum SmartContractOperationType {
...
+export interface KycLevel2Payload {
+  documentType: KYCDocType;
+}
+
+export interface KycLevel2Response {
+  id: string;
+  selfieUploadUrl: string;
+  RGFrontUploadUrl: string;
+  RGBackUploadUrl: string;
+  CNHUploadUrl: string;
```
ebma, line 231: The capitalization is a little unfortunate. How about 
```suggestion
  rgFrontUploadUrl: string;
  rgBackUploadUrl: string;
  cnhUploadUrl: string;
```

api/src/api/services/kyc/kyc.service.ts
```diff
@@ -0,0 +1,123 @@
...
+    if (!kycLevel2) {
+      return false;
+    }
+
+    return kycLevel2.status === KycLevel2Status.ACCEPTED;
+  }
+
+}
+
+export default new KycService();
```
ebma, line 123: How about we make it a singleton instead of instantiating with the default export? In the past I had to change it from this to a singleton for a few other services to be able to delay the initialization. For this one here, it works either way. Singleton would feel a bit cleaner for me though.

api/src/constants/constants.ts
```diff
@@ -9,7 +9,7 @@ const STELLAR_EPHEMERAL_STARTING_BALANCE_UNITS = '2.5'; // Amount to send to the
 const PENDULUM_EPHEMERAL_STARTING_BALANCE_UNITS = '0.1'; // Amount to send to the new pendulum ephemeral account created
 const MOONBEAM_EPHEMERAL_STARTING_BALANCE_UNITS = '30'; // Amount to send to the new moonbeam ephemeral account created
 const MOONBEAM_EPHEMERAL_STARTING_BALANCE_UNITS_ETHEREUM = '100'; // Amount to send to the new moonbeam ephemeral account created when onramping to Ethereum
-const BRLA_BASE_URL = 'https://api.brla.digital:5567/v1/business';
+const BRLA_BASE_URL = 'https://api.brla.digital:5567/v1/business'; // TODO - put back production env after testing.
```
ebma, line 12: Remove TODO

frontend/src/constants/constants.ts
```diff
@@ -6,6 +6,7 @@ export const MOONBEAM_WSS = 'wss://wss.api.moonbeam.network';
 export const WALLETCONNECT_ASSETHUB_ID = 'polkadot:68d56f15f85d3136970ec16946040bc1';
 
 export const TRANSFER_WAITING_TIME_SECONDS = 6000;
+export const BRLA_MAXIMUM_LEVEL_2_AMOUNT_UNITS = 1000000; // TODO get the actual value from BRLA.
```
ebma, line 9: Why do we even need this? I created the proper check in a different PR. Do we still need to merge those changes in here?

frontend/src/hooks/brla/useKYCStatusQuery/index.tsx
```diff
@@ -1,12 +1,13 @@
...
+import { BrlaKycStatus, KycLevel } from '../useBRLAKYCProcess';
+import { fetchKycStatus, KycStatus } from '../../../services/signingService';
 
 const POLLING_INTERVAL_MS = 2000;
 const RETRY_DELAY_MS = 5000; // 5 seconds
 const MAX_RETRIES = 5;
 
-export const useKycStatusQuery = (cpf: string | null) => {
+export const useKycStatusQuery = (cpf: string | null, level: KycLevel = KycLevel.LEVEL_1) => {
+  console.log('useKycStatusQuery', cpf, level);
```
ebma, line 10: Nit: Remove

frontend/src/pages/ramp-form/index.tsx
```diff
@@ -15,6 +15,7 @@ export const RampForm = () => {
   const activeSwapDirection = useRampDirection();
   const onSwapDirectionToggle = useRampDirectionToggle();
   const offrampKycStarted = useRampKycStarted();
+  const offrampKycLevel2Started = useRampKycLevel2Started();
```
ebma, line 18: Let's move away from the `offramp` and embrace the `ramp`
```suggestion
  const rampKycStarted = useRampKycStarted();
  const rampKycLevel2Started = useRampKycLevel2Started();
```

frontend/src/services/api/brla.service.ts
```diff
@@ -1,6 +1,12 @@
 import { BrlaEndpoints, EvmAddress } from 'shared';
 import { apiRequest } from './api-client';
 
+export enum KYCDocType {
+  RG = 'RG',
+  CNH = 'CNH',
+}
+
+export type KYCDataUploadFileFileds = BrlaEndpoints.KYCDataUploadFileFileds;
```
ebma, line 9:
```suggestion
export type KYCDataUploadFileFiles = BrlaEndpoints.KYCDataUploadFileFiles;
```

frontend/src/services/initialChecks.ts
```diff
@@ -1,45 +1,25 @@
...
-
-        if (amountNum <= remainingLimitNum) {
-          return true;
-        } else {
-          showToast(
-            ToastMessage.RAMP_LIMIT_EXCEEDED,
-            t('toasts.rampLimitExceeded', { remaining: remainingLimitInUnits }),
-          );
+        const amountBigNumber = Big(amountUnits);
+        if (amountBigNumber.gt(BRLA_MAXIMUM_LEVEL_2_AMOUNT_UNITS)) {
```
ebma, line 12: Okay now I get why you introduced that other constant. Of course, in this PR we change the behaviour and allow the user to define amounts larger than their current limit. So checking this 'withinAllowedLimits' on the form doesn't make sense. 

Instead of using that constant I would propose the following:
- We change it to check if the KYC level of the respective subaccount is already at level 2.
- If the subaccount is at KYC level 2 (or above), we keep the `useRampAmountWithinAllowedLimits` check as before, ie. we **do** check if the amount is perfectly within the limit. 
- If the subaccount is lower than KYC level 2, we just return true here, ie. we say the amount is within the allowed limits. This has an implicit assumption that, when the user is currently in KYC level 1, they don't specify an amount that is larger than what they would be allowed to ramp once they are upgraded to KYC level 2. However, at the current stage and probably never ever in the future, we'll see users do a single ramp with amounts that exceed the KYC level 2 limit. So I would not care about this edge case at the moment.

frontend/src/translations/en.json
```diff
@@ -378,6 +383,20 @@
           "required": "Birthdate is required",
           "future": "Birthdate cannot be in the future",
           "tooOld": "Invalid birthdate"
+        },
+        "companyName": {
+          "required": "Company name is required",
+          "minLength": "Company name must be at least 3 characters",
```
ebma, line 389: Are you sure it's 3 and not 5?

frontend/src/translations/pt.json
ebma, line 1: We are missing a few of the translations that we added to en.json here.

ebma (changes requested):
Looking great overall, nice job :+1: 

I left a few comments, mostly minor stuff.

api/package.json
```diff
@@ -34,6 +34,7 @@
     "@scure/bip39": "^1.5.4",
     "@types/big.js": "^6.2.2",
     "@types/cron": "^2.4.3",
+    "@types/multer": "^1.4.12",
```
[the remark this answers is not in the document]
Sharqiewicz: https://github.com/pendulum-chain/vortex/issues/611

api/src/api/controllers/brla.controller.ts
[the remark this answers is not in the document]
Sharqiewicz: @ebma right, we have `http-status` package, we should use it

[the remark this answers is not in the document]
Sharqiewicz: @pendulum-chain/devs https://github.com/pendulum-chain/vortex/issues/612

frontend/src/services/initialChecks.ts
gianfra-t: Yes I like the solution, let's do that. You're right that currently, level 2 users will incorrectly pass this check.

frontend/src/translations/en.json
gianfra-t: no idea, just copy the other validations. I don't think brla checks this right?

frontend/src/pages/ramp-form/index.tsx
gianfra-t: alright!

api/src/api/services/kyc/kyc.service.ts
gianfra-t: Mmm but isn't this essentially a singleton? Or works like it. By delay initialization you mean, waiting for the used services to load?

api/src/api/controllers/brla.controller.ts
gianfra-t: Correct.

frontend/src/hooks/brla/useKYCStatusQuery/index.tsx
gianfra-t: Yeah I was debugging the endless re-rendering, finally finished that.

api/src/api/services/brla/types.ts
gianfra-t: I had to rollback on this change, the properties are returned with this capitalization from BRLA, it may be better to leave it as-is.

api/src/api/controllers/brla.controller.ts
```diff
@@ -231,20 +237,59 @@ export const createSubaccount = async (
...
+      }
+    }
+
     const subaccount = await brlaApiService.getSubaccount(taxId);
-    if (subaccount) {
+    if (subaccount && subaccount.kyc.level !== 0) {
       res.status(400).json({ error: 'Subaccount already created' });
       return;
+    } else if (subaccount && subaccount.kyc.level === 0) {
+      console.log('subaccountPayload', subaccountPayload);
```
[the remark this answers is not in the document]
gianfra-t: I think it's not a bad idea to leave this for debugging using backend logs.

frontend/src/translations/en.json
ebma: I think they do actually. But we'll see.

api/src/api/controllers/brla.controller.ts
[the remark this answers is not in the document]
ebma: Right, we can leave it. We could still change the line to use the `logger` instance that we have. We should always try to use that `logger` over simple `console.logs` in our backend because the logger will use the respective configuration (it's logging to file as well as console). I also forget about this often though.

api/src/api/routes/v1/brla.route.ts
```diff
@@ -24,7 +25,6 @@ router.route('/triggerOfframp').post(validateBrlaTriggerOfframpInput, brlaContro
 
 router.route('/createSubaccount').post(validataSubaccountCreation, brlaController.createSubaccount);
 
-router.route('/startKYC2').post(brlaController.startKYC2);
-
+router.route('/startKYC2').post(startKYC2Validator, brlaController.startKYC2);
```
ebma, line 28: How about the following name change to align with the other namings?

```suggestion
router.route('/startKYC2').post(validateStartKyc2, brlaController.startKYC2);
```

frontend/src/components/BrlaComponents/KYCLevel2Form/index.tsx
```diff
@@ -137,6 +137,8 @@ export const DocumentUpload: React.FC<DocumentUploadProps> = ({
           console.error('Validation flags were true, but file data is missing. This is a bug.');
           return;
         }
+        console.log('CNH upload URL:', response.uploadUrls);
```
ebma, line 140: Shall we remove this?

api/src/api/controllers/brla.controller.ts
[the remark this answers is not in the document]
gianfra-t: true, I forgot we added this after the migration.

frontend/src/hooks/offramp/useSubmitRamp.ts
```diff
@@ -102,7 +102,7 @@ export const useSubmitRamp = () => {
       
               const amountNum = Number(executionInput.quote.inputAmount);
               const remainingLimitNum = Number(remainingLimitInUnits);
-              if (amountNum > remainingLimitNum) {
+              if (true) {
```
ebma, line 105: I don't think this is intended?

ebma (approved):
Looks good to me 🆗

Sharqiewicz (approved):

PR #597, opened 2025-04-24
Add check for BRL ramping amount

### LINKED ISSUE #518: Add check for BRL ramping amount

ebma:
- [x] Adds an endpoint to check the remaining limit of a BRLA subaccount. The endpoint already subtracts the used limit from the total to calculate the remaining.
- [x] Add a 'preRampCheck' that currently only checks if the amount that the user wants to ramp is within the allowed limits. For simplicity, the check is only executed once the user clicks on the confirm. It's not triggered when the user updates the value in the form. I decided to go this way as we will very soon change the logic to simply change the behaviour of the submit button if the user exceeds the remaining limit instead of showing an error. Thus, going for a better UX at the moment would be wasted effort. 

Supersedes #533 and closes #518.

### DIFF
--- a/api/src/api/controllers/brla.controller.ts
+++ b/api/src/api/controllers/brla.controller.ts
@@ -90,6 +90,51 @@ export const getBrlaUser = async (
   }
 };
 
+export const getBrlaUserRemainingLimit = async (
+  req: Request<unknown, unknown, unknown, BrlaEndpoints.GetUserRemainingLimitRequest>,
+  res: Response<BrlaEndpoints.GetUserRemainingLimitResponse | BrlaEndpoints.BrlaErrorResponse>,
+): Promise<void> => {
+  try {
+    const { taxId } = req.query;
+
+    if (!taxId) {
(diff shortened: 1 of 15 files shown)


gianfra-t (approved):
💯

PR #603, opened 2025-04-28
Improve price fetching

ebma:
This pull request introduces significant enhancements to the pricing API, focusing on error handling, input validation, and support for a new bundled endpoint that aggregates prices from multiple providers. The changes improve robustness, maintainability, and user experience by standardizing error responses, centralizing validation logic, and providing a comprehensive view of pricing data.

### New Features
* **Bundled Price Endpoint**: Added a new `/all` route to fetch prices from all providers in a single request. This includes the `getAllPricesBundled` controller and `AllPricesResponse` type for consistent response formatting. [[1]](diffhunk://#diff-33fd96bf8309be33cda840428fae32e5b13a939f5398a9623f43364fbdb00c66L102-R229) [[2]](diffhunk://#diff-970f1c63a391778b61a50cad7bfed40900a0c6a7f95c1e35ea749fb87950493dL2-R15)

### Error Handling Improvements
* **Standardized Errors**: Introduced custom error types like `InvalidAmountError`, `InvalidParameterError`, and `ProviderInternalError` to handle provider-specific and generic errors consistently across services. [[1]](diffhunk://#diff-33fd96bf8309be33cda840428fae32e5b13a939f5398a9623f43364fbdb00c66R11-R33) [[2]](diffhunk://#diff-ccad427b766b44f9d0a43e67fe33ef295adb3f8b4ba44c5f08fd8fcf8c0d2f6fR4-R9) [[3]](diffhunk://#diff-973dd59252d7f0b82135e785f91c1d382a93163374de9c9081735c7e470acd73R2-R7) [[4]](diffhunk://#diff-39168a53b1f13d8d4ad133ca169ded6a8904c9e7a119f8e6e961e51761260872R2-R7)
* **Service-Level Enhancements**: Updated `AlchemyPay`, `Moonpay`, and `Transak` services to use the new error types, log errors, and handle edge cases like JSON parsing failures and missing fields in responses. [[1]](diffhunk://#diff-ccad427b766b44f9d0a43e67fe33ef295adb3f8b4ba44c5f08fd8fcf8c0d2f6fL104-R171) [[2]](diffhunk://#diff-973dd59252d7f0b82135e785f91c1d382a93163374de9c9081735c7e470acd73L46-R117) [[3]](diffhunk://#diff-39168a53b1f13d8d4ad133ca169ded6a8904c9e7a119f8e6e961e51761260872L59-R131)

### Input Validation
* **Centralized Validation**: Added `validateBundledPriceInput` middleware to validate query parameters for the new bundled endpoint, ensuring proper input before processing.
* **Improved Error Messages**: Validation errors now include detailed messages about supported currencies and required parameters.

### API Refactoring
* **Simplified Controller Logic**: Removed redundant error-catching logic in provider handlers, allowing errors to propagate naturally for centralized handling.
* **Response Consistency**: Ensured all endpoints return structured and meaningful error responses, including HTTP status codes and error messages.

### Miscellaneous
* **Cache Key Update**: Added a new cache key `allPrices` to support caching for the bundled endpoint in the frontend.


api/src/api/controllers/price.controller.ts
```diff
@@ -99,14 +86,127 @@ export const getPriceForProvider: RequestHandler<unknown, any, unknown, PriceQue
...
+      return { provider, status: 'rejected', reason: err } as const;
+    }
+  });
+
+  // Use Promise.allSettled to wait for all promises, regardless of success/failure
+  const results = await Promise.allSettled(pricePromises);
+
+  const response: PriceEndpoints.AllPricesResponse = {};
+
+  results.forEach((result) => {
```
Sharqiewicz, line 163: What do you think about changing `.forEach` to `.map` to indicate that we don't mutate the content of the `results` array? I think it is clearer and more understandable

api/src/api/services/alchemypay/alchemypay.service.ts
```diff
@@ -101,24 +107,79 @@ async function priceQuery(
...
+  let response: Response;
+  try {
+    response = await fetch(requestUrl, request);
+  } catch (fetchError) {
+    console.error('AlchemyPay fetch error:', fetchError);
+    throw new ProviderInternalError(`Network error fetching price from AlchemyPay: ${(fetchError as Error).message}`);
+  }
+
+  let body: AlchemyPayResponse;
+  try {
```
Sharqiewicz, line 119: I'm wondering, we use this let/try/catch pattern in many places in our app. I really don't like using `let` because the type can be reassigned and doesn't impose immutability. I think we should introduce a helper function like
```
type Success<T> = {
  data: T;
  error: null;
};

type Failure<E> = {
  data: null;
  error: E;
};

type Result<T, E = Error> = Success<T> | Failure<E>;

export async function tryCatch<T, E = Error>(
  promise: Promise<T>,
): Promise<Result<T, E>> {
  try {
    const data = await promise;
    return { data, error: null };
  } catch (error) {
    return { data: null, error: error as E };
  }
}
```
to make the try/catch handling easier and clearer (maybe in another PR?)

```
const body = await tryCatch(response.json());
```

api/src/api/services/transak.service.ts
```diff
@@ -56,20 +62,73 @@ async function priceQuery(
...
   }
 
+  if (!response.ok || body.error) {
+    const errorMessage = body?.error?.message || `HTTP error ${response.status}: ${response.statusText}`;
+    console.error(`Transak API Error (${response.status}): ${errorMessage}`);
+
+    // Analyze error message for specific types
+    const lowerErrorMessage = errorMessage.toLowerCase();
+    if (
+      lowerErrorMessage.includes('invalid fiat currency') ||
```
Sharqiewicz, line 92: We could use here an enum instead of directly passing the strings, wdyt?

api/src/api/controllers/price.controller.ts
ebma: Hmm not sure. To me, `.forEach` doesn't indicate mutation. On the other hand `.map` indicates to me that I want to map results to something else and assign the new transformed value to a variable. Which we don't do here.

api/src/api/services/alchemypay/alchemypay.service.ts
ebma: I'm fine with refactoring it. We can refactor it using the [neverthrow](https://www.npmjs.com/package/neverthrow/v/2.7.0) package.

api/src/api/services/transak.service.ts
ebma: True we could. Not sure if this would improve the readability though. We have to assume that all the code in `transak.service.ts`, `moonpay.service.ts` and `alchemypay.service.ts` is completely tailored to each service provider and we can't reuse code anyway. So maybe we just live with these strings in those files for now?

frontend/src/sections/FeeComparison/FeeProviderRow/index.tsx
```diff
@@ -49,39 +49,38 @@ export function FeeProviderRow({
...
-    isLoading,
-    error,
-    data: providerPriceRaw,
-  } = useQuery({
-    queryKey: [amount, sourceAssetSymbol, targetAssetSymbol, vortexPrice.toString(), provider.name, selectedNetwork],
-    queryFn: () => provider.query(sourceAssetSymbol, targetAssetSymbol, amount, selectedNetwork),
-    retry: false,
-  });
+  // Determine if there's an error from the result
+  const error = result?.status === 'rejected' ? result.reason : undefined;
```
gianfra-t, line 53: There shouldn't be an error because we filter them on the `const providerPrices = useMemo(()` function right? Still, don't mind if we leave it as redundancy for a future change.

ebma: No, there could be an error. The check/filtering happens on a different object that is used to show the best provider. The row itself gets an 'unchecked' response and it could contain an error.

PR #604, opened 2025-04-29
[WIP] Set a mark up fee for vortex

### LINKED ISSUE #552: Set a mark-up fee for Vortex
As business owner, i want to set a transparent mark up as part of the Vortex “UI”-Fee that ends up with Satoshipay

## Todo
* Add a markup fees of 1 BPS on the vortex quote
* show this fees transparently in the details section of the vortex app

## Note
Account to deposit - 6emGJgvN86YVYj5jENjfoMfEvX5p8hMHJGSYPpbtvHNEHTgy

ebma:
Closes #552.

### DIFF
--- a/api/src/api/services/ramp/quote.service.ts
+++ b/api/src/api/services/ramp/quote.service.ts
@@ -14,6 +14,8 @@ import {
 } from 'shared';
 import { BaseRampService } from './base.service';
 import QuoteTicket, { QuoteTicketMetadata } from '../../../models/quoteTicket.model';
+import Partner from '../../../models/partner.model';
+import FeeConfiguration from '../../../models/feeConfiguration.model';
 import logger from '../../../config/logger';
 import { APIError } from '../../errors/api-error';
 import { getTokenOutAmount } from '../nablaReads/outAmount';
@@ -22,7 +24,8 @@ import { calculateTotalReceive, calculateTotalReceiveOnramp } from '../../helper
 import { createOnrampRouteParams, getRoute } from '../transactions/squidrouter/route';
 import { parseContractBalanceResponse, stringifyBigWithSignificantDecimals } from '../../helpers/contracts';
 import { MOONBEAM_EPHEMERAL_STARTING_BALANCE_UNITS, MOONBEAM_EPHEMERAL_STARTING_BALANCE_UNITS_ETHEREUM } from '../../../constants/constants';
-import { multiplyByPowerOfTen } from '../../services/pendulum/helpers';
+import { multiplyByPowerOfTen } from '../pendulum/helpers';
+
 /**
  * Trims trailing zeros from a decimal string, keeping at least two decimal places.
  * @param decimalString - The decimal string to format
@@ -68,9 +71,26 @@ export class QuoteService extends BaseRampService {
   /**
    * Create a new quote
    */
-  public async createQuote(request: QuoteEndpoints.CreateQuoteRequest): Promise<QuoteEndpoints.QuoteResponse> {
+  public async createQuote(request: QuoteEndpoints.CreateQuoteRequest & { partnerId?: string }): Promise<QuoteEndpoints.QuoteResponse & {
+    networkFee: string;
+    processingFee: string;
+    partnerMarkupFee: string;
+    feeCurrency: string;
+  }> {
     // Validate chain support
     this.validateChainSupport(request.rampType, request.from, request.to);
+
+    // Validate and get partner if provided
+    let partner = null;
+    if (request.partnerId) {
+      partner = await Partner.findOne({
+        where: {
+          id: request.partnerId,
+          isActive: true,
+        },
+      });
+      // If partnerId was provided but not found or not active, we'll proceed without a partner
+    }
 
     // Calculate output amount (this would typically involve calling an external service or using a formula)
     const outputAmount = await this.calculateOutputAmount(
@@ -80,6 +100,16 @@ export class QuoteService extends BaseRampService {
       request.rampType,
       request.from,
       request.to,
+    );
+
+    // Calculate fee components
+    const feeComponents = await this.calculateFeeComponents(
+      request.inputAmount,
+      outputAmount.outputAmountBeforeFees,
+      request.rampType,
+      request.from,
+      request.to,
+      partner,
     );
 
     // Validate that the output amount is positive after fees
@@ -100,7 +130,12 @@ export class QuoteService extends BaseRampService {
       inputCurrency: request.inputCurrency,
       outputAmount: outputAmount.receiveAmount,
       outputCurrency: request.outputCurrency,
-      fee: outputAmount.fees,
+      networkFee: feeComponents.networkFee,
+      processingFee: feeComponents.processingFee,
+      partnerMarkupFee: feeComponents.partnerMarkupFee,
+      totalFee: feeComponents.totalFee,
+      feeCurrency: feeComponents.feeCurrency,
+      partnerId: partner?.id || null,
       expiresAt: new Date(Date.now() + 10 * 60 * 1000), // 10 minutes from now
       status: 'pending',
       metadata: {
@@ -118,7 +153,11 @@ export class QuoteService extends BaseRampService {
       inputCurrency: quote.inputCurrency,
       outputAmount: trimTrailingZeros(quote.outputAmount),
       outputCurrency: quote.outputCurrency,
-      fee: trimTrailingZeros(quote.fee),
+      fee: trimTrailingZeros(quote.totalFee), // Keep the original fee field for backward compatibility
+      networkFee: trimTrailingZeros(quote.networkFee),
+      processingFee: trimTrailingZeros(quote.processingFee),
+      partnerMarkupFee: trimTrailingZeros(quote.partnerMarkupFee),
+      feeCurrency: quote.feeCurrency,
       expiresAt: quote.expiresAt,
     };
   }
@@ -126,7 +165,12 @@ export class QuoteService extends BaseRampService {
   /**
    * Get a quote by ID
    */
-  public async getQuote(id: string): Promise<QuoteEndpoints.QuoteResponse | null> {
+  public async getQuote(id: string): Promise<(QuoteEndpoints.QuoteResponse & {
+    networkFee: string;
+    processingFee: string;
+    partnerMarkupFee: string;
+    feeCurrency: string;
+  }) | null> {
     const quote = await this.getQuoteTicket(id);
 
     if (!quote) {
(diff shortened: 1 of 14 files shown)


api/src/api/controllers/quote.controller.ts
```diff
@@ -17,7 +17,7 @@ export const createQuote = async (
   next: NextFunction,
 ): Promise<void> => {
   try {
-    const { rampType, from, to, inputAmount, inputCurrency, outputCurrency } = req.body;
+    const { rampType, from, to, inputAmount, inputCurrency, outputCurrency, partnerId } = req.body;
```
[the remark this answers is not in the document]
ebma: It is marked as optional in the respective type.

api/src/api/services/phases/phase-processor.ts
```diff
@@ -99,10 +99,17 @@ export class PhaseProcessor {
     }
 
     // Try to acquire the lock
-    const lockAcquired = await this.acquireLock(state);
+    let lockAcquired = await this.acquireLock(state);
     if (!lockAcquired) {
       if (this.isLockExpired(state)) {
         logger.info(`Lock for ramp ${rampId} has expired. Ignoring previous lock and continue processing...`);
+        // Force release the expired lock and try to acquire it again
+        await this.releaseLock(state);
```
gianfra-t, line 107: Now that we're touching this part again, do you think we can say for sure that for a full ramp, 15 minutes is enough to assume a crash and deadlock? Maybe we should increase this or change the logic.
I do agree with this changes though.

api/src/api/services/ramp/quote.service.ts
```diff
@@ -217,21 +537,24 @@ export class QuoteService extends BaseRampService {
       const outputTokenPendulumDetails =
         rampType === 'on' ? getPendulumDetails(outputCurrency, toNetwork) : getPendulumDetails(outputCurrency);
 
-      const inputAmountAfterFees =
-        rampType === 'on' ? calculateTotalReceiveOnramp(new Big(inputAmount), inputCurrency) : inputAmount;
+      // Initialize networkFeeUSD with '0'
+      let networkFeeUSD = '0';
+
+      // Use the original input amount directly for the swap
+      const inputAmountUsedForSwap = inputAmount;
```
gianfra-t, line 544: I know we just discovered something different, but if the anchor fee is taken after mint during onramp, shouldn't `getTokenOutAmount` consider this? Like we did before.

api/src/api/services/ramp/quote.service.ts
```diff
@@ -75,53 +95,148 @@ export class QuoteService extends BaseRampService {
...
-        onrampInputAmountUnits: outputAmount.inputAmountAfterFees,
+        onrampOutputAmountMoonbeamRaw: outputAmountMoonbeamRaw,
+        onrampInputAmountUnits: inputAmountUsedForSwap,
+        grossOutputAmount, // Store the gross amount before fee deduction
+        usdFeeStructure,
       } as QuoteTicketMetadata,
     });
 
+    const responseFeeStructure: QuoteEndpoints.FeeStructure = {
+      network: trimTrailingZeros(networkFeeFiat),
```
gianfra-t, line 222: Maybe we can directly store the fee struct with this trailing applied, not a big deal though.

api/src/api/services/ramp/quote.service.ts
```diff
@@ -145,11 +269,204 @@ export class QuoteService extends BaseRampService {
...
+              let markupFeeComponent = new Big(0);
+              if (record.markupType === 'absolute') {
+                markupFeeComponent = new Big(record.markupValue);
+              } else {
+                markupFeeComponent = new Big(inputAmount).mul(record.markupValue);
+              }
+
+              const markupFeeComponentInFeeCurrency = await priceFeedService.convertCurrency(
+                markupFeeComponent.toString(),
+                inputCurrency,
```
gianfra-t, line 330: I think we could use `markupCurrency` from the entry. Or if it will always be the input currency then we don't need that in the model.

api/src/api/services/transactions/offrampTransactions.ts
```diff
@@ -33,6 +36,488 @@ import { preparePendulumCleanupTransaction } from './pendulum/cleanup';
...
+
+  // Convert USD fees to stablecoin raw units
+  const networkFeeStablecoinRaw = convertUSDToTokenUnits(networkFeeUSD, { decimals: stablecoinDecimals });
+  const vortexFeeStablecoinRaw = convertUSDToTokenUnits(vortexFeeUSD, { decimals: stablecoinDecimals });
+  const partnerMarkupFeeStablecoinRaw = convertUSDToTokenUnits(partnerMarkupFeeUSD, { decimals: stablecoinDecimals });
+
+  const transfers = [];
+
+  if (new Big(networkFeeStablecoinRaw).gt(0)) {
+    transfers.push(api.tx.tokens.transferKeepAlive(vortexPayoutAddress, stablecoinCurrencyId, networkFeeStablecoinRaw));
```
gianfra-t, line 114: Isn't `networkFee` what we spend on squidrouter and other transfers?

api/src/api/services/transactions/onrampTransactions.ts
```diff
@@ -27,9 +30,461 @@ import { createNablaTransactionsForOnramp } from './nabla';
...
+    inputAmountUnits,
+    quote,
+    account,
+    inputTokenPendulumDetails,
+    outputTokenPendulumDetails,
+    outputTokenDetails,
+  } = params;
+
+  // Calculate the soft minimum output based on the gross output amount
+  const grossOutputAmountPendulumUnits = new Big(quote.metadata?.grossOutputAmount || '0');
```
gianfra-t, line 279: I think here we also have to consider the anchor fee. We can subtract it here or add it in the calculation of the `grossOutputAmount` when passing `inputAmountUsedForSwap` up there in the quote file, WDYT?

api/src/api/services/ramp/quote.service.ts
```diff
@@ -145,11 +269,204 @@ export class QuoteService extends BaseRampService {
...
+            name: partnerName,
+            rampType: rampType,
+            isActive: true,
+          },
+        });
+
+        if (partnerRecords.length > 0) {
+          let hasApplicableFees = false;
+
+          for (const record of partnerRecords) {
```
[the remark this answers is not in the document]
ebma: Yes, I considered that partners might want to define both an absolute and a relative fee. We know that Anclap are doing that, maybe our future partners want to do it as well.

api/src/api/services/ramp/quote.service.ts
ebma: True, we could. I'll change it.

api/src/api/services/transactions/offrampTransactions.ts
ebma: Yes it is. But since we are subsidizing it with our own money, we are sending the fee we charge in the respective USD-stablecoin to the vortex payout address. We don't charge this fee on Moonbeam or Assethub but all fees are charged on Pendulum.

Or where you pointing at something else with your question?

api/src/api/services/phases/phase-processor.ts
ebma: You are right but I think we should consider what can possibly go wrong here. Which phases can take longer than expected? Mostly transaction submissions. 
For example, a non-expressed squidrouter transaction could take 30 minutes. In that case, the first process will eventually timeout when waiting. After 15 minutes a recovery process will try again and probably check the balance and try resubmitting the transaction. Possibly still failing. Then a third time the recovery will start. This time, the balance check would succeed and it can continue. 

Duplicate transaction submissions are not a problem as we are always submitting the _same_ transaction, if any. 
Phases like the funding of ephemerals or subsidy would be problematic if they happen multiple times. However, these phases would rarely take longer than a minute, so we should be fine here too.

I would keep the lock and recovery low to improve the UX. The more we retry, the faster the user will potentially be unblocked with their ramp.

api/src/api/services/ramp/quote.service.ts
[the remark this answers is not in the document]
gianfra-t: Sounds good! I was thinking that another option would be to have all in the same row and disallow > 1 entry for each partner, but since this is already done this is also good.

api/src/api/services/transactions/offrampTransactions.ts
gianfra-t: Yes I was just confused why we are sending it back to us, but I get it now. We subsidize it in GLMR or whatever, and extract the equivalent in the fee token to compensate.

api/src/api/services/phases/phase-processor.ts
gianfra-t: So we just need to be very sure that no ramp locks in place for more than 15 minutes. But this is not really phase related but rather ramp related, right? We don't have an overall phase executor that will timeout, so although one single phase can't take more than 15 minutes without timing out this is not explicit, and the whole ramp may actually take that much.

But to be fair, I don't have an alternative approach in mind for this... if we have locks then it is possible to have this deadlock on crash, so we need this. What if we just increase the expire lock time? It won't affect the UX since this will already be an edge case but we may be safer on incorrectly flagged dead processes. Alternative we may add a timeout on the recursive `processPhase` that ensures after <15 minutes the ramp process cannot be locked unless it crashed.

api/src/api/services/priceFeed.service.ts
```diff
@@ -0,0 +1,360 @@
...
+  }
+
+  /**
+   * Get the exchange rate from USD to another fiat currency. The source currency is always USD.
+   *
+   * @param toCurrency - The target currency code (e.g., 'BRL', 'ARS')
+   * @param inputAmount - The amount to convert (default is '1.0')
+   * @returns The exchange rate (how much of toCurrency equals 1 unit of fromCurrency)
+   */
+  public async getFiatExchangeRate(toCurrency: RampCurrency, inputAmount = '1.0'): Promise<number> {
```
Sharqiewicz, line 168: Perhaps we should consider renaming the function to better reflect its purpose—specifically, that it retrieves the exchange rate from USD to another fiat currency, with USD always as the source currency. For example, we could use a name like `getUSDToFiatExchangeRate`.

api/src/api/services/priceFeed.service.ts
```diff
@@ -0,0 +1,360 @@
...
+    toCurrency = toCurrency.toLowerCase() as RampCurrency;
+    try {
+      // If currencies are the same, return the original amount
+      if (fromCurrency === toCurrency) {
+        logger.debug(`Currencies are the same (${fromCurrency}), returning original amount: ${amount}`);
+        return amount;
+      }
+
+      // Check if both currencies are USD-like stablecoins
+      const isUsdLikeCurrency = (currency: RampCurrency): boolean =>
```
Sharqiewicz, line 283: Perhaps we could use the existing `EVMTokens` enum, or create a new one such as `USDLikeEVMTokens`, and iterate through this enum here. This way, we would have a single source of truth for these tokens, making them easier to manage and reuse throughout the entire app.

```
export declare enum EvmToken {
    USDC = "usdc",
    USDT = "usdt",
    USDCE = "usdce"
}
```

or

```
export declare enum USDLikeEvmToken {
    USDC = EvmToken.USDC,
    USDT = EvmToken.USDT,
    USDCE = EvmToken.USDCE
}
```

api/src/api/services/priceFeed.service.ts
```diff
@@ -0,0 +1,360 @@
...
+        currency === EvmToken.USDC || currency === EvmToken.USDT || currency === EvmToken.USDCE;
+
+      if (isUsdLikeCurrency(fromCurrency) && isUsdLikeCurrency(toCurrency)) {
+        logger.debug(
+          `Both currencies are USD-like (${fromCurrency} -> ${toCurrency}), using 1:1 conversion for: ${amount}`,
+        );
+        return amount;
+      }
+
+      // USD -> Fiat conversion
```
Sharqiewicz, line 293: Maybe we could extract each of these cases into separate functions to keep the case-specific logic isolated. The current function is already quite large, and this refactor would help simplify it and reduce cognitive load for anyone reading or maintaining the code.

For example, we could have something like

```
if (isUsdLikeCurrency(fromCurrency) && isUsdLikeCurrency(toCurrency)) {
        return this.convertUsdLikeToUsdLike(amount, fromCurrency, toCurrency);
      }

      if (isUsdLikeCurrency(fromCurrency) && isFiatToken(toCurrency)) {
        return this.convertUsdToFiat(amount, toCurrency, decimals);
      }

      if (isFiatToken(fromCurrency) && isUsdLikeCurrency(toCurrency)) {
        return this.convertFiatToUsd(amount, fromCurrency, decimals);
      }

      if (isUsdLikeCurrency(fromCurrency) && !isFiatToken(toCurrency) && !isUsdLikeCurrency(toCurrency)) {
        return this.convertUsdToCrypto(amount, toCurrency, decimals);
      }

      if (!isFiatToken(fromCurrency) && !isUsdLikeCurrency(fromCurrency) && isUsdLikeCurrency(toCurrency)) {
        return this.convertCryptoToUsd(amount, fromCurrency, decimals);
      }
      ```

Sharqiewicz (commented):
It looks good!

PR #614, opened 2025-05-02
Refactor FeeComparison to conditionally render based on ramp direction

ebma:
Add logic to hide the fee comparison section for the ONRAMP direction. Ensures the section only displays when relevant, improving user experience and layout consistency.


gianfra-t (approved):
👍

Issue #620, opened 2025-05-05
Improve reliability of onramp


ebma:
## Context

Currently, it can sometimes happen that multiple onramps where a user confirmed that they made a payment interfere with each other. This is because there is no check linking payments to ramps. 

## TODO

- Use a reference number contained in the PIX pay-in to link the transaction to a particular ramp
- Add a timeout to the brlaTeleport phase so that the ramp enters the 'failed' state if the payment was not received within 30 minutes.

PR #631, opened 2025-05-09
Improve pix onramp payment handling

gianfra-t:
Issue: #620.

### Change description

- Adds a reference label check to the pix payment. Reference is derived as the first 8 characters of the quote id used.
- Force the ramp to a `failed` state after 30 minutes.
- Teleport service now supports deleting an entry, which will be used when the ramp is declared a `failed` one. Additionally, when requesting, a teleport will not be added if existing already in the set.

### Additional changes

- Increase delay time to 30 seconds between teleport balance check and starting `moonebamToPendulumXcm` operation.


api/src/api/services/phases/handlers/brla-teleport-handler.ts
```diff
@@ -57,25 +62,42 @@ export class BrlaTeleportPhaseHandler extends BasePhaseHandler {
...
-        if (balanceCheckError.message === 'Balance did not meet the limit within the specified time') {
-          throw new Error(`BrlaTeleportPhaseHandler: balanceCheckError ${balanceCheckError.message}`);
-        } else {
-          logger.error('Error checking Moonbeam balance:', balanceCheckError);
-          throw new Error(`BrlaTeleportPhaseHandler: Error checking Moonbeam balance`);
-        }
+      if (!(balanceCheckError instanceof Error)) throw balanceCheckError;
+      const { message } = balanceCheckError;
+    
+      const isCheckTimeout = message.includes('did not meet the limit within the specified time');
```
ebma, line 76: This check seems quite brittle to me, any chance we can refactor to make it less reliant on the exact wording of the error message?

api/src/api/services/phases/handlers/brla-teleport-handler.ts
```diff
@@ -57,25 +62,42 @@ export class BrlaTeleportPhaseHandler extends BasePhaseHandler {
...
+    
+      throw isCheckTimeout
+        ? this.createRecoverableError(`BrlaTeleportPhaseHandler: ${message}`)
+        : new Error(`Error checking Moonbeam balance: ${message}`);
     }
 
     return this.transitionToNextPhase(state, 'moonbeamToPendulumXcm');
   }
+
+  protected finalPaymentTimeout(state: RampState): boolean {
```
ebma, line 90: I think the name is not great. It returns a `boolean` so maybe we make it start with `is`. How about
```suggestion
  protected isPaymentTimeoutReached(state: RampState): boolean {
```

ebma (commented):
Looks good to me overall, cool that only the few changes are necessary to make this work. I suppose you still need to test whether it actually works @gianfra-t?

api/src/api/services/phases/handlers/brla-teleport-handler.ts
gianfra-t: Yes, we can start implementing some better practices and return custom errors probably, not only here but everywhere...

gianfra-t:
@ebma yes only testing, and also removing the entry from the teleport service was missing. I decided to use the quote id instead of the address encoding to have more control over each specific ramp and the associated teleport operation.

For testing, well I think the best would be to merge and test on staging a true ramp, I don't think this can be mocked. We just need to be sure we are encoding correctly the reference on the br code.

PR #632, opened 2025-05-12
fix multiple progress page animation intervals causing % above 100

Sharqiewicz:
In order to avoid issues with percentage shown in the Progress page animation:
- [x] Ensuring that only one interval is active
- [x] Remove use of ref for following current percentage to avoid rendering errors


ebma:
@Sharqiewicz can you please briefly describe in the description of your PR what the issue is? Otherwise I wouldn't really know what to pay attention to for the review.

ebma (approved):
I didn't test it but it looks good ✅

PR #635, opened 2025-05-12
[READY] filter out Nova Wallet option if not in the Nova app

Sharqiewicz:
Show Nova Wallet connection option only when dApp is open in Nova app.

### DIFF
--- a/frontend/src/components/BrlaComponents/KYCForm/useKYCFormLocalStorage.ts
+++ b/frontend/src/components/BrlaComponents/KYCForm/useKYCFormLocalStorage.ts
@@ -1,0 +1,50 @@
+import { useEffect, useCallback } from 'react';
+import { Path, PathValue, UseFormReturn } from 'react-hook-form';
+import { debounce } from '../../../hooks/useLocalStorage';
+import { useTaxId } from '../../../stores/ramp/useRampFormStore';
+
+export const BRLA_KYC_FORM_STORAGE_KEY = 'brla_kyc_form_data';
+
+export const useKYCFormLocalStorage = <T extends object>(form: UseFormReturn<T>) => {
+  const taxId = useTaxId();
+
+  const { watch, setValue } = form;
+
+  const saveToStorage = debounce((data: T) => {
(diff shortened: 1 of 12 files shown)


ebma (approved):
Let's address the nit from copilot but rest looks good to me ✅

frontend/src/components/BrlaComponents/KYCForm/useKYCFormLocalStorage.ts
```diff
@@ -0,0 +1,50 @@
...
+            setValue(key as Path<T>, value as PathValue<T, Path<T>>, { shouldValidate: true });
+          }
+        });
+      } catch (error) {
+        console.error('Error loading form data from localStorage:', error);
+      }
+    } else {
+      localStorage.setItem(BRLA_KYC_FORM_STORAGE_KEY, JSON.stringify({ taxId }));
+    }
+  }, [setValue, taxId]);
```
gianfra-t, line 34: Maybe we could ignore here reacting to `taxId` and setting? I think the effect bellow (with the subscription) should handle already setting the `taxId`, or do we need both effects for different scenarios?

Sharqiewicz: The first useEffect is used for the form initialization with the data stored in the LocalStorage. I see it might look confusing so I reorganized the code in a better way now to reduce the cognitive load. let me know if it looks better for you now

gianfra-t: Yes I was just confused about the else branch of the first one (where the storage is set), and the second `useEffect` that is also subscribed to changes to the form via `watch`. But I see now that the second one will set the storage not only when `savedData` doesn't exists but on any refresh of the form. Nevermind then, let's go!

Sharqiewicz:
@pendulum-chain/devs Can I merge?

ebma (approved):
I tested a bit and seems to do what it's supposed to do. I think the localStorage logic is a bit scattered but I'm fine if we keep it that way.

PR #661, opened 2025-05-22
handle the response when quote exceeds the pool size

Sharqiewicz:


ebma (commented):
I think we should still allow changing the asset when the quote does not load. It could be that for a different output asset, we can get a quote. Otherwise the user would first have to edit their input amount which could be confusing again. 

We could also consider adding an error message.

ebma (commented):
I can now still change the output token which is great. We are still not telling the user what's wrong though. 
To keep it simple, how about we just check the response code instead of looking at the contained error? If it's a `500` we assume that the quote cannot be served because the input amount is too low. 


image


The ideal solution would be to change the responses of the backend to tell the frontend what exactly is wrong. This is a little trickier and maybe not worth it just for this. Your call @Sharqiewicz but we need to have at least some messaging telling the user what's wrong and what they can do.

ebma (approved):
Works, let's go ✅

PR #665, opened 2025-05-22
Add unhandled payment alert worker

gianfra-t:
### About
Adds a worker that fetches ramp entries with status `failed` and `initial`, using a predefined time window. 

For each ramp, the worker checks:
- If there was a payment with a reference label corresponding to the entry. This should not be the case and should be alerted.
- Checks if there is a duplicate reference label on the list of payments from the BRLA account corresponding to the entry. Rationale is that this may be the cause of a `failed` or `invalid` state.
- If there is a payment with an invalid label (no label, wrong length, etc).

The worker will not alert more than once per day for each`subaccountId`. If from a particular ramp state we detect multiple inconsistencies, then all will be triggered.


Sharqiewicz (approved):
LGTM ✅

Issue #673, opened 2025-05-28
Add CPF field to the KYC form


prayagd:
## Context
Currently when the user enters CPF and proceeds to do KYC, if the user sees the error Incorrect CPF. They need to navigate back to the first page to change the CPF

## Todo
* Add CPF field to the KYC form, which should be prefilled with the same CPF that was captured on the first page
* If the CPF is incorrect they should be able to edit the CPF field on the KYC page itself

prayagd:
cc @pendulum-chain/devs

prayagd:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Pendulum-6408a0365cda59c57d39bee5/issues/gh/pendulum-chain/vortex/673?planning-poker) with Zenhub @ebma @gianfra-t @Sharqiewicz

Issue #630, opened 2025-05-09
Improve KYC level 2 handling.


gianfra-t:
To improve the error handling of the document uploads, we can implement the following slight changes to the KYC level 2 flow for BRLA:

Change the validation screen to not automatically go back to the swap component upon validation failure or acceptance. For the failure case, we can show the user a validation error feedback and the possibility to retry. So far we can at least differentiate between a selfie and a document error, for which the user can retry and adjust.

For the success case, it could be beneficial for the user to show this screen until clicked away. The rationale is that a long waiting time for the KYC result will probably make the user look away from the screen and the success event may be missed.

PR #677, opened 2025-05-28
Improve kyc level 2 ux

gianfra-t:
Issues: #630 and #673.

### Changes
- Adds a new `taxId` field to KYC level 1 form that is connected to the main `taxId` state. If `taxId` is detected invalid, it displays a message on the field and allows for change.
- Validation screens for both levels do not close automatically anymore, and require user interaction. For accepted and for rejected, showing the error message for the later.

### DIFF
--- a/apps/frontend/src/hooks/brla/useBRLAKYCProcess/index.tsx
+++ b/apps/frontend/src/hooks/brla/useBRLAKYCProcess/index.tsx
@@ -288,7 +288,6 @@ export function useKYCProcess() {
   useEffect(() => {
     if (!kycResponse) return;
     if (kycResponse.level !== 2) return;
-    console.log('KYC level 2 response:', kycResponse);
     const handleStatus = async (status: string) => {
       const mappedStatus = status as KycStatus;
 


frontend/src/hooks/brla/useBRLAKYCProcess/index.tsx
```diff
@@ -295,18 +322,37 @@ export function useKYCProcess() {
   ]);
 
   useEffect(() => {
-    if (error) {
-      handleError(error.message);
+    const threshold = 60000;
```
gianfra-t, line 325: This effect is for not failing immediately if a connection error happens. But > 60 seconds it will in fact show a network error failure.
We can adjust this threshold.

ebma (commented):
Great stuff 👌 

When testing locally using the BRLA sandbox environment, I did the following:
- Generated a random CPF and changed the last digit to a different one. 
- I was able to proceed to the KYC form 1 (totally fine)
- I filled in all the other fields with dummy data and proceeded by clicking on the button.
- I was quickly taken back to KYC form 1 telling me that my CPF was invalid. 
Perfect flow, I love it. 

I noticed you changed the form field validation to trigger `onChange` and not `onBlur`. Is there a reason for this change? I think `onBlur` is better because otherwise the user would often times see an error already when they try to fill in the field for the first time. 

I noticed one bug though. Once I used a 'valid' dummy CPF, the KYC verification for level 1 succeeded, at least the UI told me so, and I clicked on the 'Confirm' button. I was taken back to the base swap page and immediately presented with the ramp summary dialog. All good, so far. Now the proplem is that I never saw a pix key and the button stays at processing because apparently the registration of the ramp failed because the subaccount does not exist. This doesn't really make sense because in the KYC verification, the subaccount must have been created a very long time ago. My assumption would be that for some reason, not an up-to-date subacount or taxID was used, maybe the invalid one I originally entered with the other digit. I only corrected it to a 'correct' CPF on the KYC form 1 and not the base swap page, so maybe it was not correctly applied to the overall state again? It worked fine after I closed and reopened the ramp summary dialog again. 
image

apps/frontend/src/hooks/brla/useBRLAKYCProcess/index.tsx
```diff
@@ -255,21 +288,16 @@ export function useKYCProcess() {
   useEffect(() => {
     if (!kycResponse) return;
     if (kycResponse.level !== 2) return;
-
+    console.log('KYC level 2 response:', kycResponse);
```
ebma, line 291: Nit: remove
```suggestion
```

ebma (approved):
I didn't test it again but looks good to me 👍

Issue #654, opened 2025-05-19
Query transaction while execution and top-up the fees if needed


prayagd:
## Context
While the squid transaction is executed, if there is a delta between the quoted gas fee and actual fee. This solution focuses on increasing the fee if needed for the transaction to execute

## Todo
* Query the fees needed while the transaction is been executed
* If higher fees needed top-up the contract from the funding account

## Note
* This is a long term fix

prayagd:
cc @pendulum-chain/devs

PR #681, opened 2025-05-30
Add new axelar gas phase.

### LINKED ISSUE #668: Add checks for squidrouter bridge when onramping.
### Description

Currently we are not checking if the final Squidrouter execution has actually delivered the funds or not to the user's account on the destination chain.
A relatively common failure mode (connected to fix #652) happens when the Axelar transaciton runs out of gas. From the perspective of our backend, the onramp is declared `complete` yet some action must be taken.

### Potential improvement

We should check that the funds have arrived at the destination before setting the ramp to `complete`. Options are:
- Check the status of the Axelar bridge. 
- Check the `in` transactions on the destination wallet.

gianfra-t:
Closes #668 and closes https://github.com/pendulum-chain/vortex/issues/654. 

### About

Changes fundamentally the logic of how we pay for Squidrouter's underlying Axelar transaction, in order to:
- Have a precise quote of the fee needed at the time of the execution of the bridge. This prevents unforeseen insufficient gas errors and overpaying.
- Avoid funding Moonbeam ephemerals with too much GLMR.

### Changes
- Adds a new phase for onramps, `squidrouterPay`, and corresponding handler.
- Reduces all GLMR funding constants.
- Removes checks in quote and transaction building about sufficient ephemeral funding regarding squidroute payment in the last phase.
- Ramps are considered completed only if the status endpoint specifies an execution operation.

As bonus, we don't treat `failed` transactions anymore on the recovery worker.


api/src/api/services/phases/handlers/squid-router-pay-phase-handler.ts
```diff
@@ -0,0 +1,257 @@
...
+    };
+  };
+
+  id: string; // the id of the swap.
+}
+
+const AXELAR_POLLING_INTERVAL_MS = 10000; // 10 seconds
+const AXL_GAS_SERVICE_MOONBEAM = '0x2d5d7d31F671F86C782533cc367F14109a082712';
+/**
+ * Handler for the squidRouter pay phase. Checks the status of the Axelar bridge and pays on native GLMR fee if.
```
ebma, line 65: if what?

api/src/api/services/phases/handlers/squid-router-pay-phase-handler.ts
```diff
@@ -0,0 +1,257 @@
...
+    destination_express_fee: {
+      total: number;
+    };
+  };
+
+  id: string; // the id of the swap.
+}
+
+const AXELAR_POLLING_INTERVAL_MS = 10000; // 10 seconds
+const AXL_GAS_SERVICE_MOONBEAM = '0x2d5d7d31F671F86C782533cc367F14109a082712';
```
ebma, line 63: This is the address of the contract that we need to call to fill up the gas I assume?

api/src/api/services/phases/handlers/squid-router-pay-phase-handler.ts
```diff
@@ -0,0 +1,257 @@
...
+  /**
+   * Execute the phase
+   * @param state The current ramp state
+   * @returns The updated ramp state
+   */
+  protected async executePhase(state: RampState): Promise<RampState> {
+    logger.info(`Executing squidRouterPay phase for ramp ${state.id}`);
+
+    if (state.type === 'off') {
+      logger.info(`SquidRouter phase is not supported for off-ramp`);
```
ebma, line 98:
```suggestion
      logger.info(`squidRouterPay phase is not supported for off-ramp`);
```

api/src/api/services/phases/handlers/squid-router-pay-phase-handler.ts
```diff
@@ -0,0 +1,257 @@
...
+    const moonbeamExecutorAccount = privateKeyToAccount(MOONBEAM_FUNDING_PRIVATE_KEY as `0x${string}`);
+    const { walletClient } = createMoonbeamClientsAndConfig(moonbeamExecutorAccount);
+    this.walletClient = walletClient;
+  }
+
+  /**
+   * Get the phase name
+   */
+  public getPhaseName(): RampPhase {
+    return 'squidrouterPay';
```
ebma, line 86: I think we are capitalizing like this most of the time. And so do they in their docs. 

```suggestion
    return 'squidRouterPay';
```

api/src/api/services/phases/handlers/squid-router-phase-handler.ts
```diff
@@ -83,13 +82,12 @@ export class SquidRouterPhaseHandler extends BasePhaseHandler {
         state: {
           ...state.state,
           squidRouterApproveHash: approveHash,
-          squidRouterSwapHash: swapHash,
+          squidrouterSwapHash: swapHash,
```
ebma, line 85: We are getting inconsistent here, let's make it `squidRouter` everywhere and not `squidrouter`.

api/src/api/services/ramp/quote.service.ts
ebma, line 1: Did you already merge `staging` into this branch? If so, I think something went wrong.

api/src/constants/constants.ts
```diff
@@ -7,8 +7,8 @@ const SUBSIDY_MINIMUM_RATIO_FUND_UNITS = '5'; // 5 Subsidies considering maximum
 const MOONBEAM_RECEIVER_CONTRACT_ADDRESS = '0x2AB52086e8edaB28193172209407FF9df1103CDc';
 const STELLAR_EPHEMERAL_STARTING_BALANCE_UNITS = '2.5'; // Amount to send to the new stellar ephemeral account created
 const PENDULUM_EPHEMERAL_STARTING_BALANCE_UNITS = '0.1'; // Amount to send to the new pendulum ephemeral account created
-const MOONBEAM_EPHEMERAL_STARTING_BALANCE_UNITS = '30'; // Amount to send to the new moonbeam ephemeral account created
-const MOONBEAM_EPHEMERAL_STARTING_BALANCE_UNITS_ETHEREUM = '150'; // Amount to send to the new moonbeam ephemeral account created when onramping to Ethereum
+const MOONBEAM_EPHEMERAL_STARTING_BALANCE_UNITS = '1'; // Amount to send to the new moonbeam ephemeral account created
+const MOONBEAM_EPHEMERAL_STARTING_BALANCE_UNITS_ETHEREUM = '1'; // Amount to send to the new moonbeam ephemeral account created when onramping to Ethereum
```
ebma, line 11: Let's reduce it to just one constant then if they have the same value. We don't need to distinguish anymore (probably).

api/src/contracts/AxelarGasService.ts
```diff
@@ -0,0 +1,591 @@
+export const axelarGasService = [
```
ebma, line 1:
```suggestion
export const axelarGasServiceAbi = [
```

ebma (commented):
Looks good to me overall 👍  Let's check all the names of 'squidrouter' vs 'squidRouter' again and make it consistent (changing it to 'squidRouter'). 
Also, I think you need to double-check the quote.service file, or be careful when merging with staging. It doesn't exist anymore on staging as I split it up into an extra directory.

api/src/constants/constants.ts
gianfra-t: Agreed. I was also doubting if we should keep it as a legacy thing, or not.

api/src/api/services/phases/handlers/squid-router-pay-phase-handler.ts
gianfra-t: Let's maybe leave the phases like this, as we also use `squidrouterSwap` as phase. Or would you prefer to change that also?

api/src/api/services/phases/handlers/squid-router-pay-phase-handler.ts
gianfra-t: Haha. No condition. Was part of a larger comment.

PR #704, opened 2025-06-06
UI Changes

Sharqiewicz:
- [x] Hide PIX section when quote is expired.
- [x] Update default currencies and amounts
- [x] Change English flag
- [x] Add toast when PIX is copied
- [x] Fix Ramp Form animations and tooltips


ebma (approved):
Works overall, we just need to touch the default amount again, it's not properly applied for the EURC and ARS currency.

prayagd:
We also missed adding a CTA here on the summary dailog to click on the "I have made the payment" cc @pendulum-chain/devs

PR #731, opened 2025-06-19
Add monerium interactive

gianfra-t:


apps/api/src/api/services/phases/handlers/squid-router-pay-phase-handler.ts
```diff
@@ -95,12 +119,28 @@ export class SquidRouterPayPhaseHandler extends BasePhaseHandler {
    */
   private async checkStatus(state: RampState, swapHash: string): Promise<void> {
     try {
-      // Found to be unreliable. We call it anyway so that squidrouter can log the status.
-      getStatus(swapHash).catch(error => logger.error(`Couldn't fetch status for ${swapHash} from squidrouter`, error.message));
-
       let isExecuted = false;
       let payTxHash: string | undefined = state.state.squidRouterPayTxHash; // in case of recovery, we may have already paid.
+      // initial delay to allow for API indexing.
+      await new Promise(resolve => setTimeout(resolve, SQUIDROUTER_INITIAL_DELAY_MS));
```
gianfra-t, line 125: I found that there is always some noticeable delay between sending the transaction and this endpoint returning something.

apps/api/src/api/services/phases/handlers/squid-router-pay-phase-handler.ts
```diff
@@ -200,30 +321,51 @@ export class SquidRouterPayPhaseHandler extends BasePhaseHandler {
...
+    const sourceGasPriceBig = Big(feeResponse.source_token.gas_price);
+
+    // Calculate base execution fee: gasLimit * gasPrice
+    const executionFeeUnits = estimatedGasBig.mul(sourceGasPriceBig);
+
+    // Apply the gas multiplier.
+    const multiplier = feeResponse.execute_gas_multiplier;
+    const executionFeeWithMultiplier = executionFeeUnits.mul(multiplier);
+
+    // L1 data fee??
```
gianfra-t, line 358: This is taken from the implementation of `AxelarSdk` function to calculate native [fee](https://github.com/axelarnetwork/axelarjs-sdk/blob/main/src/libs/TransactionRecoveryApi/AxelarGMPRecoveryAPI.ts#L823).

There is an assumption taken for `estimatedGas` that may need to be adjusted.

apps/api/src/api/services/phases/handlers/squid-router-phase-handler.ts
```diff
@@ -27,6 +33,35 @@ export class SquidRouterPhaseHandler extends BasePhaseHandler {
...
+      if (!quote) {
+        throw new Error(`Quote not found for ramp ${state.id}`);
+      }
+
+      if (quote.inputCurrency === FiatToken.EURC) {
+        return this.polygonClient;
+      } else if (quote.inputCurrency === FiatToken.BRL) {
+        return this.moonbeamClient;
+      } else {
+        logger.info(
```
gianfra-t, line 54: Defensive, it is more likely we add onramps that go through moonbeam than polygon.

apps/api/src/api/services/transactions/moneriumEvmOfframpTransactions.ts
```diff
@@ -0,0 +1,137 @@
...
+  return '';
+}
+
+export interface MoneriumOfframpTransactionParams {
+  quote: QuoteTicketAttributes;
+  userAddress?: string;
+  moneriumAuthToken?: string;
+}
+
+export async function prepareMoneriumEvmOfframpTransactions({
```
gianfra-t, line 36: Offramp is not ready and will be completed on a different PR.

apps/frontend/src/hooks/offramp/useRampService/useRegisterRamp.ts
```diff
@@ -4,60 +4,51 @@ import {
...
+ * A utility hook to manage signature traces using localStorage.
+ * This prevents a process from running more than once.
  */
-const useProcessLock = (lockKey: string) => {
-  // Checks if a lock exists and returns true if the process can proceed
-  const checkLock = useCallback(() => {
-    const existingLock = localStorage.getItem(lockKey);
-    if (existingLock !== null) {
-      console.log(`Process for ${lockKey} already started, skipping...`);
+const useSignatureTrace = (traceKey: string) => {
```
gianfra-t, line 34: This is a slight renaming, the functionality is intended to stay similar but a bit simplified.

apps/frontend/src/hooks/ramp/useRampForm.ts
```diff
@@ -57,9 +57,9 @@ export const useRampForm = (): {
   const enforceTokenConstraints = useCallback(
     (token: FiatToken): FiatToken => {
       // For onramp, we only allow BRL
```
gianfra-t, line 59: I believe we could remove this constraint and only leave it to the selector to filter what tokens can be onramped/offramped.

apps/frontend/src/components/RampSummaryDialog/index.tsx
```diff
@@ -20,7 +20,7 @@ export const RampSummaryDialog: FC = () => {
   const { resetRampState } = useRampActions();
   const executionInput = useRampExecutionInput();
   const visible = useRampSummaryVisible();
-  const rampDirection = useRampDirection();
+  const rampDirection = executionInput?.quote.rampType === "off" ? RampDirection.OFFRAMP : RampDirection.ONRAMP;
```
ebma, line 23: Shouldn't these checks have one more condition? ie. if there is a quote on the execution input, use the ramp direction from there. If there is no quote on the execution input, use the directon from `useRampDirection()`. Actually you could also just adjust the `useRampDirection()` hook directly IMO.

apps/frontend/src/components/RampSummaryDialog/RampSummaryButton.tsx
```diff
@@ -33,21 +33,22 @@ interface UseButtonContentProps {
...
     const isOfframp = rampDirection === RampDirection.OFFRAMP;
-    const isBRCodeReady = Boolean(rampState?.ramp?.brCode);
+    const isDepositQrCodeReady = Boolean(rampState?.ramp?.depositQrCode);
 
     // BRL offramp has no redirect, it is the only with type moonbeam
     const isAnchorWithoutRedirect = toToken.type === "moonbeam";
     const isAnchorWithRedirect = !isAnchorWithoutRedirect;
 
-    if ((isOnramp && isBRCodeReady && isQuoteExpired) || (isOfframp && isQuoteExpired)) {
+    console.log("isQuoteExpired:", isQuoteExpired);
```
ebma, line 50: Nit: remove

apps/frontend/src/components/RampSummaryDialog/TransactionTokensDisplay.tsx
```diff
@@ -68,7 +70,7 @@ export const TransactionTokensDisplay: FC<TransactionTokensDisplayProps> = ({ ex
     if (targetTimestamp === null) {
       // If no valid timestamp, mark as expired immediately
       setTimeLeft({ minutes: 0, seconds: 0 });
-      setIsQuoteExpired(true);
+      //setIsQuoteExpired(true);
```
ebma, line 73: Why not?

apps/frontend/src/hooks/offramp/useRampService/useRegisterRamp.ts
ebma: Why are we touching this logic in the first place?

apps/frontend/src/hooks/offramp/useRampService/useRegisterRamp.ts
```diff
@@ -140,23 +122,22 @@ export const useRegisterRamp = () => {
       }
 
       if (!canRegisterRamp) {
+        console.log("Cannot register ramp, canRegisterRamp is false");
         throw new Error("Cannot proceed with ramp registration, canRegisterRamp is false");
       }
 
-      // Verify we still own the lock before proceeding
```
ebma, line 146: Why are we removing this check?

apps/frontend/src/hooks/offramp/useRampService/useRegisterRamp.ts
```diff
@@ -220,20 +233,23 @@ export const useRegisterRamp = () => {
         ephemeralTxs,
         executionInput.ephemerals,
         pendulumApiComponents.api,
-        moonbeamApiComponents.api
+        moonbeamApiComponents.api,
+        config.alchemyApiKey
```
ebma, line 237: I don't think we need to pass this here. The config can just be accessed in the other function, no?

gianfra-t: `signUnsignedTransactions` is in our shared package, I thought we couldn't. Do you if the configs are taken anyway?

apps/frontend/src/components/RampSummaryDialog/TransactionTokensDisplay.tsx
gianfra-t: This was causing the onramp dialog to immediately show an "expired quote..." message right after click on confirm, which would then be cleared. I commented it because I wasn't sure if this could bring any unwanted side effects.

apps/frontend/src/hooks/offramp/useRampService/useRegisterRamp.ts
gianfra-t: Also answering [here](https://github.com/pendulum-chain/vortex/pull/731#discussion_r2180450386). I found that this was causing issues when using Monerium, probably because we must keep the functionality after the redirect/refresh. I agree the name changes were a bit unnecessary, but I tried to simplify the logic.

For this particular case, there was also another lock check on this [line](https://github.com/pendulum-chain/vortex/blob/1e920becc997a805ef19ee643a82f35c5ba0e29f/apps/frontend/src/hooks/offramp/useRampService/useRegisterRamp.ts#L122C5-L122C36) which should prevent the execution of the [callback](https://github.com/pendulum-chain/vortex/blob/1e920becc997a805ef19ee643a82f35c5ba0e29f/apps/frontend/src/hooks/offramp/useRampService/useRegisterRamp.ts#L243) anyway. Is this intentional ? Maybe I'm not seeing the use of the second (inner) check.

apps/api/src/api/services/evm/clientManager.ts
```diff
@@ -0,0 +1,119 @@
+import { Account, Chain, createPublicClient, createWalletClient, http, PublicClient, WalletClient } from "viem";
+import { moonbeam, polygon } from "viem/chains";
+import logger from "../../../config/logger";
+import { ALCHEMY_API_KEY } from "../../../constants/constants";
+
+// So far we only support (and need) Polygon and Moonbeam as EVM networks.
+export type EvmNetwork = "polygon" | "moonbeam";
```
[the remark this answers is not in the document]
gianfra-t: What about `RpcSupportedNetworks` ? True, the name is unfortunate. I forgot about the existing one, but we can't use the other one as I would prefer this class to handle only the chains we know we interact with during the ramps.

apps/api/src/api/services/monerium/index.ts
```diff
@@ -0,0 +1,172 @@
...
+  FetchIbansParams,
+  FetchProfileParams,
+  IbanData,
+  IbanDataResponse,
+  MoneriumResponse,
+  MoneriumTokenResponse,
+  MoneriumUserProfile
+} from "./types";
+
+const MONERIOUM_API_URL = `https://api.monerium.app`;
```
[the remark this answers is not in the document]
gianfra-t: Ups...

apps/api/src/api/services/phases/handlers/monerium-onramp-mint-handler.ts
```diff
@@ -0,0 +1,72 @@
...
+        ERC20_EURE_POLYGON,
+        walletAddress,
+        inputAmountBeforeSwapRaw,
+        pollingTimeMs,
+        EVM_BALANCE_CHECK_TIMEOUT_MS,
+        "polygon"
+      );
+
+      // Add delay to ensure the transaction is settled
+      await new Promise(resolve => setTimeout(resolve, 30000)); // 30 seconds.
```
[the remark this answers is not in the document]
gianfra-t: Yes, I encountered some edge cases where the rpc would not react to the balance change, and the next phase would fail. 
We had to do something similar on `BrlaTeleport` and some other phases.

apps/api/src/api/services/phases/handlers/fund-ephemeral-handler.ts
```diff
@@ -134,6 +168,43 @@ export class FundEphemeralPhaseHandler extends BasePhaseHandler {
...
+    try {
+      const ephmeralAddress = state.state.polygonEphemeralAddress;
+      const fundingAmountRaw = multiplyByPowerOfTen(
+        POLYGON_EPHEMERAL_STARTING_BALANCE_UNITS,
+        polygon.nativeCurrency.decimals
+      ).toFixed();
+
+      // We use Moonbeam's funding account to fund the ephemeral account on Polygon.
+      const fundingAccount = privateKeyToAccount(MOONBEAM_FUNDING_PRIVATE_KEY as `0x${string}`);
+      const walletClient = createWalletClient({
```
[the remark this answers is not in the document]
gianfra-t: Yes, was pushed a bit after.

apps/api/src/api/services/phases/handlers/squid-router-pay-phase-handler.ts
```diff
@@ -95,12 +115,28 @@ export class SquidRouterPayPhaseHandler extends BasePhaseHandler {
...
+        const squidrouterStatus = await this.getSquidrouterStatus(swapHash, state);
+
+        if (squidrouterStatus.status === "success") {
+          isExecuted = true;
+          logger.info(`SquidRouterPayPhaseHandler: Transaction ${swapHash} successfully executed on Squidrouter.`);
+          break;
+        }
+        if (!squidrouterStatus) {
+          logger.warn(`SquidRouterPayPhaseHandler: No squidrouter status found for swap hash ${swapHash}.`);
+          throw this.createRecoverableError("No squidrouter status found for swap hash.");
```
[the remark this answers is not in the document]
gianfra-t: So far I found that after the fix on the query params it always returns the status, at least at some point. In any case, e need to rely on their API, since an onramp to `Polygon` will not use Axelar at all.

apps/api/src/api/services/phases/handlers/squid-router-pay-phase-handler.ts
```diff
@@ -140,34 +176,114 @@ export class SquidRouterPayPhaseHandler extends BasePhaseHandler {
...
+
+  /**
+   * Execute a call to the Axelar gas service on Polygon network.
+   * @param tokenValueRaw The amount of MATIC to fund the transaction with.
+   * @param swapHash The swap transaction hash.
+   * @param logIndex The log index from Axelar scan.
    * @returns Hash of the transaction that funds the Axelar gas service.
    */
-  private async executeFundTransaction(gmlrValueRaw: string, swapHash: `0x${string}`, logIndex: number): Promise<string> {
+  private async executeFundTransactionOnPolygon(
```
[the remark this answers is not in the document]
gianfra-t: I had issues with the typings of the wallet client. But this was before the new `EvmClientManager`. Maybe now we can combine them.

apps/frontend/src/components/RampSummaryDialog/index.tsx
gianfra-t: I opted for using the execution input because it should be defined at this stage on these components. And since `useRampDirection` reacts to the UI values, I thought it would be better to use the execution state directly in case the UI would change after confirm, which happens when refreshing for Monerium, for example.

apps/api/src/api/services/transactions/moneriumEvmOnrampTransactions.ts
```diff
@@ -0,0 +1,266 @@
...
+import erc20ABI from "../../../contracts/ERC20";
+import { QuoteTicketAttributes } from "../../../models/quoteTicket.model";
+import { EvmClientManager } from "../evm/clientManager";
+import { getMoneriumEvmDefaultMintAddress } from "../monerium";
+import { multiplyByPowerOfTen } from "../pendulum/helpers";
+import { StateMetadata } from "../phases/meta-state-types";
+import { encodeEvmTransactionData } from "./index";
+import { createOnrampSquidrouterTransactionsToEvm } from "./squidrouter/onramp";
+
+export const ERC20_EURE_POLYGON: `0x${string}` = "0x18ec0a6e18e5bc3784fdd3a3634b31245ab704f6"; // EUR.e on Polygon
```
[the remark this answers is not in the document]
gianfra-t: True.

apps/api/src/api/services/transactions/moneriumEvmOnrampTransactions.ts
```diff
@@ -0,0 +1,266 @@
...
+  // Find required ephemeral accounts
+  // We use Moonbeam as the generic EVM chain.
+  const polygonEphemeralEntry = signingAccounts.find(ephemeral => ephemeral.network === Networks.Moonbeam);
+  if (!polygonEphemeralEntry) {
+    throw new Error("Polygon ephemeral not found");
+  }
+
+  // Calculate amounts
+  const inputAmountPostAnchorFeeUnits = new Big(quote.inputAmount).minus(quote.fee.anchor);
+  const inputAmountPostAnchorFeeRaw = multiplyByPowerOfTen(inputAmountPostAnchorFeeUnits, 18).toFixed(0, 0);
```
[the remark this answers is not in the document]
gianfra-t: Added it to the `index.ts` for Monerium's service. Good call.

apps/api/src/api/services/ramp/ramp.service.ts
```diff
@@ -139,22 +136,98 @@ export class RampService extends BaseRampService {
...
+    depositQrCode?: string;
+    ibanPaymentData?: IbanPaymentData;
+  }> {
     if (quote.rampType === "off") {
       if (quote.outputCurrency === FiatToken.BRL) {
         return this.prepareOfframpBrlTransactions(quote, normalizedSigningAccounts, additionalData);
-      } else {
+        // If the property moneriumAuthToken is not provided, we assume this is a regular Stellar offramp.
+        // otherwise, it is automatically assumed to be a Monerium offramp.
+      } else if (!additionalData?.moneriumAuthToken) {
```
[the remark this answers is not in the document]
gianfra-t: Recall this is for offramp, not onramp. I would say that we should have different token definitions entirely, assuming we keep Mykobo. Right now, we could not differentiate them except for the additional data used.

apps/api/src/api/services/phases/handlers/squid-router-pay-phase-handler.ts
```diff
@@ -200,30 +316,51 @@ export class SquidRouterPayPhaseHandler extends BasePhaseHandler {
...
     try {
-      const routeParams = createOnrampRouteParams(
-        stateMeta.moonbeamEphemeralAddress,
-        stateMeta.outputAmountBeforeFinalStep.raw,
-        outputTokenDetails,
-        toNetwork,
-        ramp.state.destinationAddress
-      );
-      const routeResult = await getRoute(routeParams);
+      const fromChainId = getNetworkId(Networks.Polygon)?.toString(); // Always Polygon for Monerium onramp.
```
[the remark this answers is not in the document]
gianfra-t: Ah! Nice find indeed. Thanks.

apps/frontend/src/components/RampSummaryDialog/index.tsx
ebma: I see. We could still change the `useRampDirection` hook to do the same. Then we might also get away with the 'buy vs sell is switching on page reload'-issue. If, in `useRampDirection` we check if a quote is available and return the direction based on that, it could do the trick, right?

apps/frontend/src/components/RampSummaryDialog/TransactionTokensDisplay.tsx
ebma: Ah okay 👍

apps/frontend/src/hooks/offramp/useRampService/useRegisterRamp.ts
ebma: Ah I didn't pay attention to that. Nevermind then, let's keep it.

apps/api/src/api/services/phases/handlers/squid-router-pay-phase-handler.ts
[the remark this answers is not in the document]
ebma: True 👍  I was just curious if we are confident enough in their endpoint. Let's try it like this then.

apps/api/src/api/services/phases/handlers/monerium-onramp-mint-handler.ts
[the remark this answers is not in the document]
ebma: Alright, fine with me 👍

apps/api/src/api/services/evm/clientManager.ts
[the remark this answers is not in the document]
ebma: Hmm. Honestly, I would not impose this limitation. I would stick to the `EvmNetworks` type we already have and change the struct for the `EvmNetworkConfig[]` to have info for all of them. While we might use only two of them, it doesn't hurt to have the functionality for all of them at the same time. It comes for free so to say. The only 'problem' I see is your call to `getAllClients()` to initialize them on startup. I don't think we necessarily need that. I suppose you just added it for convenience so that ramps are a little faster?

apps/api/src/api/services/monerium/index.ts
```diff
@@ -0,0 +1,172 @@
...
+  } catch (error) {
+    console.error("Failed to fetch addresses:", error);
+    return null;
+  }
+};
+
+export const getMoneriumEvmDefaultMintAddress = async (token: string): Promise<string | null> => {
+  // Assumption is the first linked address is the default mint address for Monerium EVM transactions.
+  // TODO: this needs to be confirmed.
+  return getFirstMoneriumLinkedAddress(token);
```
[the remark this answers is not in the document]
ebma: True. And how do we know? Can we decide that somehow? Sorry for the noob question, but asking that is faster than me trying to look it up 😅

apps/api/src/api/services/ramp/ramp.service.ts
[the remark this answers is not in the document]
ebma: Alright, fine with me. Let's maybe add a `// FIXME change to a better check once Mykobo support is dropped` or something? So we have a little reminder that this can be improved.

apps/frontend/src/hooks/offramp/useRampService/useRegisterRamp.ts
gianfra-t: Wait but , comparing from staging again, I only see a few `if` conditions between the two. Hook triggers, then we get the first check, and immediately after the function `registerRampProcess` gets called (the rest is the definition of the function itself).
And once inside it, the second check is one of the first operations.

apps/frontend/src/hooks/ramp/useRampValidation.ts
```diff
@@ -35,6 +35,8 @@ function validateOnramp(
     trackEvent: (event: TrackableEvent) => void;
   }
 ): string | null {
+  // Monerium EUR has no minimum
```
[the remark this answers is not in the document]
gianfra-t: Yeah we could. Any particular reason?

apps/frontend/src/hooks/offramp/useSubmitRamp.ts
```diff
@@ -202,7 +288,23 @@ export const useSubmitRamp = () => {
...
+      getMessageSignature,
+      setRampSigningPhase,
+      setMoneriumTriggered,
+      resetRampState
     ]
   );
 };
+
+// Helper function to determine if EUR flow should use Monerium
+async function shouldRouteToMonerium(executionInput: RampExecutionInput): Promise<boolean> {
```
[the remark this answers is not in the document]
gianfra-t: Yeah the idea was to leave it up to the backend. At least originally.

apps/api/src/api/services/evm/clientManager.ts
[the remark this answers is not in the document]
gianfra-t: Yeah we can do that of course. No strong opinion on it. The `getAllClients` is useful for initializing them so they are quicker on the first call.

apps/api/src/api/services/monerium/index.ts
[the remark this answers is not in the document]
gianfra-t: Honestly, I guess we need to ask someone at Monerium. I couldn't figure it out from the API alone.

apps/api/src/api/services/monerium/index.ts
```diff
@@ -0,0 +1,172 @@
...
+    const response = await fetch(url, { headers });
+
+    if (!response.ok) {
+      throw new Error(`HTTP error! status: ${response.status}`);
+    }
+
+    const data: MoneriumResponse = await response.json();
+
+    if (data.addresses && data.addresses.length > 0) {
+      const firstAddress = data.addresses[data.addresses.length - 1].address; // Ordered by creation date, so last is the most recent.
```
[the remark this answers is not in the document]
gianfra-t: I found that they are ordered from newest to older. Not very intuitive !

apps/frontend/src/components/RampSummaryDialog/index.tsx
gianfra-t: Yeah sure, I may need help with the default values on refresh though. I couldn't prevent the switch to 'buy' after refresh, but the issue got "solved" because the onramp dialog blocks it a bit.

apps/frontend/src/hooks/ramp/useRampValidation.ts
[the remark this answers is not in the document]
ebma: I'd be afraid of users depleting our nabla pre/post swap subsidy with useless cent ramps.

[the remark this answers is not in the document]
gianfra-t: True but there are no subsidies on this new route. The point still stands though, as we pay for the bridge and swap ourselves. 
I thought such a small amount would have been a nice to have, but let's put back the limit.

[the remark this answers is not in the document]
ebma: Haha you're right. I think 1 EUR is a nice minimum to have. It's lower than our previous one of 10 EUR.

apps/api/src/api/services/monerium/index.ts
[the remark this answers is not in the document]
ebma: But doesn't the point still stand that this function should be called `getLatest..`? It's not the first after all?

apps/api/src/api/services/transactions/moneriumEvmOnrampTransactions.ts
```diff
@@ -0,0 +1,266 @@
...
+    abi: erc20ABI,
+    args: [fromAddress, toAddress, amountRaw],
+    functionName: "transferFrom"
+  });
+
+  const { maxFeePerGas } = await publicClient.estimateFeesPerGas();
+
+  const txData: EvmTransactionData = {
+    data: transferCallData as `0x${string}`,
+    gas: "100000",
```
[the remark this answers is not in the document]
gianfra-t: I just picked a previous value we used that would be "enough" for that transaction. Let me see if I can either remove it or estimate it.

apps/api/src/api/services/phases/meta-state-types.ts
```diff
@@ -46,5 +46,8 @@ export interface StateMetadata {
   squidRouterPayTxHash: string;
   unhandledPaymentAlertSent: boolean;
   pendulumToMoonbeamXcmHash?: string;
-  brCode: string | undefined;
+  depositQrCode: string | undefined;
+  // Only used in onramp, offramp - monerium
+  polygonEphemeralAddress: string;
+  ibanPaymentData: IbanPaymentData;
```
Sharqiewicz, line 52: @pendulum-chain/devs  This object begins to become really complex and intuitive. I think we should group it into smaller object-properties, wdyt? (in a separate PR)

```
import { PendulumTokenDetails, RampCurrency, StellarTokenDetails } from "@packages/shared";
import { ExtrinsicOptions } from "../transactions/nabla";

export interface CoreTransactionData {
  inputAmount: string;
  outputAmount: string;
  inputCurrency: RampCurrency;
  outputCurrency: RampCurrency;
  inputTokenPendulumDetails: PendulumTokenDetails;
  outputTokenPendulumDetails: PendulumTokenDetails;
  outputTokenType: RampCurrency;
  inputAmountBeforeSwapRaw: string;
  // The final step for onramp is the squidRouterSwap or XCM transfer, for offramps it's the anchor payout
  outputAmountBeforeFinalStep: { units: string; raw: string };
}

export interface AddressData {
  pendulumEphemeralAddress: string;
  moonbeamEphemeralAddress: string;
  walletAddress: string | undefined;
  destinationAddress: string;
  brlaEvmAddress: string;
}

export interface TransactionHashes {
  moonbeamXcmTransactionHash: string;
  squidRouterReceiverHash: string;
  pendulumToAssethubXcmHash?: string;
  assetHubToPendulumHash: string;
  squidRouterApproveHash: string;
  squidRouterSwapHash: string;
  squidRouterPayTxHash: string;
  pendulumToMoonbeamXcmHash?: string;
}

export interface StellarData {
  // Only used in offramp - eurc & ars route
  stellarEphemeralAccountId: string;
  stellarTarget: {
    stellarTargetAccountId: string;
    stellarTokenDetails: StellarTokenDetails;
  };
}

export interface BrlaData {
  // Only used in onramp - brla
  inputAmountUnits: string;
  inputAmountBeforeSwapUnits: string;
  taxId: string;
  pixDestination: string;
  receiverTaxId: string;
}

export interface MoneriumData {
  // Only used in onramp, offramp - monerium
  polygonEphemeralAddress: string;
  ibanPaymentData: string;
}

export interface NablaData {
  nablaSoftMinimumOutputRaw: string;
  nabla: {
    approveExtrinsicOptions: ExtrinsicOptions;
    swapExtrinsicOptions: ExtrinsicOptions;
  };
}

export interface SystemState {
  // Only used in offramp
  squidRouterReceiverId: string;
  executeSpacewalkNonce: number;
  unhandledPaymentAlertSent: boolean;
  depositQrCode: string | undefined;
}

export interface StateMetadata {
  core: CoreTransactionData;
  addresses: AddressData;
  hashes: TransactionHashes;
  stellar?: StellarData;
  brla?: BrlaData;
  monerium?: MoneriumData;
  nabla: NablaData;
  system: SystemState;
}

```

gianfra-t: Yes, I agree with that. Even better, we should try to define different meta-states for our different routes. Otherwise we end up with a very "sparse" state.

Sharqiewicz: Definitely agree!

package.json
```diff
@@ -1,5 +1,6 @@
 {
   "dependencies": {
+    "big.js": "^6.2.1",
```
ebma, line 3: I don't think we need this in the root workspace.

calculate-gas-fee.ts
```diff
@@ -0,0 +1,100 @@
...
+  is_insufficient_fee: boolean;
+  status: string;
+  fees: AxelarScanStatusFees;
+  id: string;
+}
+
+const DEFAULT_SQUIDROUTER_GAS_ESTIMATE = "800000";
+
+function calculateGasFeeInUnits(feeResponse: AxelarScanStatusFees, estimatedGas: string | number): string {
+  console.log("fee response object", feeResponse);
```
ebma, line 34: Nit: let's remove?

apps/api/src/api/services/phases/handlers/squid-router-pay-phase-handler.ts
```diff
@@ -331,6 +330,7 @@ export class SquidRouterPayPhaseHandler extends BasePhaseHandler {
   }
 
   private calculateGasFeeInUnits(feeResponse: AxelarScanStatusFees, estimatedGas: string | number): string {
+    console.log("fee response object", feeResponse);
```
ebma, line 333: Nit: also remove

ebma (approved):

PR #734, opened 2025-06-23
[READY] Implement automatic quote refresh

Sharqiewicz:
Introduced an animated progress indicator that provides visual feedback when quotes are being refreshed.
Refresh the quote every 30s until the summary is shown.


ebma (approved):
Tested it locally, seems to work as expected 🙏  nice job @Sharqiewicz, let's merge

PR #741, opened 2025-06-27
[DRAFT] Improve resilience of presigned stellar transactions

Sharqiewicz:


apps/api/src/api/services/phases/post-process/stellar-post-process-handler.ts
```diff
@@ -42,7 +42,7 @@ export class StellarPostProcessHandler extends BasePostProcessHandler {
    */
   public async process(state: RampState): Promise<[boolean, Error | null]> {
     try {
-      const expectedLedgerTimeMs = state.createdAt.getTime() + SEQUENCE_TIME_WINDOW_IN_SECONDS * 1.1 * 1000; // Add some safety margin in case ledger producton was slower.
+      const expectedLedgerTimeMs = state.createdAt.getTime() + SEQUENCE_TIME_WINDOWS.FIFTH_TX * 1.1 * 1000; // Add some safety margin in case ledger production was slower. - Use the longest timeframe (90 days)
```
gianfra-t, line 45: I think most transactions will use the default shift of 5 minutes (more or less). Do you think we could use that as default and eventually handle or live with the cleanup error?

apps/api/src/api/services/transactions/stellar/offrampTransaction.ts
```diff
@@ -88,7 +110,7 @@ export async function buildPaymentAndMergeTx({
           source: ephemeralAccountId
         })
       )
-      .setTimebounds(0, maxTime)
+      .setTimebounds(0, maxTimeForTx)
```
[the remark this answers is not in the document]
gianfra-t: We should be carefully with the `setTimebounds` value. The shift we define is an estimate used for ledger sequence but it could be that if the ledger production is different from the estimate, let’s say slower, our pre-signed transaction can still be valid in terms of sequence number, but not this time bound.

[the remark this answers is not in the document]
Sharqiewicz: Yes, but we use the incremental timebounds right? Do you want to always set the 90-days timebound? Before it was about 10 minutes.

apps/api/src/api/services/phases/post-process/stellar-post-process-handler.ts
Sharqiewicz: Yes, sure. We can do that 👍

apps/api/src/api/services/transactions/stellar/offrampTransaction.ts
```diff
@@ -41,20 +45,37 @@ export async function buildPaymentAndMergeTx({
...
+    SEQUENCE_TIME_WINDOWS.FIFTH_TX
+  ];
+
+  const sequenceNumbers: string[] = [];
+  for (let i = 0; i < NUMBER_OF_PRESIGNED_TXS; i++) {
+    const sequenceNumber = await getFutureShiftedLedgerSequence(horizonServer, 32, timeWindows[i]);
+    sequenceNumbers.push(sequenceNumber);
+  }
+
+  const expectedSequenceNumber = sequenceNumbers[0]; // Use first sequence number as expected
```
ebma, line 70: This will be put onto the `meta` of each Stellar transaction when we store them in the database. I think it would be better to have each transaction stored with the respective sequence number. This makes it easier for us later on to see what transactions can still be used and which can't in case that we manually need to do something here.

apps/api/src/api/services/phases/helpers/stellar-payment-verifier.ts
```diff
@@ -0,0 +1,92 @@
...
+
+  if (!stateMetadata.stellarTarget) {
+    throw new Error("Stellar target information not found in state metadata");
+  }
+
+  if (!stateMetadata.outputAmountBeforeFinalStep) {
+    throw new Error("Output amount information not found in state metadata");
+  }
+
+  const { stellarEphemeralAccountId, stellarTarget, outputAmountBeforeFinalStep } = stateMetadata;
```
ebma, line 27: Nit: We can move this declaration above the `if() throw Error` statements.

ebma (commented):
Nice job, looks good to me overall :ok:. Have you already tried testing this locally?

apps/api/src/api/services/transactions/stellar/offrampTransaction.ts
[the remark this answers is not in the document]
gianfra-t: Yeah I guess `10` minutes was already quite "close" in the past, but now that the transaction could eventually take days to be sent I think we should relax this value even more. We could even set '0' to make it valid [indefinitely](https://stellar.github.io/js-stellar-sdk/TransactionBuilder.html#setTimebounds).

ebma (approved):
Looks good to me 👍

Issue #716, opened 2025-06-12
Capture the subsidy amounts in the backend


prayagd:
## Context
Currently we are subsidizing couple of costs both for Buy and Sell operations, in order to understand them we want to capture them per transaction in the backend

## Todo
* Capture XCM cost both Moonbeam and Pendulum[In BRL case], Forex AMM swap cost, Stellar transfer cost for the Sell operations
* Capture XCM cost both Moonbeam and Pendulum, Forex AMM swap cost for the Buy operations

prayagd:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Pendulum-6408a0365cda59c57d39bee5/issues/gh/pendulum-chain/vortex/716?planning-poker) with Zenhub @ebma @gianfra-t @Sharqiewicz

ebma:
Right, I think this ticket here is kinda a duplicate of #617 but a little more detailed. I'll close the other one.

gianfra-t:
@pendulum-chain/devs I check the most relevant costs for each phase and so far chose to register the cost of the following phases:
- Squidrouter pay: This is the main expense for most ramps. It should eventually be compensated by the Network fee we extract.
- Pre/Post swap subsidy: Also a potentially relevant expense, that can go unnoticed. 
- Pendulum to Moonbeam XCM: We pay a fixed `GLMR` amount per transaction.
- For the cases of Moonbeam to Pendulum XCM, this is payed with the currency being transferred on Pendulum, and is then reflected on the amount to subsidize pre-swap, so I think it can be redundant to register both.
- For Assethub transfers, I believe these are payed by the user.

Of course there are many other "expenses" the flow has we could register, but most of them are payed by our ephemerals which we then merge. We can keep an eye on these costs by just tracking the balances of the ephemeral fundings. 
If we wanted to do this more precise, we could register the balance during the cleanup phase and registering the difference between the initially funded amount vs what was merged.

PR #745, opened 2025-07-04
Adding subsidies table.

gianfra-t:
Issue: #716 .

### Additions
- Adds a new model to register subsidy entries, connected to the `ramp_id`.
- Register the subsidy of the pre/post swap phases,
- Register the cost of the `squidrouter-pay` phase.
- Register the cost of `pendulumToMoonbeam` XCM transaction.


apps/api/src/database/migrations/009-token-migrator.ts
```diff
@@ -0,0 +1,20 @@
+import { QueryInterface } from "sequelize";
+
+// This migration adds new tokens to the ENUM type used for subsidies. Needed after SubsidyToken enum was extended.
+const ENUM_NAME = "enum_subsidies_token";
+
+const NEW_TOKENS: any = [];
+
+export async function up(queryInterface: QueryInterface): Promise<void> {
```
ebma, line 8: From what I understand, you only added this file in case we need to extend the enum in the future? If so, let's remove it. If we keep it like this, it will automatically run, but do nothing. And then in the future, we need to add yet another migration that is a copy of this. It's just a bit redundant, so let's only include it once necessary.

ebma (commented):
Looks good to me overall 👍 I didn't test it locally but there is not much to go wrong. I'd only like us to remove that one migration file and then we can merge. 

The ticket #716 asks to also cover the XCM costs. This is not done here. I'm not sure how easy it would be to get access to that data, I suppose we'd have to wait for the XCM transfer to finalize and then check in the block events how much was charged for it. Or do you think there is an easier way to do that @gianfra-t? 
Also, we are now tracking the funding of the GLMR amount, but not the XLM for Stellar. We are getting parts of both values back as part of the cleanup. Ideally, we'd also account for those values in the subsidy as it would give valueable insights. However, it's difficult to implement due to the timing difference here. The payback can happen with quite a delay.

gianfra-t:
Thanks @ebma. Regarding XCM and the choices of what subsidy to register, I left a comment [here](https://github.com/pendulum-chain/vortex/issues/716#issuecomment-3053688824). Let me know if it makes sense.

Regarding the funding amounts, we are not really tracking the funding of the Moonbeam account , but rather the fixed XCM cost only which happens in Pendulum. And as you say, a next iteration of this would probably involve subtracting the difference between what's funded vs what's recovered, which "should" be equal to the funds spend on pure transactions cost.

apps/api/src/database/migrations/009-token-migrator.ts
gianfra-t: Mmm I thought we could execute this migration later, but you're right it doesn't make sense.

PR #765, opened 2025-07-11
[READY]  implement translating quote error messages

Sharqiewicz:
This pull request introduces a new `QuoteError` enum to standardize error handling in the quote service and frontend, along with user-friendly error messages for better user experience. The changes span across the backend, frontend, and shared packages, ensuring consistent error management and improved error communication.

### Backend Changes:
* Introduced `QuoteError` enum in `packages/shared/src/endpoints/quote.endpoints.ts` to define standardized error codes for quote-related operations.
* Updated error handling in `quote-fees.ts` to use `QuoteError` instead of hardcoded error messages for `calculatePreNablaDeductibleFees` and `calculateFeeComponents` functions. [[1]](diffhunk://#diff-977fba745a9b3539edb7d11b3c32cabc9d640b159495cee5b88bf31c8e2a57c2R314-R316) [[2]](diffhunk://#diff-977fba745a9b3539edb7d11b3c32cabc9d640b159495cee5b88bf31c8e2a57c2L370-R371)

### Frontend Changes:
* Added `getFriendlyErrorMessage` function in `useQuoteStore.ts` to map `QuoteError` codes to user-friendly error messages.
* Updated `useQuoteStore` to use `getFriendlyErrorMessage` for displaying errors when fetching quotes fails.

### Localization Updates:
* Added new user-friendly error messages for `QuoteError` codes in English (`en.json`) and Portuguese (`pt.json`) translation files. [[1]](diffhunk://#diff-5299939190b7c8ef5d8f4cb57b5a73c4c520de4181a827d82d88a6ac9ca22181R392) [[2]](diffhunk://#diff-5299939190b7c8ef5d8f4cb57b5a73c4c520de4181a827d82d88a6ac9ca22181L410-R412) [[3]](diffhunk://#diff-1132459c669bcf90aa4f884760a2566b0d6073a834abfe55c06893c686da727dR391) [[4]](diffhunk://#diff-1132459c669bcf90aa4f884760a2566b0d6073a834abfe55c06893c686da727dL409-R411)


gianfra-t (commented):
Looks good to me! I think this is a nice improvement for our UX. 
The only thing I would change is the use of `tryDifferentAmount` friendly error. I would be a bit confused as to go with a larger or lower amount.

I know that the fine-grained errors that map to it are hard to "dumb-down", we don't even know most of the times, but maybe we could use by default a "try lower amount", wdyt?

ebma (approved):
I love it the way it is. 

> I know that the fine-grained errors that map to it are hard to "dumb-down", we don't even know most of the times, but maybe we could use by default a "try lower amount", wdyt?

I kinda see your point but what if we'd then actually point the user into the wrong direction? As a user I would prefer to get the hint to try a different amount than be wrongly advised to try a lower one. We had this happen with squidrouter once that they were suggesting to try a lower amount but all I needed was a larger amount. Due to the hint I didn't even realize that.

PR #766, opened 2025-07-11
Vortex client.

gianfra-t:
Implements a client that abstracts Vortex's API and ephemeral handling into a self contained package.


packages/vortex-signer/example.ts
```diff
@@ -0,0 +1,85 @@
...
+    console.log(`   Output: ${quote.outputAmount} ${quote.outputCurrency}`);
+    console.log(`   Fee: ${quote.fee}`);
+    console.log(`   Expires at: ${quote.expiresAt}\n`);
+
+    const brlaOnrampData: BrlaOnrampAdditionalData = {
+      destinationAddress: "0x1234567890123456789012345678901234567890",
+      taxId: ""
+    };
+
+    const registeredRamp = await signer.registerBrlaOnramp(quote.id, brlaOnrampData);
```
ebma, line 47: Let's try to make all the names a little more generic for the average consumer of this. Let's a) not mention Brla here explicitly and call the function `registerRamp`. b) It's clear from the quote that it's a brla onramp, so we can just inside of that function query that quote, and check the direction. The second parameter can be generalized to 'additionalData'.

packages/vortex-signer/package.json
```diff
@@ -0,0 +1,39 @@
...
+    ".": {
+      "import": "./dist/index.js",
+      "require": "./dist/index.js",
+      "types": "./dist/index.d.ts"
+    },
+    "./package.json": "./package.json"
+  },
+  "main": "./dist/index.js",
+  "module": "./dist/index.js",
+  "name": "vortex-sign",
```
ebma, line 20: Let's rename this to @ vortexfi/sdk.

packages/vortex-signer/src/VortexSigner.ts
```diff
@@ -0,0 +1,127 @@
...
+} from "@packages/shared";
+import { Networks, signUnsignedTransactions } from "@packages/shared";
+import { createMoonbeamEphemeral, createPendulumEphemeral, createStellarEphemeral } from "./ephemeralHelpers";
+import { EphemeralGenerationError, TransactionSigningError } from "./errors";
+import { BrlaHandler } from "./handlers/BrlaHandler";
+import { ApiService } from "./services/ApiService";
+import { NetworkManager } from "./services/NetworkManager";
+import type { BrlaOnrampAdditionalData, VortexSignerConfig, VortexSignerContext } from "./types";
+
+export class VortexSigner {
```
ebma, line 17: We can also rename this to `VortexSdk` so we can logically extend it in the future.

ebma (commented):
Looks good to me overall, just some nitpicks about the naming.

packages/vortex-signer/example.ts
gianfra-t: But ultimately that was the idea of separating each ramp logic into it's own class, such that we can keep the methods generic and each ramp type can freely use them as it is required by the type.

packages/vortex-signer/src/VortexSigner.ts
gianfra-t: Renamed!

packages/sdk/src/VortexSdk.ts
```diff
@@ -122,31 +125,58 @@ export class VortexSdk {
...
+
+        if (ephemeral) {
+          accountMetas.push({
+            address: ephemeral.address,
+            network
+          });
+
+          if (this.storeEphemeralKey) {
+            const { address, secret } = ephemeral;
+            const fileName = `${network}_ephemeral_key.json`;
```
ebma, line 154: It will be very hard to find the respective keys for a stuck ramp later, if needed. This file will just grow very large and there is no reference to the ramp. We need to include the ramp ID of the ramp that is related to these ephemeral keys. Thus, we need to change the signature of this function to also take a ramp ID as a parameter and then I would propose we change the file name to `${rampId}_keys.json`.

packages/sdk/example.ts
```diff
@@ -41,7 +43,7 @@ async function runBrlOnrampExample() {
 
     const brlOnrampData = {
       destinationAddress: "0x1234567890123456789012345678901234567890",
-      taxId: "123.456.789-00"
+      taxId: "718.784.051-95"
```
ebma, line 46: Let's not dox a valid tax ID. Let's stick to the 123...

packages/sdk/example.ts
```diff
@@ -12,7 +12,9 @@ async function runBrlOnrampExample() {
       // Optional: provide custom WebSocket URLs
       // pendulumWsUrl: 'wss://custom-pendulum-rpc.com',
       // moonbeamWsUrl: 'wss://custom-moonbeam-rpc.com',
-      // autoReconnect: true, // default is true
+      // autoReconnect: true, // default is true,
+      // Optional: store ephemeral keys for later use
+      //storeEphemeralKey: true,
     };
```
ebma, line 18: I'm still no fan of all these comments. 
How about we just do
```
const config: VortexSdkConfig = {
      apiBaseUrl: "http://localhost:3000"	 
      // Optional: provide custom WebSocket URLs	,
      pendulumWsUrl: undefined, // 'wss://custom-pendulum-rpc.com',	
      moonbeamWsUrl: undefined, // 'wss://custom-moonbeam-rpc.com',
      autoReconnect: true, // default is `true`
      // Optional: store ephemeral keys for later use
      storeEphemeralKey: true,
    };
	    };
```

packages/sdk/src/VortexSdk.ts
```diff
@@ -25,10 +26,12 @@ export class VortexSdk {
   private networkManager: NetworkManager;
   private brlHandler: BrlHandler;
   private initializationPromise: Promise<void>;
+  private storeEphemeralKey: boolean;
```
ebma, line 29: Nit:
```suggestion
  private storeEphemeralKeys: boolean;
```

packages/sdk/src/VortexSdk.ts
gianfra-t: I pushed the proposed idea [here](https://github.com/pendulum-chain/vortex/pull/766/commits/035cf9e5661bb765a51809befeb6502d2c00888d), but with the large file. We can split it if you still prefer the individual-ramp files.

ebma: Fine with me since we now include the rampId in each entry.

packages/sdk/package.json
```diff
@@ -0,0 +1,53 @@
...
+    "dev": "bun run --watch ./src/index.ts",
+    "format": "prettier --write .",
+    "lint": "eslint . --ext .ts",
+    "prepublishOnly": "bun run build",
+    "test": "bun test",
+    "typecheck": "bun x --bun tsc"
+  },
+  "type": "module",
+  "types": "./dist/index.d.ts",
+  "version": "0.1.7"
```
ebma, line 52: Shall we set this to 
```suggestion
  "version": "1.0.0"
```

ebma (commented):
I think our package is way too large 😅  if someone wanted to use it in a browser environment, it would dramatically increase the time of the page load. 
The size mostly comes from the `browser` and `node` dependencies that you included in the package.json. Any chance we can get rid of them while still having it work as expected?

image

ebma (approved):

Sharqiewicz (approved):
Awesome, so removing two different builds succeeded 💪 Let's go

Issue #775, opened 2025-07-24
BRL offramp via API


prayagd:
## Context
As a brazilian user of vortex offramp, want to use vortex via API

## Todo
* Provide a library for signing and ephemeral account creation (Vortex JS)
* No KYC for new users - Consider that only the existing KYC'ed users are going to use it for the first iteration
* Partners can use their custom fee config, when transacting via API.

PR #782, opened 2025-07-28
Add BRL offramp feature to vortex-sdk.

gianfra-t:
Issue: #775.

### About
- Adds logic to handle the BRL off-ramp.
- Return (when necessary) transactions user must sign before starting the ramp.


gianfra-t:
Thanks @ebma, and yes I agree that example could some more explaining. Let me add some extra explanation but let's do the bulk of the documentation in the API docs and link to there.

ebma (approved):

PR #790, opened 2025-08-05
[WIP] Revamp token selection 

Sharqiewicz:


ebma (commented):
I love it 🥹

1) I noticed that the assets are not sorted by the value I own. I think it would be very nice for users if they see the tokens with the highest balance up top. Or at least some ordering of 
1st group: Assets with some balance, ordered by name
2nd group: Assets with no balance, ordered by name
We don't have to make this grouping visible with a divider or something, but as I said, having assets I own shown at the top would be great. 
image

2) I'd like to have the transitions a little quicker, also the transition for the transaction history. It takes a little too long for my taste until the dialogs are gone and I can interact with the app again. 

3) Sometimes when you select an asset from the list in the dropdown, the remaining items are re-ordered arbitrarily before the dialog is closing. This doesn't happen all the time and it's not a big deal but it's still weird to see as a user. 

4) The card is truncated too much so that not all the content is visible anymore. On 'Buy', it's missing the icons below the button row, on 'Sell' it's even missing the buttons to confirm. 
image

gianfra-t (commented):
Looking good! Really like the new components. For the preview, I tested a few combinations but I get `Invalid token details for EVM bridge`. Maybe because the object changed?

Sharqiewicz:
@gianfra-t The problem was that we were using several variants to name our tokens: ‘ETH’, “eth”, ‘USDC.e’, ‘usdc.e’, ‘usdce’. I fixed it now

PR #793, opened 2025-08-11
create helper functions to detect current env

Sharqiewicz:
Detect runtime environment (browser vs. server) and avoid importing winston in the browser.

Add a getEnvVar helper to validate and retrieve environment variables safely


ebma (approved):
Looks good to me 👌

PR #821, opened 2025-09-08
Add support for onramping to assethub with USDC/USDT/DOT

### LINKED ISSUE #820: Add support for onramping to Assethub with USDC/USDT/DOT

ebma:
Closes #820.


apps/api/src/api/services/quote/core/nabla.ts
```diff
@@ -0,0 +1,79 @@
...
+  effectiveExchangeRate?: string;
+}
+
+async function getNablaSwapOutAmount(
+  apiInstance: { api: ApiPromise },
+  fromAmountString: string,
+  inputTokenPendulumDetails: PendulumTokenDetails,
+  outputTokenPendulumDetails: PendulumTokenDetails
+): Promise<TokenOutData> {
+  return await getTokenOutAmount({
```
gianfra-t: Nit: can we use directly `getTokenOutAmount`?

apps/api/src/api/services/quote/engines/finalize/onramp.ts
```diff
@@ -0,0 +1,76 @@
...
+      if (!output) {
+        throw new APIError({
+          message: "OnRampFinalizeEngine requires bridge output for EVM",
+          status: httpStatus.INTERNAL_SERVER_ERROR
+        });
+      }
+      finalOutputAmountDecimal = new Big(output);
+    }
+
+    if (finalOutputAmountDecimal.lte(0)) {
```
gianfra-t: This should be positive at this stage right? As the the network/vortex fee we take should have been subtracted before this step (before the input to squidrouter or final xcm). 

I'm fine with leaving as defensive but just to be sure I understand.

apps/api/src/api/services/quote/engines/pendulum-transfers/offramp-stellar.ts
```diff
@@ -0,0 +1,54 @@
...
+    }
+  }
+
+  protected async compute(ctx: QuoteContext): Promise<PendulumTransferComputation> {
+    // biome-ignore lint/style/noNonNullAssertion: Context is validated in validate
+    const nablaSwap = ctx.nablaSwap!;
+
+    const fee = new Big(0); // The fee is not paid in the token being transferred
+    const inputAmountDecimal = this.mergeSubsidy(ctx, new Big(nablaSwap.outputAmountDecimal));
+    const inputAmountRaw = this.mergeSubsidyRaw(ctx, new Big(nablaSwap.outputAmountRaw)).toFixed(0, 0);
```
gianfra-t: Nit: let's have one `mergeSubsidy` that returns both units and raw.

packages/shared/src/services/xcm/hydrationToAssethub.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  const info = await transaction.senderAddress("5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY").getXcmFeeEstimate();
+
+  const destinationAmountRaw = info.destination.fee.toString();
+  // The destination fee is always in AssetHub DOT which has 10 decimals
+  const destinationAmountDecimals = multiplyByPowerOfTen(destinationAmountRaw, -10).toString();
+
+  const originAmountRaw = info.origin.fee.toString();
+  // The origin fee is always in HDX which has 12 decimals
+  const originAmountDecimals = multiplyByPowerOfTen(originAmountRaw, -12).toString();
```
gianfra-t: I am a bit confused about hydration fees. I understand that some extra fee is subtracted on Assethub, but we also pay on the sender chain some `HDX`?  I saw [here](https://github.com/pendulum-chain/vortex/blob/ab11d0bc20a3e0a3e189c35c1b5a323d811dcb42/apps/api/src/api/services/transactions/hydration/index.ts#L45), which I believe is the counterpart for the transaction builder, that we pay in the asset itself.

apps/api/src/api/services/quote/engines/finalize/onramp.ts
ebma: Yes it should, it's just a final failsafe.

apps/api/src/api/services/quote/engines/pendulum-transfers/offramp-stellar.ts
gianfra-t: Alright then 👍, I didn't see that other use case.

apps/api/src/api/services/quote/index.ts
```diff
@@ -0,0 +1,92 @@
...
+    });
+
+    const orchestrator = new QuoteOrchestrator();
+    const resolver = new RouteResolver();
+    const strategy = resolver.resolve(ctx);
+    try {
+      await orchestrator.run(strategy, ctx);
+    } catch (error) {
+      logger.error(error instanceof Error ? error.message : String(error));
+      throw new APIError({ message: QuoteError.FailedToCalculateQuote, status: httpStatus.INTERNAL_SERVER_ERROR });
```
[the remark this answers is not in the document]
ebma: I think it's not that important to give messages that fit the error case. As you can see, we are logging the error we receive with `logger.error()` so we can see the problem on the service logs. The returned API error is quite generic and only says that we failed to calculate the quote but that's enough for the average consumer of the API.

Issue #831, opened 2025-09-15
Make widget page automatically create/refresh the quote if no quoteId is passed to it


ebma:
## Context

The widget should be able to automatically create and refresh new quotes if no existing quote ID was passed to it. This brings several benefits, e.g. integrators don't have create a quote beforehand, and users can take as long as they want on the widget, without having the quote expire. 

## TODO

- Look at the parameters defined in https://github.com/pendulum-chain/vortex/issues/828. 
- Change the widget so that it automatically creates quotes based on the URL parameters passed to it.

ebma:
Done in #836.

Issue #828, opened 2025-09-12
Add endpoint to generate a widget URL


ebma:
## Context

With this endpoint, we are able to add more security to the widget in the future. For example, on the endpoint, we can require an api-key, or access-token in order to make it callable only from supported entities. 

## TODO

Add a new POST endpoint '/session/create' that can take the following body parameters:

For passing an existing quote to a widget: 
```
{
   "quoteId": "xxx-xxx",
   "walletAddressLocked": "1x0232r23jnjnfjk2n3fjkn434jknf443jnf4j4nf",
   "externalSessionId": "third-party-session-id-1",
   // "externalTransactionId": "third-party-transaction-id-1",
   // "externalCustomerId": "third-party-customer-id-1"
}
```

For creating a link to the widget that automatically refreshes the quote (without showing the form again)
```
{
   "inputAmount": "100",
   "sourceCurrencyCode": "USD",
   "destinationCurrencyCodeLocked": "BTC",
   "countryCode": "US",
   "paymentMethod": "CARD",
   "walletAddressLocked": "1x0232r23jnjnfjk2n3fjkn434jknf443jnf4j4nf",
   "externalSessionId": "meldSession1",
   // "externalTransactionId": "meldTransaction1",
   // "externalCustomerId": "meldCustomer1"
}
```

The response should contain the following:
```
{
   "widgetUrl": “https://app.vortexfinance.co/widget?rampType=BUY&inputAmount=100&fiat=usd&cryptoLocked=btc&payment=card&walletLocked=1x0232r23jnjnfjk2n3fjkn434jknf443jnf4j4nf
}
```

or for a widget that does not automatically refresh the quote (pass the `quoteId`):
```
{
   "widgetUrl": “https://app.vortexfinance.co/widget?quoteId=xxx-xxx&walletLocked=1x0232r23jnjnfjk2n3fjkn434jknf443jnf4j4nf
}
```

PR #836, opened 2025-09-16
Auto refresh quote widget

gianfra-t:
Issue #831 

Depending on the query params passed, we can load the quote based on the `quoteId` and lock it, in this case there is no refresh behavior.
Alternatively, we can fetch an initial quote based on the parameters and only then, move to `LoadingQuote`. This time, the quote is not locked and an internal process will fetch a new one based on remaining expiration time. This process will be triggered during the `KycComplete` state, and will be active as long as we are in that state.

### Description of changes:
- Implements logic to parse url params.
- Separates current effects to read url params, between `/widget` and regular swap paths.
- On the details step, check the wallet locked matches the connected wallet. As the other params, `walletLocked` is set the first time the component loads and cannot be changed.
- Adjust the states, and add process to automatically refresh the quote on the `KycComplete`. This is the last state before actually registering the ramp.


apps/frontend/src/machines/ramp.machine.ts
```diff
@@ -154,16 +204,43 @@ export const rampMachine = setup({
...
+      invoke: {
+        input: ({ context }) => ({ context }),
+        src: "quoteRefresher"
+      },
       on: {
         PROCEED_TO_REGISTRATION: {
           target: "RegisterRamp"
+        },
+        // TODO we need to define the action to take on quote refresh failure.
+        REFRESH_FAILED: {},
```
gianfra-t, line 233: We can either keep the old quote, or keep retrying the re-fetch. Alternatively we can fail the ramp assuming the backend is down.

gianfra-t:
Thanks @ebma you had some good finds and points to discuss. From bottom to top:
- I'll check why the confirm button didn't set the latest quote. I triple-checked that the state machine was updating it's context so it must be fetching it from a stale react state.
- Add what we discussed about the initial quote fetch error, due to invalid params.
- About refreshing the quote params
     
     > I think we should try to make sure that the params are always applied anew on page reload.
     I thought it would be better to "fix" the quote params from the initial load, because this way it is easier to handle redirects from our partners and also unintended reloads/persistance from the user side. But if replacing an existing quote with new params is something we need, we can make the changes.
     
- Finally for the sep case... I missed it completely. Since in the initial changes the quote was refreshing on the details page only. I like better the idea of going back to the details page, but mostly to respect the states we've defined and avoid having too much conditional logic on our components. After all, from the user's perspective, the actions to take are exactly the same!

apps/frontend/src/machines/ramp.machine.ts
gianfra-t: It won't keep refreshing right now, but we can make it so by just re-entering the same state like on UPDATE_QUOTE, except no quote is modified. Let's do that, and then the UI will handle the actions by blocking the confirm.

gianfra-t:
Pushed a few extra things besides the above. Most importantly, `BRL` onramp doesn't require a wallet connection now. And the parameter `walletLocked` is now taken into account to pre-fill and fix the onramp address.

apps/frontend/src/components/RampSubmitButton/RampSubmitButton.tsx
```diff
@@ -27,16 +29,27 @@ const useButtonContent = ({ toToken, submitButtonDisabled }: UseButtonContentPro
...
+      machineState: state.value,
+      quote: state.context.quote,
+      rampPaymentConfirmed: state.context.rampPaymentConfirmed,
+      rampState: state.context.rampState,
+      walletLocked: state.context.walletLocked
+    })
+  );
 
   return useMemo(() => {
+    if (walletLocked && accountAddress && getAddressForFormat(accountAddress, 0) !== walletLocked) {
```
ebma, line 47: We would need to normalize both addresses, just to be safe. 
```suggestion
    if (walletLocked && accountAddress && getAddressForFormat(accountAddress, 0) !== getAddressForFormat(walletLocked, 0)) {
```

apps/frontend/src/components/RampSubmitButton/RampSubmitButton.tsx
```diff
@@ -171,6 +185,9 @@ export const RampSubmitButton = ({ className }: { className?: string }) => {
   const toToken = isOnramp ? getOnChainTokenDetailsOrDefault(selectedNetwork, onChainToken) : getAnyFiatTokenDetails(fiatToken);
 
   const submitButtonDisabled = useMemo(() => {
+    if (walletLocked && accountAddress && getAddressForFormat(accountAddress, 0) !== walletLocked) {
```
ebma, line 188: Same here.

apps/api/src/api/services/ramp/quote.service/validation-helpers.ts
```diff
@@ -30,6 +30,7 @@ export function validateAmountLimits(
   limitType: "min" | "max",
   operationType: RampDirection
 ): void {
+  return; // Temporarily disable limit checks
```
ebma, line 33: Let's remove that check again.

ebma:
> I thought it would be better to "fix" the quote params from the initial load, because this way it is easier to handle redirects from our partners and also unintended reloads/persistance from the user side. But if replacing an existing quote with new params is something we need, we can make the changes.

Yeah, I think it's not easy to get this right. Especially if we consider the user could open different instances of the widget in different tabs. However, it just came to me that, once we have our endpoint for generating a widget URL, we can make it always pass a session ID. It's in the description of https://github.com/pendulum-chain/vortex/issues/828 already. It defines an `externalSessionId`. We can adjust the logic of the widget page so that it will override a previous quote with a fresh one on a page load if the `externalSessionId`s don't match. And if they do match, we do nothing. WDYT? This means it would be okay to keep this logic as is for now but adjust it once we have the widget endpoint.

gianfra-t:
@ebma the last commit adds the behavior we were discussing with `externalSessionId`. It does not make any assumptions about how the `externalSessionId` looks or what to do with it besides storing it, but we can take care of that in  https://github.com/pendulum-chain/vortex/issues/828.

apps/frontend/src/machines/ramp.machine.ts
```diff
@@ -103,6 +105,9 @@ export const rampMachine = setup({
       await new Promise(resolve => setTimeout(resolve, 30000));
       await refetchQuote(quote, partnerId, event => self.send(event));
     },
+    reloadKeepingParams: () => {
```
ebma, line 108: This seems to be unused?

apps/frontend/src/machines/ramp.machine.ts
```diff
@@ -13,29 +13,65 @@ import { aveniaKycMachine } from "./brlaKyc.machine";
...
+const refetchQuote = async (
+  quote: QuoteResponse,
+  partnerId: string | undefined,
+  sendBack: (event: RampMachineEvents) => void
+) => {
+  const now = Date.now();
+  const expires = new Date(quote.expiresAt).getTime();
+  const secondsLeft = Math.round((expires - now) / 1000);
+
+  console.log("Quote expires in", Math.round((expires - now) / 1000), "seconds");
```
ebma, line 52: Let's remove this log again.

ebma (approved):
It finally seems to work as expected 🙏 I was not able to test the Mykobo/SEP-24 flow for when the quote changes and we need to lead the user to Mykobo again. The quotes I had always had the same output amount so it didn't trigger.

I think we shouldn't show this "Quote expires in xxxx" text when we are in auto-refetch  mode though as it's putting unnecessary stress on the user.

PR #843, opened 2025-09-18
add squidrouter's quote id param to status request

gianfra-t:


ebma (approved):
Looks good to me 👌 did you do a dry run by e.g. creating an onramp and checking if the respective fields were populated as expected?

Issue #826, opened 2025-09-12
Change fields of quote and ramp endpoints


ebma:
## Context

Integrators shouldn't need to call both the 'get quote' and 'get ramp status' endpoint in order to get all the details of a pending ramp/transaction. Also, additional parameters are required. 

## TODO

Change the request of the quote endpoint. Allow passing the paymentMethod, a country code and the (blockchain) network. 
If a country code and network parameter are present in the request to the quote endpoint, the "from" and "to" field of the created quote should be derived from that. E.g. if the passed `countryCode` is 'BR', the `network` is 'polygon' and the `rampType` is 'BUY', the quote should be `"from": "pix"` and `"to": "polygon"`. Store the country code and network on the quote, too. 

Change the response of the quote endpoint. Unfold the existing `fee` field and instead of nesting it in `fee`, expand each item to separate line items with the suffix 'FeeFiat'. Also add the USD representation of the fee. Note that this data is already available on the `metadata` field of the quote as the `usdFeeStructure`:
```
"anchorFeeUsd": "0.75",
"networkFeeUsd": "0.507286",
"partnerFeeUsd": "0.00",
"totalFeeUsd": "3.81",
"vortexFeeUsd": "2.55",
"processingFeeUsd":"3.30" // anchor + vortex fee
```
 
----- 

Change the request of the 'register ramp' endpoint. Add the following parameters:
- `sessionId`: A session identifier passed from an integrator to the widget. 

Change the response of the 'get ramp' endpoint. Add the following parameters. 
- `status`: Similar to the 'currentPhase' field but simpler. Can have values 'PENDING', 'FAILED', and 'COMPLETE'
- `paymentMethod`: The payment method passed to the quote
- usd and fiat fees as included in the quote
- `sessionId`: The identifier passed to the register ramp endpoint
- `walletAddress`: The destination address where the funds will land
- `inputAmount`: Same as quote
- `outputAmount`: Same as quote
- **DON'T** return the `unsignedTxs` by default. Add an optional parameter `?showUnsignedTxs=true` to the endpoint so that this field is only returned when the parameter is passed and `true`.

PR #869, opened 2025-10-14
Change fields of quote and ramp endpoints

Sharqiewicz:
📋 Summary
This PR refactors the Quote and Ramp flows to explicitly include payment method, network, and country code data, flattens the fee model, updates backend models and migrations, and aligns frontend/UI with the new schema for clearer integration and display.

🎯 Key Changes

 Added paymentMethod, network, and countryCode fields to Quote creation and Ramp endpoints, ensuring corridor context is explicit in API responses.

 Flattened the fee model across backend and frontend (including RampFeeCollapse) for simpler representation; ensured usdFeeStructure support for Monerium.

 Refactored services to consistently derive {from,to} from {countryCode,network} and guarantee non-null network in QuoteTicket; added UI and docs updates for better UX and clarity.

📛 Breaking changes: Yes

 network is now required (non-nullable) in QuoteTicket and related API responses.

 Fee model flattened, removing nested structures — client updates required.

🆙 Dependencies: None

⚙️ Configuration: Added new DB migration for quote_tickets to include and enforce new fields.

✅ Review Focus Areas

Confirm network derivation logic never returns null and aligns with currency-country mapping rules.

Validate all UI components consume the flattened fee structure correctly, especially where nested data was previously expected.

Verify paymentMethod mapping consistency across Quote and Ramp endpoints.

Check that migrations correctly enforce network non-null constraints without data loss.

Closes https://github.com/pendulum-chain/vortex/issues/826. 

The PR includes improved copy animation:

https://github.com/user-attachments/assets/27500c9c-d474-44d5-bac7-6d1621bdbd72

### DIFF
--- a/apps/frontend/src/machines/ramp.machine.ts
+++ b/apps/frontend/src/machines/ramp.machine.ts
@@ -214,14 +214,7 @@ export const rampMachine = setup({
           assign({
             externalSessionId: ({ event }) => event.externalSessionId
           })
-        ],
-        // Assumed to be a new session, so we reset everything and reload the page.
-        // This will reload the new parameters and fetch a new quote.
-        guard: ({ context, event }) =>
-          event.externalSessionId !== undefined &&
-          context.externalSessionId !== undefined &&
-          event.externalSessionId !== context.externalSessionId,
-        target: ".Idle"
+        ]
       }
     ],
     SET_GET_MESSAGE_SIGNATURE: {


ebma:
Next time, please link the ticket that this PR is supposed to close in the description.

ebma (commented):
I like the changes, great job 👍 

How about we make sure that the countryCode and the network are always filled for a quote, with a reverse derivation? You already added a derivation of { countryCode + network } -> { from , to }. How about we also add a derivation of { from, to } -> { countryCode + network }, so that we can fill the QuoteTicket model with these attributes in case the creator of the quote didn't provide them.

PR #887, opened 2025-10-20
Use KYC sandbox

gianfra-t:
### Description
This PR allows to configure a sandbox environment for both the UI and backend. The tester will be able to create a KYC and test ramp flows.

### Changes
- Adds dynamic selection between production and sandbox environments of partners.
- Modifies Monerium's logic and related transactions to work on Polygon testnet (`Amoy`).
- Mocks squidrouter transactions when in sandbox mode.
- Transitions phases from `initial` to `complete` after a short delay, mocking a successful ramp.

### DIFF
--- a/apps/frontend/src/services/monerium/moneriumAuth.ts
+++ b/apps/frontend/src/services/monerium/moneriumAuth.ts
@@ -17,7 +17,9 @@ export class MoneriumAuthError extends Error {
 
 export const MONERIUM_MINT_NETWORK = config.isSandbox ? "amoy" : "polygon";
 const MONERIUM_MINT_NETWORK_CHAIN_ID = config.isSandbox ? 80002 : 137;
-const VORTEX_APP_CLIENT_ID = import.meta.env.VITE_MONERIUM_CLIENT_ID || "e7b56f39-b0ff-11f0-a4ad-fabb3106d2e3";
+const VORTEX_APP_CLIENT_ID =
+  import.meta.env.VITE_MONERIUM_CLIENT_ID ||
+  (config.isSandbox ? "e7b56f39-b0ff-11f0-a4ad-fabb3106d2e3" : "eac7a71a-414d-11f0-bea7-ce527adad61b");
 // Use custom API URL if provided, otherwise use default sandbox/dev endpoints
 const MONERIUM_API_URL =
   import.meta.env.VITE_MONERIUM_API_URL || (config.isSandbox ? "https://api.monerium.dev" : "https://api.monerium.app");


apps/api/src/api/services/transactions/moneriumEvmOnrampTransactions.ts
```diff
@@ -18,12 +18,14 @@ import {
...
 import Big from "big.js";
 import { encodeFunctionData } from "viem";
+import { SANDBOX_ENABLED } from "../../../constants/constants";
 import erc20ABI from "../../../contracts/ERC20";
 import { QuoteTicketAttributes } from "../../../models/quoteTicket.model";
 import { multiplyByPowerOfTen } from "../pendulum/helpers";
 import { StateMetadata } from "../phases/meta-state-types";
 import { encodeEvmTransactionData } from "./index";
 
+const MONERIUM_MINT_NETWORK = SANDBOX_ENABLED ? Networks.PolygonAmoy : Networks.Polygon;
```
ebma: I think this is already defined elsewhere, so we can just import it here.

apps/frontend/src/services/anchor/sep24/first.ts
```diff
@@ -22,6 +22,7 @@ export async function sep24First(
     ...(usesMemo && { account: ANCLAP_sep10Account })
   };
 
+  console.log("Initiating SEP-24 with params:", params);
```
ebma: Let's remove this

packages/shared/src/helpers/networks.ts
```diff
@@ -13,7 +13,8 @@ export enum Networks {
   Polygon = "polygon",
   Moonbeam = "moonbeam",
   Pendulum = "pendulum",
-  Stellar = "stellar"
+  Stellar = "stellar",
+  PolygonAmoy = "polygonAmoy"
```
ebma: Maybe we should add this type to the enum only conditionally based on the sandbox config so that it's not automatically shown on the network selector

packages/shared/src/tokens/evm/config.ts
```diff
@@ -72,6 +72,18 @@ export const evmTokenConfig: Record<EvmNetworks, Partial<Record<EvmToken, EvmTok
       type: TokenType.Evm
     }
   },
+  [Networks.PolygonAmoy]: {
+    [EvmToken.USDC]: {
```
ebma: Let's remove this token since it's not really usable for a transfer as far as I understand.

ebma (commented):
Looks good overall. Left a few comments, and let's also have support for Assethub.

apps/api/src/api/services/transactions/moneriumEvmOnrampTransactions.ts
gianfra-t: Yes same var is defined on the frontend. We could define it on the `shared` package, but I find this more cumbersome, since shared can be used on both node and browser. So reading env variables requires extra handling.

PR #912, opened 2025-10-31
Add account creation to Vortex

### LINKED ISSUE #882: Account creation on Vortex
## Context
To enforce KYC status and limits, txn history it is necessary to introduce account creation in the Vortex app. 

## Todo
* Add a email login/signup once the user get's the quote
* Login/Signup should be one more card in the flow, card should have an input field for email and a "Submit" button
* On entering the email and submitting user should receive an code using which the user can login, after login they should land on details page. Ideally the code entering screen should have only fields for code and not button if the code is correct we automatically take the user to details page if not show and error saying "Incorrect code"
* Email should be used an unique identifier on the app for enforcing KYC status, Txn limits and also showing txn history
* The user should recieve an email for the following conditions
    * KYC status of Monerium
    * KYB status of Avenia
    * KYC status of MYKOBO 

## Note
* As of now this ticket does not include building a dashboard to show the limits and KYC status level, we can take that in the next iteration
* Following are the services that i found that can be used for AUTH (TBD by Techies)
    *  https://auth0.com/ - 25k monthly active users for free
    * Supabase also have Auth0 solution, with 50k user tier. I guess this would be easier to set as we already have supabase database. I assume.

ebma:
Closes #882.

This pull request introduces Supabase-based authentication to the API, including OTP (one-time password) flows, token verification, and user session management. It adds new authentication endpoints, middleware for extracting and verifying user identity from requests, and propagates the authenticated `userId` throughout the business logic and data models. Several routes now optionally or mandatorily require authentication, and the system is prepared for future endpoint protection. Additionally, the codebase is updated to store and utilize the `userId` in key models and service layers.

**Supabase Authentication Integration**

* Added Supabase configuration variables to `.env.example` and included `@supabase/supabase-js` as a dependency in `package.json`. [[1]](diffhunk://#diff-0ab7147cd8ad1eddcd936d4014fc9778eb565a51aed784bbdd32ffcba52cf22bR13-R17) [[2]](diffhunk://#diff-2c40985d6d91eed8ae85ec1c8e754a85984ee32e156a600d2b7a467423d7e338R14)
* Implemented `SupabaseAuthService` with methods for checking user existence, sending and verifying OTPs, verifying and refreshing tokens, and retrieving user profiles. [[1]](diffhunk://#diff-68cd0c0d1e099e39d947e3552961b9c234b9eea05628ad24217dbc343984254eR1-R123) [[2]](diffhunk://#diff-22746d1708a158bb169c7587c7d77e778c4a79afecc1312234bae67cc7eb3525R1)
* Created `AuthController` with endpoints for checking email registration, requesting OTP, verifying OTP, refreshing tokens, and verifying tokens. [[1]](diffhunk://#diff-11592c2b0b4781c39969ce572ad23ba0008c678cd60b78fe49d12a1fe10fdbacR1-R162) [[2]](diffhunk://#diff-10b45ae5cd3f68c532347e19e74f051da4c93fa5a8e9b2f71d44284de15c2d21R1-R12)
* Added `/v1/auth` routes for authentication endpoints and registered them in the main API router. [[1]](diffhunk://#diff-10b45ae5cd3f68c532347e19e74f051da4c93fa5a8e9b2f71d44284de15c2d21R1-R12) [[2]](diffhunk://#diff-c85c7f09aa62a7e695c63f32c5509b7dd3629b5cae5f06113a76082fff7a640bR6) [[3]](diffhunk://#diff-c85c7f09aa62a7e695c63f32c5509b7dd3629b5cae5f06113a76082fff7a640bR149-R158)

**Authentication Middleware and Route Protection**

* Added `requireAuth` and `optionalAuth` middleware to extract and verify user identity from Bearer tokens, attaching `userId` to the request.
* Updated BRLA, Quote, and Ramp routes to use `optionalAuth` middleware, allowing endpoints to access `userId` if provided. [[1]](diffhunk://#diff-cc056071743601e7f06b74c1da89586fcd2164f6d5cab8c22af67ad340b1faacR3) [[2]](diffhunk://#diff-cc056071743601e7f06b74c1da89586fcd2164f6d5cab8c22af67ad340b1faacL17-R28) [[3]](diffhunk://#diff-96ee41ea52d07414b976847906dbb10ba171ad0bd90c983e44f920e55513d762R5) [[4]](diffhunk://#diff-96ee41ea52d07414b976847906dbb10ba171ad0bd90c983e44f920e55513d762R46) [[5]](diffhunk://#diff-96ee41ea52d07414b976847906dbb10ba171ad0bd90c983e44f920e55513d762R105) [[6]](diffhunk://#diff-016833e3a0906840d05b278a2ecda8339cc3c85b1e7ed851be231bca9320182fR3) [[7]](diffhunk://#diff-016833e3a0906840d05b278a2ecda8339cc3c85b1e7ed851be231bca9320182fL32-R33)

**User Identity Propagation**

* Modified controllers (`brla.controller.ts`, `quote.controller.ts`, `ramp.controller.ts`) to store `userId` from the request in database records and service calls. [[1]](diffhunk://#diff-af106d92265e78561239a7f7a24d59b6286793b533d900e4540cd9ffb3729600R206-R213) [[2]](diffhunk://#diff-af106d92265e78561239a7f7a24d59b6286793b533d900e4540cd9ffb3729600L220-R226) [[3]](diffhunk://#diff-af106d92265e78561239a7f7a24d59b6286793b533d900e4540cd9ffb3729600R326-R342) [[4]](diffhunk://#diff-c5494f8a1a22eae1220bc526283c72ea801b78ee3d4ed78a6bb0ab6aba8e42b2L51-R52) [[5]](diffhunk://#diff-c5494f8a1a22eae1220bc526283c72ea801b78ee3d4ed78a6bb0ab6aba8e42b2L88-R90) [[6]](diffhunk://#diff-a0ecd784899c1184f33e112c97845c5123cc344f8c0a7c0368bb2d98886117adL40-R41)
* Updated service and type definitions to support `userId` as part of the request context and data models. [[1]](diffhunk://#diff-629d2bcfaec19e57ad2e490bd2d8aa5c76da4c8042c2899e0d9348170496a41dL94-R94) [[2]](diffhunk://#diff-36d4144483c2e0901dd05ba5b6e6dcde1a3fb4af9a556e5b603fe95d753ab4edL146-R147)

These changes collectively enable secure, user-specific flows across the API, laying the groundwork for robust authentication and authorization using Supabase.


ebma:
@copilot open a new pull request to apply changes based on the comments in [this thread](https://github.com/pendulum-chain/vortex/pull/912#pullrequestreview-3577519360)

apps/api/src/api/controllers/auth.controller.ts
```diff
@@ -0,0 +1,162 @@
...
+          error: "Email and token are required"
+        });
+      }
+
+      const result = await SupabaseAuthService.verifyOTP(email, token);
+
+      // Sync user to local database (upsert)
+      await User.upsert({
+        email: email,
+        id: result.user_id
```
gianfra-t: Is the `upsert` only for handling new users? I assume the `userId` should never change but I may be wrong.

apps/frontend/src/services/api/api-client.ts
```diff
@@ -14,10 +15,14 @@ export const apiClient: AxiosInstance = axios.create({
   timeout: 30000
 });
 
-// Add request interceptor for common headers
+// Add request interceptor for common headers and auth token
 apiClient.interceptors.request.use(
   config => {
-    // Add any common headers here
+    // Add Authorization header if user is authenticated
```
gianfra-t: Do you know, if the header refreshes after the user gets authenticated? 
What I mean is, if the `apiClient` initially starts with an authenticated state, do the headers get added after the Auth state changes?

apps/frontend/src/machines/ramp.machine.ts
```diff
@@ -278,6 +332,78 @@ export const rampMachine = setup({
...
+            callbackUrl: ({ event }) => event.callbackUrl,
+            partnerId: ({ event }) => event.partnerId,
+            walletLocked: ({ event }) => event.walletLocked
+          })
+        }
+      }
+    },
+    EnterOTP: {
+      on: {
+        CHANGE_EMAIL: {
```
gianfra-t: What do we want to allow changing the email at this stage? I assume Supabase already created the user.

apps/api/src/database/migrations/022-add-user-id-to-entities.ts
```diff
@@ -0,0 +1,203 @@
...
+    name: "idx_quote_tickets_user_id"
+  });
+
+  // Add user_id to ramp_states
+  await queryInterface.addColumn("ramp_states", "user_id", {
+    allowNull: true,
+    type: DataTypes.UUID
+  });
+
+  await queryInterface.sequelize.query(`
```
gianfra-t: Can we keep the userId relation for this one (and any that allows for `null`) empty instead of using the dummy? Or do you think there is a benefit of using the dummy user here?

apps/frontend/src/services/api/api-client.ts
ebma: Not sure I understand the question. Is the question whether the apiClient still works as expected, if it started in the authenticated state, then the user logs out and back in again?
The function inside the interceptor is a callback function so it will always use the very latest return values of  `AuthService.getTokens()` whenever a request is made. Does that answer the question?

apps/frontend/src/machines/ramp.machine.ts
ebma: It's for the 'Use a different email' feature, see screenshot. And the account/profile is only truly created after the user entered the OTP from the received email.

image

apps/api/src/api/services/auth/supabase.service.ts
```diff
@@ -0,0 +1,131 @@
...
+      // Query the profiles table directly for better performance
+      const { data, error } = await supabaseAdmin
+        .from("profiles")
+        .select("id")
+        .eq("email", email)
+        .single();
+
+      if (error) {
+        // If error is "PGRST116" (no rows returned), user doesn't exist
+        if (error.code === "PGRST116") {
```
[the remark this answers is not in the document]
ebma: Apparently it can happen for Supabase queries, see [this](https://www.answeroverflow.com/m/1325167977342963814) thread. Seems to be a postgres thing.

apps/api/src/database/migrations/022-add-user-id-to-entities.ts
ebma: Hmm, well maybe we should just allow it to be null on each table. You are right that the dummy user is not great. I used it so we can enforce the non-null constraint that is logical on some of the tables but now that I think of it, it might be better if it can be null and we just add it later..

Sharqiewicz:
@ebma Please let me know when it's ready for review again 🙏

apps/api/src/database/migrations/022-add-user-id-to-entities.ts
ebma: Some of them did, some didn't. But I adjusted it so that the userId is nullable on all the other relations and the code checks if necessary.

Sharqiewicz:
@ebma Ok, thanks. I'll review it today

Sharqiewicz:
@ebma Do we want to add the back button to the email auth flow?

PR #955, opened 2025-11-28
Remove polygon funding when not needed on the flow

gianfra-t:
These changes solve an unnecessary funding of the polygon ephemeral address, that was being funded for other ramps besides EURO onramp.


Sharqiewicz (approved):
Looks good! I like you extracted the checks to separate functions.
The only logic-change I see is checking the inputCurrency and it's correct and fixes our issue.

Issue #966, opened 2025-12-12
Configure dynamic pricing for ramps


prayagd:
## Context
In order to provide better pricing for both onramps and offramps, this ticket is to implement it

## Todo
* Configure dynamic pricing in a defined range
* The pricing should start from middle of the configured range
* If the quote is consumed, we reduce the quote by 0.3 BPS
* If the quote is not consumed within 10 minutes, we increase the quote by 0.3 BPS
* The above should be repeated until the max or min of the range is reached.

## Note
* Following parameters should be configurable
    * Range i.e Min and Max
    * Difference (for ex. 0.3 BPS)
    * Time (For ex. 10 mins)

PR #970, opened 2025-12-15
Dynamic discount

gianfra-t:
Closes: #966


apps/api/src/api/services/quote/engines/discount/helpers.ts
```diff
@@ -49,33 +62,80 @@ export async function resolveDiscountPartner(ctx: QuoteContext, rampType: RampDi
...
+
+  const partnerState = partnerDiscountState.get(partnerId);
+  const now = new Date();
+
+  if (!partnerState) {
+    partnerDiscountState.set(partnerId, { difference: new Big(0), lastQuoteTimestamp: now });
+    return new Big(0);
+  }
+
+  if (!partnerState.lastQuoteTimestamp) {
```
gianfra-t: We are in this state after a quote is consumed (and difference is increased).

apps/api/src/api/services/quote/engines/discount/helpers.ts
```diff
@@ -49,33 +62,80 @@ export async function resolveDiscountPartner(ctx: QuoteContext, rampType: RampDi
...
 
   // For offramps, we need to invert the oracle price
   // Oracle price is FIAT-USD, so for offramps we want USD-FIAT
   const effectivePrice = isOfframp ? new Big(1).div(oraclePrice) : oraclePrice;
 
-  // Apply target discount to the rate
-  const discountedRate = effectivePrice.mul(new Big(1).plus(targetDiscount));
+  // Apply target discount to the rate, adjusting first for dynamic penalty
+  const adjustedTargetDiscount = new Big(targetDiscount).minus(getAdjustedDifference(partnerId));
+  const discountedRate = effectivePrice.mul(new Big(1).plus(adjustedTargetDiscount));
```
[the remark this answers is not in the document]
gianfra-t: Variable is now not a penalty.

apps/api/src/api/services/quote/engines/discount/helpers.ts
```diff
@@ -49,33 +73,86 @@ export async function resolveDiscountPartner(ctx: QuoteContext, rampType: RampDi
...
+export function getAdjustedDifference(partner?: ActivePartner): Big {
+  if (!partner?.id) {
+    return new Big(0);
+  }
+
+  const partnerState = partnerDiscountState.get(partner.id);
+  const now = new Date();
+
+  // Use partner's min caps if available, otherwise fall back to constants
+  const maxCap = partner.maxTargetDiscount ?? MAX_DIFFERENCE_CAP;
```
ebma: I think we shouldn't fall back to the constants if it's not available. We should rather fall back to the `partner.targetDiscount` field if available.

PR #999, opened 2026-01-08
Validate startRamp

gianfra-t:
### About
Improve `startRamp` validations of additional data. Does not allow a ramp to be started if data is missing.


apps/api/src/api/services/ramp/ramp.service.ts
```diff
@@ -976,6 +978,31 @@ export class RampService extends BaseRampService {
     }
   }
 
+  private validateRampStateData(rampState: RampState, quote: QuoteTicket): void {
+    if (rampState.type === RampDirection.SELL) {
+      if (rampState.from === "assethub" && !rampState.state.assethubToPendulumHash) {
```
Sharqiewicz: Please use Networks.Assethub enum

PR #1031, opened 2026-01-22
Fix invalid typecheck

gianfra-t:


Sharqiewicz:
The deploy/netlify failing seems to be not caused by the changes

PR #1043, opened 2026-01-27
[READY] fix: Back button KYC/KYB Flows

Sharqiewicz:
- Fix BackButton appearance when user enters with already generated quote (doesn't show the button at all)

- Fix using BackButton in BRL KYC/KYB flow skips verification


ebma (approved):
Seems to work ✔️

PR #1065, opened 2026-02-12
[READY] Fix URL Params

Sharqiewicz:
We could create a zustand store for storing the tokens data but we need the data also in the `apps/api` for the `getTokenUSDPrice`. That's why we use an object to store the data and `useSyncExternalStore` on the frontend to reactively subscribe to it.


ebma (approved):
Looks good, let's go 🚀

PR #1109, opened 2026-04-02
[READY] Replace axios with native fetch wrapper

Sharqiewicz:
### Summary

- Replace axios with native fetch across the entire codebase - frontend services, shared package (SquidRouter, dynamic EVM tokens)
- Introduce ApiError class and isApiError type guard as typed replacements for axios.isAxiosError and AxiosError
- Wire AbortSignal from TanStack Query's queryFn context through to fetch() - queries for fiat accounts, ramp history, and fee comparison now cancel in-flight requests on unmount or stale invalidation
- Replace SquidRouter's axios instance with a dedicated squidFetch helper that separates HTTP error handling from response validation, preserving HTTP status codes on thrown errors


ebma (approved):
I added axios v1.15.0 to the overrides so it's patched in the peer dependencies, even if our main apps don't use it.

Sharqiewicz:
thanks @ebma

PR #1138, opened 2026-05-11
[READY] Alfredpay currency limits

Sharqiewicz:
Adds dynamic AlfredPay limit enforcement for USD/MXN/COP quotes, keyed by customer type × stablecoin × direction. Replaces hardcoded static caps with live limits fetched from AlfredPay's /configurations endpoint, refreshed daily, with a hardcoded fallback snapshot.

####  Shared types (packages/shared)
  - New reusable primitives: AmountLimits (decimal {min, max}) and RawAmountLimits (raw {minRaw, maxRaw}).
  - New AlfredpayLimitsTable type - full onramp/offramp × USDC/USDT × INDIVIDUAL/BUSINESS matrix.
  - New AlfredpayCustomerType enum and AlfredpayStablecoinKey ("USDC" | "USDT").
  - BaseFiatTokenDetails gains optional alfredpayLimits; USD/MXN/COP get hardcoded fallback values (May 11 2026 snapshot).
  - QuoteResponse gains optional alfredpayInputLimits: AmountLimits.
  - isAlfredpayToken widened to type-guard RampCurrency -> FiatToken.
  - New AlfredpayConfigPair + getAllConfigs() on the AlfredPay API service.

####  Backend (apps/api)
  - AlfredpayLimitsService - singleton that boots on server start, refreshes the cache daily, and falls back to hardcoded values when
  the upstream fetch fails. Wildcard rows (typeCustomer: null) populate both customer types but specific rows always win.
  - lookupAlfredpayCustomerType resolves user -> customer type (defaults to INDIVIDUAL when anonymous or no KYC row).
  - resolveAlfredpayQuoteLimits returns axes + decimal min/max for a quote, or null for non-AlfredPay quotes.
  - New applyAlfredpayLimits(ctx, amount) helper sets ctx.alfredpayInputLimits and validates; both finalize engines (onramp.ts,
  offramp.ts) collapse to a single call with legacy validateAmountLimits fallthrough.
  - QuoteContext and QuoteResponse builders surface alfredpayInputLimits (both persisted + in-memory skipPersistence paths).

#### Frontend (apps/frontend)
  - useRampValidation consumes quote.alfredpayInputLimits when present and validates inputAmount against those bounds; falls back to
  legacy min/maxBuy/SellAmountRaw for BRL/EURC. No new fetch - rides the existing quote response.
  - Validator branches collapsed: 4 ternaries in validateOfframp consolidated into one named check object.
  - Dropped dead inputAmount && ... / inputAmount ? ... : "0" guards (inputAmount is always a Big).

###  Why

Static, single-axis caps no longer reflect reality - AlfredPay's actual limits vary by Individual/Business KYC status, by USDC vs USDT. This change pulls them from the source of truth without coupling the request path to AlfredPay's API (cache + fallback), and surfaces them to the frontend so users see accurate bounds and localized error messages.

### DIFF
--- a/apps/api/src/api/services/alfredpay/alfredpay-limits.service.ts
+++ b/apps/api/src/api/services/alfredpay/alfredpay-limits.service.ts
@@ -1,23 +1,20 @@
 import {
   AlfredpayApiService,
   AlfredpayConfigPair,
-  AlfredpayCustomerKey,
   AlfredpayCustomerType,
-  AlfredpayLimitsBucket,
   AlfredpayStablecoinKey,
   FiatToken,
   getAnyFiatTokenDetails,
-  isAlfredpayToken
+  RampDirection,
+  RawAmountLimits
 } from "@vortexfi/shared";
 import Big from "big.js";
 import logger from "../../../config/logger";
 
-export type AlfredpayLimitsDirection = "onramp" | "offramp";
-
 /** Refreshed once on startup, then daily. Limits don't change often, so this avoids beating the API. */
 const REFRESH_INTERVAL_MS = 24 * 60 * 60 * 1000;
 
-const CUSTOMER_TYPES: AlfredpayCustomerKey[] = ["INDIVIDUAL", "BUSINESS"];
+const CUSTOMER_TYPES: AlfredpayCustomerType[] = [AlfredpayCustomerType.INDIVIDUAL, AlfredpayCustomerType.BUSINESS];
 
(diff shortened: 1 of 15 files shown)


ebma:
@Sharqiewicz once your PRs are ready for review, please not only change the title but also add me and @gianfra-t as reviewers or let us know on Slack. Otherwise, this gets missed and gets stale.

ebma (approved):

PR #1146, opened 2026-05-14
Add backup transactions validation for pre-signed ephemeral transactions

gianfra-t:
#### About

Adds a validation check to ensure any ephemeral-signed transaction has the required number of backup transactions, with the right nonce sequence. 

Note: it does NOT validate the correctness of the data itself.

#### Additional fixes
Adds delays before checking balances in subsidy phases. This ensure the balance from a previous `distributeFee` phase is properly settled.


apps/api/src/api/services/transactions/validation.ts
```diff
@@ -72,6 +74,39 @@ function getTransactionTypeForPhase(phase: RampPhase | CleanupPhase): EphemeralA
   }
 }
 
+function validateBackupTransactions(tx: PresignedTx, ephemerals: { [key in EphemeralAccountType]: string }) {
+  const isEphemeralSigner =
+    (ephemerals.EVM && tx.signer.toLowerCase() === ephemerals.EVM.toLowerCase()) ||
```
Sharqiewicz: We could just use:
```
const signer = tx.signer.toLowerCase();
  const isEphemeralSigner = Object.values(ephemerals).some(addr => addr && addr.toLowerCase() === signer);
```
wdyt?

Sharqiewicz: I'm not sure if Substrate/Stellar are case insensitive, but I assume we don't need to count them in

apps/api/src/api/services/transactions/validation.ts
```diff
@@ -72,6 +74,39 @@ function getTransactionTypeForPhase(phase: RampPhase | CleanupPhase): EphemeralA
...
+
+  const additionalTxs = tx.meta?.additionalTxs;
+  if (!additionalTxs || Object.keys(additionalTxs).length < NUMBER_OF_PRESIGNED_TXS - 1) {
+    throw new APIError({
+      message: `Transaction for phase ${tx.phase} must include at least ${NUMBER_OF_PRESIGNED_TXS - 1} backup transactions in meta.additionalTxs`,
+      status: httpStatus.BAD_REQUEST
+    });
+  }
+
+  const backupNonces = Object.values(additionalTxs)
```
Sharqiewicz, line 92: What do you think about 
```
 const actualNonces = Object.values(additionalTxs).map(b => b.nonce).sort((a, b) => a - b);
  const expectedNonces = Array.from({ length: EXPECTED_BACKUP_COUNT }, (_, i) => tx.nonce + 1 + i);

  if (!actualNonces.every((n, i) => n === expectedNonces[i])) {
    throw new APIError({
      message: `Transaction for phase ${tx.phase} has invalid backup nonce sequence. Expected ${expectedNonces.join(",")}, got
  ${actualNonces.join(",")}`,
      status: httpStatus.BAD_REQUEST
    });
  }
```

apps/api/src/api/services/transactions/validation.ts
```diff
@@ -20,6 +20,8 @@ import logger from "../../../config/logger";
 import { SANDBOX_ENABLED } from "../../../constants/constants";
 import { APIError } from "../../errors/api-error";
 
+const NUMBER_OF_PRESIGNED_TXS = 5;
```
Sharqiewicz: I think it's duplicated in validation.ts:23 and packages/shared/src/helpers/signUnsigned.ts:23.

apps/api/src/api/services/transactions/validation.ts
gianfra-t: Sure I guess that also works. The key thing is we only check for the ones the ephemeral signed

apps/api/src/api/services/transactions/validation.ts
gianfra-t: True, we may export that constant to have a single source of truth 👍

apps/api/src/api/services/phases/handlers/subsidize-post-swap-evm-handler.ts
```diff
@@ -61,6 +61,8 @@ export class SubsidizePostSwapEvmPhaseHandler extends BasePhaseHandler {
         );
       }
 
+      await new Promise(resolve => setTimeout(resolve, 15000));
+
```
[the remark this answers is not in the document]
gianfra-t: Accept the inefficiency for now.

PR #1157, opened 2026-05-21
Fix Monerium flow after API change

gianfra-t:


ebma (commented):
## Independent Code Review

Reviewed all 11 modified files against `staging`. Below: my own findings + verdicts on the Copilot review (#pullrequestreview-4338359773). Accepted improvements have been pushed as commit `99710cd60`.

### Verdicts on Copilot's Comments

**1. Extract `"EURe"` symbol to a constant** — **Skip.** The literal has a single consumer (`squidRouterPhaseHandler` balance check) and no UI/display surface. Adding a shared constant for a single internal comparison would be over-engineering (CLAUDE.md "No Over-Engineering").

**2. Nonce comparison treats gap-ahead as okay** — **Accepted.** The original `current > signed` silently proceeded when the current nonce was greater than signed (treating it as already-included) but did not flag the inverse case (signed nonce ahead of current). On an ephemeral account, a gap-ahead signed nonce stalls the mempool indefinitely. Changed to strict `!==` with directional error messages.

**3. Missing chainId validation on signed self-transfer** — **Accepted.** A self-transfer presigned on the wrong chain would otherwise pass inspection. Added optional `expectedChainId` to `MoneriumSelfTransferExpectation` and pass `getNetworkId(Networks.Polygon)` from the handler. Tests added for both positive and negative cases.

**4. `assertEqual` undefined coercion in permit context** — **Accepted.** Previously `String(undefined) === "undefined"` would produce a confusing "undefined does not match expected X" error. Now throws an explicit `"missing from signed context"` error. Test added for the entirely-missing-context branch.

### My Independent Findings

**Accepted & applied** (in commit `99710cd60`):
- Same as Copilot #2, #3, #4 above
- Added explanatory comment for the 30s `setTimeout` post-broadcast wait (CLAUDE.md: always comment `setTimeout` rationale)

**Considered but skipped:**
- *Same-chain passthrough balance gate* — `squidRouterPhaseHandler` returns early without an explicit balance check, but the downstream `destinationTransfer` phase enforces this. Adding a duplicate check would be over-engineering.
- *Infura URL redaction in error messages* — Already covered by `sanitizeRpcErrorMessage` in `clientManager.ts`.
- *Address/name casing nits* — Cosmetic only.

### Test Coverage Added

Tests committed for the affected areas:
- `monerium-permit.test.ts` — value/version mismatch, missing context, nonce drift, allowance-sufficient skip
- `monerium-self-transfer.test.ts` — amount mismatch, chainId positive + negative
- `squid-router-phase-handler.test.ts`, `monerium.test.ts`, `clientManager.test.ts`, `helpers.test.ts`

### Verification

- 10/10 monerium tests pass (`bun test`)
- `bun typecheck` shows no new errors introduced by these edits (4 pre-existing branch errors remain; pre-existing on `06ca1a5a9` before any of my changes)
- `bun lint:fix` clean on touched files

### Note on `mykobo-eur-offramp.integration.test.ts`

This untracked file references several exports that don't exist in `@vortexfi/shared` (`MykoboApiService`, `MykoboCurrency`, etc.) and `RampPhase` values that aren't defined. It appears to be WIP for a separate feature. **Not committed** as part of this PR — recommend handling separately.

PR #1175, opened 2026-05-28
Deploy missing Squidrouter relayer contracts

gianfra-t:
Deploys relayer contracts for missing EVM chains. Adds addresses to config object.


ebma (approved):
Looks good to me 👍

PR #1176, opened 2026-05-29
Make monerium vs mykobo flows distinguishable

ebma:
Adds a new column to the ramp and quotes table that denotes the flow for which the quote/ramp was created. The flow_variant will be assigned by the respective backend, thus we need to set the env variables accordingly. In the future, all backends will use the Mykobo flow, but for now, it's still the Monerium variant. While this flow variant is only relevant for the respective onramps, we'll add it for all quotes/ramps for now and remove it later once the fallback is no longer needed.


Sharqiewicz (approved):
LGTM

PR #1188, opened 2026-06-02
Mykobo base improvements - adjustments proposal

Sharqiewicz:
### Refactor / clarity (no behavior change)
  - Extracted the repeated `isEurToEurcBaseDirect || isBrlToBrlaBaseDirect` check into a
    single `isFiatToOwnStablecoinBaseDirect` helper, used across the three phase handlers —
    one source of truth instead of a disjunction duplicated per site.
  - Named the bridge-delivery threshold constant (`MIN_BRIDGE_DELIVERY_RATIO`) so the
    90% intent is readable without parsing a comment.
  - Documented why both direct-transfer gates (state flag + quote-derived check) exist.
  - Frontend: replaced a `userEmail as string` cast with TanStack Query `skipToken`, which
    narrows the type without an assertion and folds in the `enabled` gate.


ebma (approved):
Let's fix the CI but looks good to me 👍

PR #1211, opened 2026-06-09
Add morpho demo

### LINKED ISSUE #1202: Test Alphaping vault onramps and offramps
## Context
This is to test whether we can directly onramp or offramp to Alphaping vaults

## Todo
* Use the Morpho API to test onramp and offramp to and from a Alphaping vault
* Create a EVM account as in this flow the vault deposit and withdraw needs to go through an wallet. 
* Test could be done for $10
* The tests should be able to replicated with the potential client

## Notes
* https://docs.morpho.org/tools/offchain/api/get-started/ 
* LLM Version - https://docs.morpho.org/tools/ai/llms/ 
* Vault address - https://etherscan.io/address/0x0bF0164D17469241B6E086dA4016DCc54FEAA334 
* Vault - https://app.morpho.org/ethereum/vault/0x0bF0164D17469241B6E086dA4016DCc54FEAA334/alpha-usdc-delta-v2

## Question
* Check if Morpho provides and option to create wallet on the fly (ex. via privy or any other setup)

gianfra-t:
Closes: #1202


ebma:
When testing the onramp I noticed two issues:
- when a subsidy is configured in the partners table, the ramping flow will move to the 'finalSettlementSubsidy' phase but cannot move out of it because all the transition handlers are gone
- the wording of one of the phases is wrong on the UI. It says 'Transferring Alpha Vault from Polygon to arbitrum' although 1) it's not the Alpha Vault token being send around by squidrouter but EURC, and 2) it's not to arbitrum. 


Just feedback so we don't forget if this becomes relevant again.

ebma:
Closing this as it's not required right now.

PR #1227, opened 2026-06-17
Fix biome issues

ebma:


ebma:
@copilot can you please do a full code review of the changes here instead of just doing a partial review? You keep coming up with new issues each time I ask for a review, and I'd just like to have all your suggestions all at once.

PR #1258, opened 2026-07-08
Add BlindPay shadow quote for USDC↔BRLA rebalances

ebma:
## What

Adds an **observational** BlindPay price comparison to the \`usdc-brla-usdc-base\` rebalancing route, to evaluate whether BlindPay is a cheaper source of stablecoin (USDT to feed into Nabla) than the routes we already use.

On each route comparison, \`compareRoutesUpfront\` now fetches a **BlindPay shadow quote** (payin BRL → USDT/USDB) in parallel with the SquidRouter / Avenia / Nabla quotes. It's:
- **Logged in the console comparison**, alongside the executed routes.
- **Surfaced in the Slack completion notification**, e.g.:
  \`\`\`
  *BlindPay shadow* (not routed): 95.220000 USDC-eq | vs SquidRouter 95.400000: -0.180000
  \`\`\`

## Not routed

The shadow quote is **never added to \`candidates\`**, so BlindPay is never actually selected/executed — this is a data-gathering feature only. It never throws, and skips cleanly when BlindPay is unconfigured or the amount is below BlindPay's 5 BRL minimum.

## Implementation

- New \`apps/rebalancer/src/services/blindpay/\` — a minimal client for BlindPay's lightweight payin FX-rate endpoint (\`POST /instances/{id}/payin-quotes/fx\`), which needs only an API key + instance ID (no receiver/KYC/blockchain-wallet setup).
- \`fetchBlindpayShadowQuoteUsdc\` returns a USDC-equivalent (6dp) quote for apples-to-apples comparison.
- Threaded through preflight quotes → rebalance state → Slack completion message.
- Optional \`BLINDPAY_*\` config (documented in \`.env.example\`).

## Config

\`\`\`
BLINDPAY_API_KEY=...
BLINDPAY_INSTANCE_ID=in_...
BLINDPAY_BASE_URL=https://api.blindpay.com/v1   # optional
BLINDPAY_TOKEN=USDT                              # sandbox/dev instances only support USDB
\`\`\`

When \`BLINDPAY_API_KEY\`/\`BLINDPAY_INSTANCE_ID\` are unset, the shadow quote is skipped and the rebalancer behaves exactly as before.

## Testing

- \`tsc\` clean, all 56 rebalancer tests pass.
- Verified live against the BlindPay sandbox: 500 BRL → 95.22 USDC-eq (~0.190 USD/BRL, tracks real BRL/USD); below-minimum amounts return null; Slack line renders with the delta vs the executed route and is absent when unconfigured.

## Notes for reviewers / before prod

- Sandbox returns realistic BRL/USD FX and exposes BlindPay's spread (~0.1%) + flat fee (~\$0.10 on the test instance). Re-confirm fees on the production instance.
- For production, set \`BLINDPAY_TOKEN=USDC\` or \`USDT\` — \`USDB\` is sandbox-only.

### DIFF
--- a/apps/rebalancer/.env.example
+++ b/apps/rebalancer/.env.example
@@ -41,3 +41,14 @@ REBALANCING_DAILY_BRIDGE_LIMIT_USD=10000
 # Main Nabla instance on Base (BRL→USDC route). Leave empty to disable this route.
 MAIN_NABLA_ROUTER=0x...
 MAIN_NABLA_QUOTER=0x...
+
+# BlindPay (optional): logs an observational "shadow" quote (fiat BRL -> stablecoin) during
+# the usdc-brla-usdc-base route comparison, to evaluate BlindPay as a cheaper stablecoin
+# source over time. Never routed. When BLINDPAY_API_KEY/BLINDPAY_INSTANCE_ID are unset, the
+# shadow quote is skipped.
+# BLINDPAY_API_KEY=your_blindpay_api_key
+# BLINDPAY_INSTANCE_ID=in_000000000000
+# Optional, defaults to https://api.blindpay.com/v1
+# BLINDPAY_BASE_URL=https://api.blindpay.com/v1
+# Target stablecoin. Sandbox/dev instances only support "USDB"; production uses "USDC"/"USDT".
+# BLINDPAY_TOKEN=USDT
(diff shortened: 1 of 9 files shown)


ebma:
Superseded by #1257, which already merged this feature (plus unit tests + a request timeout) into staging. Nothing left to merge here.

PR #1262, opened 2026-07-08
Surface fiat-provider failures on ramp endpoints with clear messages and logs

ebma:
## Context

A partner's `POST /v1/ramp/register` failed three times in a row in production. Investigation on the `vortex-production` service showed BRLA/Avenia rejecting the end-user's pay-in ticket with `{"error":"user is blocked"}` on `createPixInputTicket` (`POST /v2/account/tickets`). Two gaps made this hard to handle:

1. **Opaque caller response** — the provider error bubbled up as a plain `Error`, was converted to a `500` and masked to `{"code":500,"message":"Internal server error"}`. The partner couldn't tell it was an account/provider issue and retried uselessly.
2. **No diagnostic trail** — logs only said `Error registering ramp`; nothing recorded *which* Avenia call rejected the account.

## Changes

- **`ProviderApiError` base class** (`packages/shared/src/services/providerApiError.ts`) carrying `provider`/`endpoint`/`method`/`status`/`responseBody`. `BrlaApiError` (`avenia`) and the new `AlfredpayApiError` (`alfredpay`) extend it and are thrown from their services' `sendRequest` paths. The message shape (`Request failed with status '<code>'. Error: <body>`) is **preserved**, so existing consumers (BRLA controller `handleApiError`, Alfredpay controller substring guards) keep working.
- **`mapProviderFailure`** in `ramp.controller.ts`, wired into the `register`, `update` and `start` catches:
  - maps a `ProviderApiError` to **422** (upstream `4xx` — account/request rejected) or **502** (upstream `5xx`/transport — provider unavailable) with a generic "payment provider" message;
  - **never** forwards the raw upstream body to the caller;
  - logs the failing `provider`/`endpoint`/`method`/`status` (never query params, which may carry a PIX key / PII);
  - classifies the event as `provider_error` instead of `internal_error`.
- Removed a stray `console.log` in `createPixInputTicket`.

Alfredpay onramp runs inside the register flow, so its provider failures now get the same treatment automatically. The Avenia and Alfredpay controllers handle their errors inline (`res.status().json()` / `handleApiError`) and never reach the global error middleware, so no middleware catch-all was needed (it would have been dead code).

## Tests & verification

- `mapProviderFailure` unit tests: Avenia `4xx`→422 (sanitized, no body leak), Avenia `5xx`→502, Alfredpay tagging, telemetry classification, non-provider passthroughs.
- BRLA controller regression test locking the message-format invariant (`BrlaApiError` still parses via `handleApiError`).
- `bun build:shared` ✓ · `bun typecheck` ✓ · `bun lint:fix` (no changes) ✓ · affected test suites 20/20 pass.

## Security spec

`docs/security-spec/07-operations/api-surface.md` error-handling section updated (supports invariant #5 — no internal-detail leak in production).


ebma:
Thanks for the review — addressed all six in f8a1c2ac6:

1. **Prototype chain** — `ProviderHttpError` now calls `Object.setPrototypeOf(this, new.target.prototype)` and sets `this.name = this.constructor.name`, matching the convention in `api/errors/providerErrors.ts` and `xcm/send.ts`. `mapProviderFailure`'s `instanceof` check relies on this.
2. **Name collision** — renamed the shared class to **`ProviderHttpError`** (file `providerHttpError.ts`) so it no longer clashes with the price-layer `ProviderApiError` in `api/errors/providerErrors.ts`. Subclasses (`BrlaApiError`/`AlfredpayApiError`) and the controller import updated.
3. **Spec vs behavior (body)** — `mapProviderFailure` now logs a **truncated (300-char) `providerResponseBody`** in the structured context, so the reason (e.g. `user is blocked`) is captured server-side. It's never returned to the caller. Spec wording updated to match.
4. **Duplicate `./squidrouter` export** — removed the redundant second export from the services barrel.
5 & 6. **`fetch()` transport failures** — both the BRLA and Alfredpay request paths now wrap `fetch()`; DNS/timeout/connection-reset errors throw the provider error with **`status: 0`**, which `mapProviderFailure` normalizes to **502** (as documented) instead of a generic 500.

Tests updated for the new `providerResponseBody` field, plus new cases for the status-0 transport mapping and body truncation. `typecheck`/`lint` clean, affected suites green (38 tests).

ebma:
Addressed the endpoint-path nit in ba5d8a988 — the provider-error spec bullet now spells out the full paths (`POST /v1/ramp/register`, `POST /v1/ramp/update`, `POST /v1/ramp/start`) instead of the `/update`, `/start` shorthand.

PR #1263, opened 2026-07-08
fix(api): embed provider error context in ramp log message (metadata was dropped)

ebma:
## Context

Follow-up to #1262. After that PR shipped to production I checked the logs: the caller-facing fix works (`POST /v1/ramp/register` now returns **422** instead of an opaque 500), but the **structured provider logging never actually appeared in the logs**.

Root cause: the app logger (`config/logger.ts`) uses a `printf` format that only interpolates `{ timestamp, level, message, label }` and **drops any metadata object** passed as the second argument. So:

```js
logger.error("Error registering ramp", { ...logContext })  // logContext silently discarded
```

logged only `Error registering ramp` — we still couldn't tell *which* Avenia call failed (`POST /v2/account/tickets`) or why. Searching production logs for `providerEndpoint` / `/v2/account/tickets` returns nothing, confirming the drop.

## Change

- Add **`formatProviderContext(logContext)`** which renders the provider context as a message **suffix**, and append it to the register/update/start error logs:
  ```
  Error registering ramp — provider=avenia call=POST /v2/account/tickets status=400 body={"error":"user is blocked"}
  ```
- Non-provider failures get an empty suffix, so their log line is unchanged.
- The body stays **truncated to 300 chars** and **server-side only** (never returned to the caller).
- Removed the now-misleading `...logContext` spread into the (dropped) metadata object.

Scope is deliberately narrow: I did **not** change the shared logger to serialize metadata globally — that would alter log volume/format across the whole app and risk surfacing previously-dropped fields. This fixes only the three ramp error logs that need the diagnostic.

## Tests & verification

- `formatProviderContext` unit tests (suffix contents for a provider error; empty string for non-provider).
- `bun typecheck` ✓ · `bun lint:fix` (no changes) ✓ · ramp controller suite 10/10 pass.
- Security spec (`07-operations/api-surface.md`) updated to note the logger drops metadata objects and the context is embedded in the message.

## Related (not in this PR)
- The blocked user itself is external (Avenia-side); root cause unchanged.
- The partner client is retrying the blocked CPF on a ~16-minute loop — worth flagging to them separately.


ebma:
Addressed in 77fce4c03.

The log-forging concern is valid — the provider response body is untrusted external input (echoed from the provider's HTTP response), so a `\r`/`\n` in it could split the log line or forge a fake entry (CWE-117), and an ESC could inject terminal color codes. Added `sanitizeProviderBody()` which collapses control characters (`0x00`–`0x1F`, `0x7F`) to a single space before the (already 300-char-truncated) body is stored in the log context. Added a test with a body containing `\r\n` + an ESC sequence asserting neither the logged body nor the formatted suffix can span multiple lines.

On the "undefined if fields missing" part: the other interpolated fields (`provider`, `providerEndpoint`, `providerMethod`, `providerStatus`) are code-defined constants/numbers set together by the sole caller (`mapProviderFailure`), so they can't be partially missing in practice — only the body is untrusted, and that's now sanitized.

PR #1264, opened 2026-07-08
feat(api): use Binance USDT/BRL spot as primary USD<>BRL rate source

ebma:
## What & why

Brazilian clients care about the **Binance USDT/BRL spot price** more than the official fiat FX mid-market rate. This makes `PriceFeedService.getUsdToFiatExchangeRate()` derive the USD↔BRL rate from Binance's [`USDT_BRL` spot](https://www.binance.com/en/trade/USDT_BRL?type=spot) first, falling back to fastforex, then CoinGecko.

New provider priority: **Binance USDT spot → FastForex → CoinGecko**.

The Binance USDT price is treated as USD (consistent with the existing USD-like-stablecoin 1:1 model), so `USDTBRL` yields the USD→BRL rate that feeds all quote / conversion / subsidy math.

## Changes

- **`priceFeed.service.ts`** — `getBinanceUsdtToFiatRate()` hits the public `GET /api/v3/ticker/price?symbol=USDTBRL` endpoint (no API key); a Binance-first branch in `getUsdToFiatExchangeRate()`.
- Generalized the sanity-band guard (`assertFastforexRateWithinSanityBand` → `assertRateWithinSanityBand(provider, …)`) so the Binance rate is validated against the same CoinGecko reference and per-currency spread limits (2% for BRL) as fastforex. A broken/out-of-band Binance feed falls through to fastforex.
- **`verifyBinanceReachability()`** runs at startup (`index.ts`) and logs an **error** for each unreachable mapped market, so a geo-block (HTTP 451) or outage is loud instead of a silent fallback to the fiat rate.
- **`config/vars.ts` + `.env.example`** — `BINANCE_API_URL` (defaults to `https://api.binance.com`).
- **Security spec** — new `05-integrations/binance.md`; updated `fastforex.md` + `README.md` for the new provider order.
- **Tests** — rewrote the rate-priority suite for the new order + dedicated Binance and reachability tests.

## Scope / decisions

- **Only BRL is mapped** (`BINANCE_USDT_FIAT_SYMBOLS = { BRL: "USDTBRL" }`). Other fiats (ARS/EUR/MXN/COP) have no reliable Binance USDT market, so they skip Binance and keep today's fastforex behavior unchanged.
- Kept the **2% sanity band** on the Binance rate (checked vs CoinGecko USDC/BRL — same nature of price, so it doesn't fight intent; it only rejects a broken feed).
- Fails safe: any Binance error → fastforex → CoinGecko; all-provider failure still fails closed.

## Verification

- `bun test priceFeed.service.test.ts` → **44 pass**.
- `bun lint:fix` clean; `bun typecheck` adds no new errors (pre-existing `@vortexfi/shared` export errors on `staging` are unrelated — they need `bun build:shared`).
- Live end-to-end call from the service returned the real Binance spot rate (`getUsdToFiatExchangeRate("BRL")` ≈ 5.20). Service is deployed in Frankfurt, which is not on Binance's geo-block list.

## Ops note

Binance enforces geo-blocks at the **IP level** (primarily US → HTTP 451). Frankfurt/EU is fine in general, but confirm the deployed egress IP once — the new startup log will surface a block as an error line (`Binance price feed UNREACHABLE for USD-BRL`).


ebma:
Addressed both review comments in e38b18fb8:

1. **Symbol validation** — `getBinanceUsdtToFiatRate()` now rejects any response whose `symbol` doesn't match the requested one, so a misconfigured `BINANCE_API_URL` (or a proxy returning a different market) can't be accepted as a valid-but-wrong price. Added a symbol-mismatch fallback test.
2. **Non-blocking probe** — `verifyBinanceReachability()` is now fire-and-forget at startup (`void ...`), so a blocked/hanging Binance call (up to the 30s fetch timeout) can't delay boot. It still emits the same diagnostic error/info logs.

Security spec (`binance.md`) updated for both. All 45 priceFeed tests pass.

PR #1266, opened 2026-07-10
docs(api): SEO metadata + all-corridor updates for AI-agent page and integration skill

ebma:
## Summary

Brings the partner docs and the `vortex-integration` skill in line with what the SDK actually supports today (all six fiat corridors), and adds SEO metadata for the Apidog docs site.

### SEO
- Each page entry in `docs/api/apidog/page-manifest.json` now carries an `seo` block (`metaTitle`, `metaDescription`, `keywords`) targeting integrator search intent. Apidog has no API for SEO settings, so these are the repo source of truth to be applied manually in the Apidog UI (Page → SEO Settings).
- Convention documented in `docs/api/README.md` (§ SEO Settings).

### AI Agent Integration page (`12-ai-agent-integration.md`)
- States explicitly that every integration path supports all live corridors: BRL (PIX), EUR (SEPA), USD (ACH), MXN (SPEI), COP, ARS (CBU).
- Rewrote the BRL-only fiat-settlement section (§ D.5) to cover all corridors: PIX QR at register (BRL), `ibanPaymentData` at register (EUR), `achPaymentData` after start (USD/MXN/COP/ARS), plus per-corridor payout rails on sells.

### `vortex-integration` skill
Verified against the SDK source before rewriting:
- Corridor reality updated — the skill still claimed "BRL only, EUR handlers throw at runtime"; EUR (Mykobo) and the Alfredpay corridors are fully implemented in both directions.
- New recipes: `start-ramp-eur-sepa` and `start-ramp-bank-transfer` (incl. `listAlfredpayFiatAccounts`, `achPaymentData`, user-linked `sk_*` key requirement, `FiatToken.EURC` gotcha).
- BRL recipes updated for the `taxId` deprecation (derived server-side from the user-linked key) and the `submitUserTransactions` helper.
- Error-recovery table extended with `InsufficientBalanceError`, `MykoboKycRequiredError`, `AlfredpayOnrampKycRequiredError`, and the ephemeral-freshness errors.
- Fixed the sandbox base URL: `api.sandbox.vortexfinance.co` does not resolve; the live host is `api-sandbox.vortexfinance.co` (matches the OpenAPI `servers` entry; confirmed with an HTTP probe).

## Notes for reviewers
- ARS is type-wired in the SDK but has no worked example; documented as live per code — worth a sandbox run before promoting it.
- EUR can be gated off server-side per environment; the skill now says to verify with a quote.
- After merge, the SEO values and the updated page 12 need to be applied/re-imported in Apidog manually.

## Verification
- `bun run docs:api:check` passes for the changed files (the known pre-existing sandbox-page finding is untouched).
- `bun lint` clean.


ebma:
Second commit: swept the remaining doc pages against the API backend code (staging **and** main) and fixed everything that had drifted from actual behavior:

- **Auth model** (pages 02, 03, 09): ramp registration requires a user-linked `sk_*` key in *every* corridor — `ramp.service.ts` rejects registration when the key resolves to no user. The docs still said partner-scoped keys suffice for BRL/EUR with `taxId` identifying the user; the `taxId` is actually derived from the account and any provided value is only cross-checked (`resolveAveniaAccountForRamp`).
- **Quote response shape** (page 06): the sample showed a nested `fee` object; the real `QuoteResponse` has flattened `networkFeeFiat`/`totalFeeFiat`/`feeCurrency` fields plus `*Usd` twins.
- **Payment-instruction timing** (pages 04, 12): `depositQrCode`/`ibanPaymentData` are released only after presigned txs validate (update response / `GET /v1/ramp/{id}`), not on register.
- **Whitelabel** (page 13): provider names removed from KYB prose.
- **Skill**: EUR register gate returns `503` while EURC quotes still succeed; webhook `transactionType` values are `BUY`/`SELL` (not `onramp`/`offramp`).

⚠️ **Needs a product decision, not changed here**: pages 01/09 present EUR/SEPA as live, but `be52569e4` ("disable euro flows") is on `main` — EUR registration currently 503s in production. If EUR enablement isn't imminent, the overview and corridors pages should say "temporarily unavailable".

ebma:
Thanks — all three review comments were accurate against the backend; fixed in 1eb4562d0.

1. **Register request shape** (SKILL.md REST example + AI-agent § D.3): `POST /v1/ramp/register` takes `signingAccounts` (array of `{ address, type }` where `type` is `"Substrate"` or `"EVM"` — the only two values `AccountMeta`/`EphemeralAccountType` accept), not `ephemeralAccounts`, and there is no `publicKey` field (attribution rides on the quote's `apiKey`). Verified in `ramp.controller.ts:109-117` and `RegisterRampRequest`. Also corrected the "Stellar leg / multiple sr25519 keypairs" guidance — `generateEphemerals()` produces exactly one Substrate + one EVM account, so the ephemeral-custody page was updated too.

2. **BRL onramp missing-param error**: replaced the `MissingBrlParametersError` entry with the real `400 "Parameter destinationAddress is required for onramp"` (`ramp.service.ts:1115`), and noted the SDK parser only maps the legacy `"...destinationAddress and taxId..."` string.

3. **BRL offramp missing-param errors**: replaced with the real `400 "pixDestination is required for offramp to BRL"` (`ramp.service.ts:1043`) and `"User address must be provided for offramping."` (`evm-to-brl-base.ts:57`), with the same parser caveat.

**Follow-up worth a separate issue (not this docs PR):** comments 2 and 3 surface a real SDK bug — `parseAPIError` (`packages/sdk/src/errors.ts:476,494`) still matches the pre-`taxId`-derivation backend messages, so `MissingBrlParametersError` / `MissingBrlOfframpParametersError` no longer fire; callers get a generic error instead. Fixing that is an SDK behavior change and needs a test, so I've left it out of here. Happy to open it as its own PR if you'd like.

ebma:
Second review pass — all three comments were accurate; fixed in 116a4a6e7.

1. **Moonbeam is EVM, not Substrate** (AI-agent § D.3): confirmed — `createMoonbeamEphemeral()` produces the EVM ephemeral (`VortexSdk.ts:364`) and `normalizeAndValidateSigningAccounts` routes Moonbeam as `EVM`. Moved Moonbeam to the EVM-account bullet and called out explicitly that it takes an `EVM` signing account. The Substrate account now lists only Pendulum, AssetHub, Hydration.

2. **OpenAPI still listed `Stellar`**: took the "update the OpenAPI to match" option. Removed `Stellar` from all three `signingAccounts`/`AccountMeta` enum sites, regenerated `vortex.openapi.d.ts`, and reworded the affected schema descriptions (which also carried stale `taxId` requirement text). The backend only recognizes `Substrate`/`EVM` — `normalizeAndValidateSigningAccounts` (`ramp.service.ts:139-143`) silently *ignores* any other type rather than rejecting it, so the AI-agent page now says "only recognized values … any other type is ignored" rather than "no other value is accepted".

3. **"5 presigned transactions per ramp"** (skill Global Context): corrected. Verified in `validation.ts:256-273` that the count is enforced **per ephemeral-signed transaction** — 1 primary + exactly 4 backups (consecutive nonces) in `meta.additionalTxs`, `NUMBER_OF_PRESIGNED_TXS = 5` — and a ramp can contain several such transactions across phases. Reworded accordingly; this now matches the existing D.4 "5 consecutive nonces" note.

One deliberate non-change: `01-overview.md` still lists Stellar in the "networks Vortex coordinates value across" sentence. That's a high-level architecture statement (Pendulum↔Stellar bridge), not a signing-account claim, so it's not affected by the enum removal.

`bun run docs:api:check` passes (only the pre-existing sandbox recovery-phrase finding), types regenerated, lint clean.

PR #1288, opened 2026-07-24
Charge and collect vortex/partner fees on Alfredpay corridors

ebma:
## Summary

The Alfredpay corridors (USDT-based MXN/COP/ARS/USD on- and off-ramps) displayed vortex and partner-markup fees but never collected them on-chain — their routes built no `distributeFees` transaction. This PR lands the fix in two commits:

1. **Interim honesty fix** (first commit): forced the vortex/partner components to zero so displayed fees matched reality.
2. **Full fee collection** (second commit): both directions now charge the components and collect them on Polygon; the zeroing is lifted.

## How collection works

- **Charging** — the onramp deducts vortex/partner fees from the Alfredpay mint (bridge/transfer legs move mint − fees); the offramp deducts them from the bridged USD leg before pricing the Alfredpay payout. The fee residual stays on the Polygon ephemeral.
- **Distribution** — a new `distributeFees` phase (last before `complete`, entered from `destinationTransfer` / `alfredpayOfframpTransfer` only when fee transfers exist) broadcasts plain sequential ephemeral-signed USDT transfers: network+vortex to the vortex `payout_address_evm` (with the env fallback), partner markup to the pricing partner's address. Per-nonce hashes make retries idempotent. The SDK signs the new unsigned txs generically — no SDK changes.
- **Failure safety** — fees are collected only after the user-facing leg succeeded. The onramp mint fallback stays full-mint; the offramp refund fallback is sized deposit + fees, so failed ramps return the user's full value. `finalSettlementSubsidy` (SELL) tops the ephemeral up to deposit + fees so the fee transfers are always covered.
- **Subsidy interaction** — both Alfredpay discount engines subtract the charged fee from the subsidy's expected-output target (spec invariant 7: subsidization must not bypass fee collection). Without this, the unbounded-subsidy string-compare bug (SPEC-009, fixed on the review branch) would silently refill the fee via `finalSettlementSubsidy`.
- **Why not Multicall3** — `aggregate3` executes calls with the Multicall3 contract as `msg.sender`, so a batched ERC-20 `transfer` cannot move the ephemeral's tokens. This also means the **existing Base partner-split path is broken** — documented as a new open checklist item in `fee-integrity.md` (separate fix).

## Notes for reviewers

- No DB/state-machine seeding needed: phase successors are decided in code (`nextPhaseSelector`); the `seed:phase-metadata` script references a deleted file and the CLAUDE.md claim about DB-seeded transitions is stale.
- `validation.ts` now classifies `distributeFees` on Polygon as an EVM tx (was Substrate-by-default outside Base).
- Zero-fee quotes build no fee transfers and skip the phase entirely — existing corridor flows unchanged.
- The offramp corridor tests now set `world.squidRouter.toTokenDecimals = 6`; the previous 18-decimal default shrank the bridged amount to ~0 and the unbounded-subsidy bug silently padded it back, masking fee math.
- Spec edits will conflict with the `codex/security-spec-review-2026-07-24` rewrite of `fee-integrity.md` — port the Alfredpay section + the new Base Multicall3 item when the branches meet.

## Follow-up: broken Multicall3 partner split on Base (fixed here too)

While implementing the Polygon distribution it turned out the existing **Base** partner-split path was broken: it batched plain ERC-20 `transfer` calls through `Multicall3.aggregate3`, which executes with the Multicall3 contract as `msg.sender` — the batch could only move the contract's (empty) balance and would revert, failing any Base ramp with nonzero partner markup plus a configured partner payout address. Third commit:

- The Base builder now shares the sequential-transfer implementation with the Polygon builder (one ephemeral-signed USDC transfer per recipient at consecutive nonces); `contracts/Multicall3.ts` is deleted (no remaining users).
- The distribute-fees handler broadcasts EVM fee transfers with per-nonce hash idempotency on both networks; the legacy single `distributeFeeHash` is honored for in-flight ramps.
- Payout addresses are now only required when a quote actually carries nonzero fees (previously zero-fee quotes also demanded a vortex payout address at registration).
- Regression test in `brl-offramp.scenario.test.ts`: a profile-priced partner (markup + vortex fee, 1 USDC each) — both payout addresses receive their exact fee on the fake ledger. The old aggregate3 batch credits nobody, so the assertions fail without the fix.
- `fee-integrity.md`: open checklist item resolved; the "Distribution Mechanisms" section now describes the sequential-transfer scheme.

## Test plan

- End-to-end "fee collection" corridor tests (BUY direct + SELL): fee charged in the quote (output reduced), `distributeFees` in the phase history, and the vortex payout address receives the exact fee on the fake ledger.
- Full API suite: 664 pass, 0 fail (isolated test DB). `bun typecheck` and Biome clean.


ebma:
Superseded by https://github.com/pendulum-chain/vortex/pull/1310.

PR #1290, opened 2026-07-27
Verify the security spec review, drop Stellar, and fix the confirmed findings

ebma:
An AI agent produced a review of `docs/security-spec/` (27 findings, `REVIEW-2026-07-24.md`, first commit here). This branch **verifies every finding against the actual code**, then acts on the confirmed ones: it removes the obsolete Stellar/Spacewalk surface and fixes seven code-bearing issues.

## 1. Review verification

Every citation was re-checked against the spec files and the implementation rather than taken at face value. Outcome: **23 of 27 confirmed, 4 overstated, none refuted.** The four are downgraded in place with a short verification note explaining why:

| Finding | Was | Now | Why |
|---|---|---|---|
| SPEC-004 optionalAuth downgrade | High | Medium | Facts correct, but the spec itself sanctions the token truncation, and the anonymous-claim design bounds the impact |
| SPEC-016 Stellar 2-of-2 cleanup | High | Low | Not a contradiction (spec describes a presigned XDR); also moot — no Stellar code remains |
| SPEC-017 identity-less admin auth | High | Medium | Invariant 7 is an implementation-consistency rule, not an affirmative counter-security mandate |
| SPEC-023 "guaranteed minimum" | High | Medium | The exception is a disclosed liveness pause, not "user receives less" |

Verification also surfaced things the review missed — e.g. `api-surface.md` claims a webhook "Partner ownership" guard that did not exist in code, and the stale-checklist problem is broader than reported.

## 2. Stellar/Spacewalk removal

Stellar support is obsolete (anchors dropped in migration 028). Removed the residue: dead constants, the `getMemoFromCookiesMiddleware` (imported nowhere), a no-op event-recovery hook, `mockSep24`, three unreferenced translation strings, and the FAQ copy still claiming Vortex "is built on the Stellar blockchain". Deleted two fully-obsolete spec modules (`stellar-anchors.md`, `bridge-security.md` — the latter was 100% Spacewalk) and excised Stellar invariants/threats/checklists from ten more. Renamed the misnomer `isSep24Redo` → `isQuoteRedo`. Kept deliberately: migration history, the `MykoboNetwork.STELLAR` enum (mirrors the partner's external API), and one explanatory error message.

## 3. Code fixes

**SPEC-001 — webhooks were cross-tenant readable/deletable** (the one live vulnerability). Registration only checked that a quote existed; the model had no owner column; deletion was by bare UUID; any HTTPS URL was accepted; and `X-Vortex-Timestamp` was not part of the signed bytes, so the replay check the docs recommended was bypassable. Now: owner columns (migration **055**), quote-ownership required at registration, owner-scoped deletion and delivery matching (uniform 404s so IDs can't be probed), SSRF guards (HTTPS-only, no credentials, private/reserved IPs rejected at registration *and* re-resolved before every delivery, redirects disabled), signature over `` `${timestamp}.${body}` ``, and a retry-stable `eventId` for consumer dedup.

**SPEC-002 — raw input amount used as USD.** When a fiat-peg oracle failed and no bridged USDC amount existed, the discount engine passed the raw input amount into USD-denominated math — the exact unit confusion behind the documented ~5× BRL over-subsidy. Now fails closed; the bridged USDC route amount is the only accepted fallback.

**SPEC-009 — `maxSubsidy = 0` meant *uncapped*,** and it is the column default. Worse, the `targetDiscount !== 0` guard in all four discount engines never fired, because **Sequelize returns Postgres DECIMAL columns as strings** (`"0.0000" !== 0` is always true) — so every partner-priced quote was being topped up to oracle-expected output with no cap. The cap now gates the subsidy: `<= 0` disables it, a positive fraction bounds it. ⚠️ *This pattern is worth a wider sweep — any strict comparison against a DECIMAL-backed field has the same bug.*

**SPEC-018 — API-key auth bcrypt-scanned every active key.** All secret keys share the constant `sk_live_`/`sk_test_` prefix, so each auth attempt ran a bcrypt round per active key — unbounded by the per-user cap. Now a 16-char lookup prefix (non-secret key identifier) + SHA-256 constant-time digest gives O(1) verification. **No migration and no partner disruption**: the column was already wide enough, legacy rows fall back to the old path and are upgraded in place on first successful use.

**SPEC-015 — ephemeral freshness.** EVM checked nonce only, so a nonce-0 account holding a native balance passed; now checks both. And every registration validated all nine EVM + two Substrate RPCs, so an outage on *any* chain blocked all new ramps — freshness is now scoped to the chains the quote's route actually signs on. That scoping also closes a latent under-check: non-USDC BRL→AssetHub onramps sign on Hydration, which the old fixed list never validated.

**SPEC-003 / SPEC-012 — `fee-integrity.md` rewritten.** The module described a retired dual fee system whose named functions no longer exist and labelled the Big.js rounding modes backwards. Rebuilt from a code-verified map of the current pipeline. The global "fees last" ordering invariant in `ramp-phase-flows.md`, which contradicted the documented BRL off-ramp flow, is now the per-corridor rule (SPEC-012).

### Deliberately *not* included: SPEC-025 (phase-transition validation)

The review's SPEC-025 finding — that the phase processor persists whatever next phase a handler returns, so the handler is its own judge — is valid and remains open. It is **excluded from this branch on purpose**: the phase-transition approach is being refactored on a separate branch, and landing an interim guard here would only have to be undone. Left for that work to address.

## Deploy notes ⚠️

1. **Run migration 055** (webhook owner columns).
2. **Webhook signature change is breaking for consumers.** Anyone verifying over the raw body alone will reject all deliveries. Partner docs and the example listener are updated; coordinate the rollout, or say the word and I'll add a transition window that sends both signatures.
3. **Set `max_subsidy` (e.g. `0.05`) on the vortex wildcard pricing configs.** With the column default of `0`, quotes now ship with no subsidy top-up and displayed rates drop to raw Nabla output when it underperforms oracle.
4. Partners registering webhooks against quotes not attributed to their key now get `404` — intended, but it may surface sloppy integrations.
5. Availability trade-off: a FIAT-USD oracle outage now fails BRLA/EURC off-ramp quote creation instead of silently mispricing (safety over liveness, as the spec now mandates).

## Testing

727 pass on the API suite, with typecheck and lint clean. New coverage: webhook ownership/SSRF/signature (incl. cross-tenant denial at the HTTP level), discount fail-closed, subsidy cap semantics, API-key lookup + legacy upgrade, and per-corridor freshness scoping. Pricing goldens were re-pinned to the capped-subsidy values — note one previously "passing" golden had snapshotted the uncapped bug (an entire 500 BRL output funded by subsidy).

One pre-existing failure remains (`invite discounts … carrying the vortex fee forward`); it fails on a clean `staging` tree too and is being fixed separately.

> Heads-up for anyone running the API suite locally: concurrent sessions sharing the `vortex_test` database cause uniform 5s `sequelize-pool` timeouts across unrelated files. Use `TEST_DB_NAME=<something> bun test` for an isolated DB.

## Known-open follow-ups

Surfaced during verification, deliberately **not** fixed here (each needs a product decision or is a separate piece of work), and recorded as OPEN items in the spec:

- **Alfredpay corridors display vortex/partner fees that are never collected on-chain** — those routes have no `distributeFees` phase. Either wire up distribution or stop displaying the components. *(Being handled separately.)*
- **EVM fee distribution rounds half-up where Substrate truncates** — should align on round-down.
- **The `anchors` table is dead weight.** It is queried on *every* quote (`calculateAnchorFee`), but all 7 fee-engine subclasses ignore the value and use the provider's own fee instead — verified with a word-boundary check that the parameter appears only in signatures, never in a method body. For BRL corridors it returns a real seeded row (`moonbeam_brla`, 0.75 BRL) that is silently discarded, so there is a plausible-looking config value in prod that has no effect. There is also a **latent hazard**: a second call site in the `BUY` branch of `calculatePreNablaDeductibleFees` *would* subtract that value from the Nabla swap input, and is unreachable today only because all five `assignPreNablaContext` callers are `SELL`. Recommend removing the code path first, then dropping the table in a separate migration after checking prod rows — kept off this PR to avoid unrelated schema churn.
- **Orphan handler**: `moonbeamToPendulumXcm` is registered but nothing transitions into it (the live onramp path uses `moonbeamToPendulum`) — relevant input for the phase-transition refactor.
- Remaining review items are spec-structural (SPEC-026/027), a design-level aggregate subsidy budget (SPEC-008), phase-transition validation (SPEC-025, see above), and the medium-severity auth-taxonomy cluster (SPEC-004/005/006/007, SPEC-017).

## Review guide

Highest-value files: `apps/api/src/api/services/webhook/` (ownership + SSRF), `apps/api/src/api/services/ramp/ephemeral-freshness.ts` (is any signing chain missing from the mapping? under-listing would be a security hole), `apps/api/src/api/middlewares/apiKeyAuth.helpers.ts` (the legacy-key upgrade path), and `docs/security-spec/03-ramp-engine/fee-integrity.md` (does it match your understanding of the fee engine?).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


ebma:
Superseded by https://github.com/pendulum-chain/vortex/pull/1296.

PR #1298, opened 2026-07-31
Add limits and api key handling to dashboard

gianfra-t:


apps/api/src/api/controllers/userApiKeys.controller.ts
```diff
@@ -80,45 +106,78 @@ export async function createUserApiKey(req: Request, res: Response): Promise<voi
...
+    const secretKeyPrefix = getSecretKeyLookupPrefix(secretKey);
+
+    // Lock the profile so concurrent requests for one user cannot both pass the cap check.
+    const records = await sequelize.transaction(async transaction => {
+      const user = await User.findByPk(userId, { attributes: ["id"], lock: transaction.LOCK.UPDATE, transaction });
+      if (!user) throw new Error("Authenticated user profile not found");
+
+      const activeKeyCount = await ApiKey.count({
+        transaction,
+        where: { isActive: true, partnerId: null, partnerName: null, userId }
```
ebma: **[P2] Exclude expired credentials from the active cap**

The validation paths already treat expired rows as unusable, but this count only filters on `isActive`. Because new credentials expire by default and are not automatically deactivated, five expired pairs still consume all ten slots and block creation until the user manually revokes them. Count only rows whose `expiresAt` is null or in the future (preferably distinct credentials after unification), or explicitly deactivate expired rows.

apps/api/src/api/controllers/userApiKeys.controller.ts
```diff
@@ -291,9 +356,20 @@ export async function revokeUserApiKey(req: Request<{ keyId: string }>, res: Res
       return;
     }
 
+    if ((primaryKey.credentialId || pairedKey.credentialId) && primaryKey.credentialId !== pairedKey.credentialId) {
```
ebma: **[P1] Resolve and revoke the credential server-side**

The server has `credentialId`, but the endpoint still revokes both halves only when the caller supplies `pairedKeyId`; omitting the body deliberately leaves one half active. The admin revoke path also updates one selected row. A public-row revocation can therefore leave the corresponding secret usable. Accept a credential ID, resolve its rows on the server, and revoke both atomically. If legacy unpaired rows must be supported temporarily, expose that as an explicit legacy-only operation rather than weakening normal credential revocation.

apps/api/src/models/apiKey.model.ts
```diff
@@ -87,6 +91,11 @@ ApiKey.init(
       field: "created_at",
       type: DataTypes.DATE
     },
+    credentialId: {
```
ebma: **[P2] Enforce public/secret pairing at the auth boundary**

This relationship is currently consumed by listing and revocation, but neither `validatePublicApiKey` nor `validateSecretApiKey` returns it. The quote routes can therefore accept a public key from credential A in the body and a secret key from credential B in the header, apply A's partner/pricing while deriving authentication and user context from B, and only fail later (or mix contexts when ownership happens to align). Return `credentialId` from both validation paths and reject a mismatch immediately whenever both values are present; the unified model can expose this as one `CredentialContext`.

apps/api/src/database/migrations/055-add-credential-id-to-api-keys.ts
```diff
@@ -0,0 +1,58 @@
...
+      SELECT
+        user_id,
+        partner_id,
+        partner_name,
+        base_name,
+        (MAX(id::text) FILTER (WHERE key_type = 'public'))::uuid AS public_id,
+        (MAX(id::text) FILTER (WHERE key_type = 'secret'))::uuid AS secret_id,
+        uuid_generate_v4() AS credential_id
+      FROM normalized
+      GROUP BY user_id, partner_id, partner_name, base_name
```
ebma: **[P2] Do not infer production pairs from display names**

Having exactly one public and one secret row for an owner plus a normalized name does not prove those values were issued together; names are mutable and legacy/manual data can drift. A wrong assignment turns a later revoke into a cross-credential operation. Given the intended no-legacy rollout, use an explicit inventory/mapping or reissue ambiguous credentials, and make incomplete or ambiguous rows fail migration preflight instead of assigning lifecycle identity heuristically.

ebma (commented):
## Review summary

I reviewed the latest implementation and the production migration implications. The limits work is coherent, and the code head immediately before the documentation-only commit had green test and CI checks. The remaining concern is that an API credential is still modeled as two independently valid rows, with a nullable correlation ID that controllers and clients must interpret correctly. That leaves lifecycle and authorization invariants distributed across the API, dashboard, and migration.

The main actionable findings are inline:

- **P1:** revocation can disable only one half of a credential.
- **P2:** public and secret values from different credentials are not rejected at the authentication boundary.
- **P2:** expired rows still consume the active credential cap.
- **P2:** the migration infers credential relationships from mutable display names.

I recommend retaining distinct public and secret values but representing them as one credential record with one profile subject, optional managing partner, environment, expiry, and atomic revocation lifecycle. Self-service and delegated credentials should use the same service; users without an interactive signup should receive unique managed profiles, never a shared dummy identity.

I published the full implementation and production rollout plan in [docs/plans/api-credential-unification.md](https://github.com/pendulum-chain/vortex/blob/add-limits-and-key-to-dashboard/docs/plans/api-credential-unification.md). Before production, run the digest backfill, explicitly map or reissue every active pair, provision missing managed profiles, and make deployment fail closed unless active legacy, unpaired, and ownerless rows are all zero. Digest backfill alone does not establish pair or profile ownership.

PR #1310, opened 2026-08-04
Collect vortex and partner fees on Alfredpay corridors and fix the Base fee split

ebma:
## Summary

Since the #1232 block-flow rewrite, the Alfredpay corridors (USD/MXN/COP/ARS, USDT on Polygon) have **deducted** vortex/partner fee components from the user-facing leg without ever collecting them: no Alfredpay flow had a `DistributeFees` block, and the cleanup lanes only sweep USDC/axlUSDC, so the residual stranded unrecoverably on the Polygon ephemeral. This PR re-establishes fee collection on the block-flow architecture, superseding #1288 (which implemented it on the pre-#1232 pipeline and was never merged). Design and decision record: `docs/adr-0002-alfredpay-fee-collection.md`.

Four commits, each independently reviewable:

1. **Sequential transfers replace the broken Multicall3 fee split on Base.** `Multicall3.aggregate3` executes batched calls with the contract as `msg.sender`, so the split path could never move the ephemeral's USDC — every Base ramp with a nonzero partner markup and a configured partner payout address reverted at `distributeFees`. The EVM builder now emits one plain ephemeral-signed transfer per recipient; the executor broadcasts them in nonce order with one durable financial-operation claim per transfer (confirmed claims replay without rebroadcasting; the balance precondition covers only unpaid transfers; a mined-but-reverted transfer halts for manual reconciliation). A positive markup without a partner payout address fails registration instead of being silently dropped, and the presign merge key now includes the nonce so split fee presigns cannot overwrite each other.
2. **Alfredpay fee collection (flow version 2).** The three Alfredpay flows append a Polygon/USDT `DistributeFees` phase after the user-facing leg, with a pass-through simulation (the deduction already happens in `AlfredpaySubsidizePre` / `AlfredpayOfframp` pricing). Zero-fee quotes flow through the phase as a no-op. Solvency: the onramp pre-swap reserve adds the fee total on top of the swap target (`feeReserveRaw`), the offramp final settlement targets deposit + fees, and the offramp refund fallback returns deposit + charged fees. The computed-markup payout guard now covers the Alfredpay corridors (a configured markup that rounds to zero raw units needs no address), and Polygon `distributeFees` presigns validate as EVM.
3. **Offramp net-rate subsidy alignment.** A target discount promises the user's final net rate; the onramp already sized the subsidy against the fee-net actual, the offramp did not. Both directions now agree (unchanged with `targetDiscount = 0`); `targetDiscount` DECIMAL-string comparisons are normalized.
4. **Security-spec sync.** `fee-integrity.md` documents the sequential-transfer scheme, closes the "uncollected displayed components" and "dropped markup" OPEN items, and corrects the stale checkbox endorsing Multicall3; `discount-mechanism.md` gets the net-rate invariant and drops the false "offramp has no deductible fee" claim; flow diagrams/integration docs updated across six more files.

## Rollout (drain-then-deploy)

Version 1 Alfredpay identities are deliberately **not** kept dispatchable — persisted v1 quotes/ramps fail closed with a clear `@1`-unsupported error. Before deploying: stop Alfredpay quoting, wait out unexpired Alfredpay quotes and in-flight Alfredpay ramps, then deploy. Before enabling nonzero fees: set `DEFAULT_VORTEX_EVM_PAYOUT_ADDRESS` in prod and `payout_address_evm` on the `vortex` (BUY and SELL) and markup-partner pricing rows. Until then the interim corridor-scoped zero-fee pricing rows keep charging at zero. No SDK or frontend changes.

## Test plan

- New corridor scenarios: onramp fee + 1% discount (net 101 USDT delivered, 1 USDT fee collected on the fake ledger), split-fee partial failure (halts for reconciliation, first transfer never double-paid across reprocessing), offramp fee collection (99 USDT deposit + 1 USDT fee; refund fallback decoded as 100 USDT), discounted offramp net-rate, quote-guard rejection + rounds-to-zero acceptance, Base split regression (both payout addresses credited exactly; the old aggregate3 batch credits nobody), v1-identity fail-closed.
- Unit tests for the transfer builder (bucket rounding, fail-closed markup, unbuffered gas) and updated flow/transactions/wiring tests for the v2 topologies.
- Full API suite: 957 pass / 0 fail (isolated test DB). `bun typecheck` and Biome clean.


ebma:
Addressed all four review findings in 546726d79:

1. **Ambiguous-op balance wedge** — the pre-pass now classifies every transfer from its financial-operation row plus chain truth before the balance precondition: `confirmed` and mined-successful `submitted`/`unknown` transfers no longer count toward the required balance; mined reverts and receipt-less ambiguous broadcasts halt for reconciliation immediately (skipping the misleading 60s balance timeout). Reconciliation markers from the durable-claim layer are also no longer downgraded to recoverable retries by the executor's catch.
2. **Discounted fallback over-refund** — the offramp fallback now refunds `bridgeOutputAmountRaw` (the user's exact bridged value) instead of deposit + fees, which under net-rate pricing included the platform subsidy. Blueprint assertion added to the discounted scenario: fallback stays 100 USDT while the deposit is 101.
3. **Unreachable mined-revert branch** — the post-broadcast receipt wait is tri-state (missing → keep polling, success → done, reverted → `FinancialOperationRejectedError` → reconciliation halt), since the operation row is confirmed at broadcast and its reconcile never runs again. New corridor regression test with FakeEvm reverted-receipt support: single broadcast, halt, no rebroadcast on reprocess.
4. **Missing token config strands fees** — registration now fails closed when `totalFeesUsd > 0` and the network/token pair has no configuration; empty intents remain valid only for zero-fee quotes. Unit tests for both branches.

Full API suite 960 pass / 0 fail, typecheck and Biome clean.

ebma:
Addressed all three comments from the Copilot review in da22e2fad:

1. **Legacy `distributeFeeHash` (execution.ts)** — valid, though the affected cohort is narrow: on current staging every EVM broadcast writes a financial-operation row before the hash, so a hash-without-row can only come from a pre-#1232 ramp still resumable at deploy. Guarded anyway: when the hash exists and no operation rows do, the executor resolves it from chain truth first — successful receipt completes the phase (the legacy model's single broadcast covered the whole distribution), a mined revert halts for reconciliation, and no receipt falls through to the pre-rewrite rebroadcast behavior.
2. **Proposal lifecycle (docs)** — correct per `docs/README.md`: the accepted proposal is now `docs/adr-0002-alfredpay-fee-collection.md` (status/context/decision/consequences/links), the proposal file is removed, the index and the PR description are updated.
3. **Untested settlement target (final-settlement-subsidy)** — correct, and a real mutation gap: the fake ledger never debits senders, so completing the corridor could not distinguish target 101 from 102. The discounted offramp scenario now starts the ephemeral at the user's bridged 100 USDT, funds the settlement from the funding account's USDT balance, and asserts the recorded `finalSettlementSubsidy` amount is exactly 2 — deleting the `.plus(fees)` term now fails the test. (The fake's persistent ledger also surfaced a cross-test dependency: the F-001 cap test relies on the funding account holding no USDT, so `beforeEach` now drains it.)

Full API suite 960 pass / 0 fail, typecheck and Biome clean.

